PR #1, opened 2018-06-08
 Update link to Mckernel.

justicezyx:
The current one is not a valid link on the site.


jbreitbart:
Thanks a lot for fixing the link. :smiley:

PR #4, opened 2018-07-25
translate gic functions to Rust

photoszzt:


photoszzt:
@ColinFinck review? 

It's a straight translation from the c code. I haven't tried to use any of the gic code in irq yet.

stlankes:
Colin finished the memory management for aarch64. The next step will be test with your patch. Thanks for the support! You will get feedback as soon as poosible.

PR #6, opened 2019-07-06
Updated gitignore file to work with Rust, CMake and VSCode.

jbreitbart:
Configs are taken from https://github.com/github/gitignore.


stlankes:
+1

PR #28, opened 2020-04-22
Readme: Fix build instructions for Qemu

jschwe:
Use make to be consistent with the gitlab pipeline.
Without the objcpy after building the loader, qemu fails at startup.

I initially wanted to create the pull request for the devel branch, but the Readme at the devel branch seems to be outdated.


stlankes:
You are right. A stupid failure in our README

PR #29, opened 2020-04-22
Ensure CpuFrequency is never explicitly set to zero

jschwe:
detect_from_cpuid could previously set CpuFrequency to zero.
To ensure that this doesn't happen all detect methods set
mhz and CpuFrequencySources via a helper method that ensures
that mhz is never explicitly set to zero (apart from the
creation of the object)


stlankes:
Yes, the idea to introduce `set_detected_cpu_frequency` is good!

PR #36, opened 2020-04-29
First implementation of virtio-fs

tlambertz:
With this pull request it is possible to mount a virtio-fs filesystem and write/read/create/delete files on it. No folder are supported at the moment.

The relevant parts are:
- File system layer. Very simple, every folder under `/` is a mount point, under that the actual filesystem.
- Fuse filesystem. provides the filesystem layer with an actual implementation. Needs a fuse-driver to send messages over
- virtio-fs as said fuse-driver. Uses virtio virtqueues to send the messages via qemu to virtiofsd running on the host
- virtio: provides an abstraction layer over the raw virtio-pci device.

Most parts have only the bare necessity implemented. Big areas for improvement are:
- Better error handling. It is quite stable, but non-happy paths will likely panic
- Folder support
- Better memory-concept for virtio queues. Due to the DMA nature, Rusts memory guarantees do not hold and we should provide a safe abstraction. Currently solved by spinlooping after each request until it is handled.
- Better fuse in-memory-structures. Currently feels a bit hacky.
- Some more abstraction in the PCI layer.
- Investigate what happens when we use a multi-core cpu. There are some global locks, it might degenerate to a serial execution when using the filesystem.
- Better architecture support? Currently a lot of things live in x86/kernel.
- Lots of parts marked with TODO
- Speed improvements. Especially writing involves too many copies right now.


stlankes:
👍 I will test!

src/arch/x86_64/kernel/virtio.rs
```diff
@@ -0,0 +1,689 @@
...
+		}
+		_ => {
+			info!("Virtio device is NOT virtio-fs device, skipping!");
+			return;
+		}
+	};
+
+	// Install interrupt handler
+	// TODO: get irqnumber from pci, don't hardcode 11.
+	irq_install_handler(11, virtio_irqhandler as usize);
```
stlankes, line 679: You should replace line 679 with `irq_install_handler(adapter.irq as u32, virtio_irqhandler as usize);`.

src/arch/x86_64/kernel/virtio.rs
```diff
@@ -0,0 +1,689 @@
...
+		return;
+	}
+	let virtio_device_id = adapter.device_id - 0x1040;
+
+	let drv = match virtio_device_id {
+		0x1a => {
+			info!("Found Virtio-FS device!");
+			// TODO: proper error handling on driver creation fail
+			virtio_fs::create_virtiofs_driver(adapter).unwrap()
+		}
```
stlankes, line 670: Do you have to check class_id and subclass_id? see [https://wiki.osdev.org/Virtio](https://wiki.osdev.org/Virtio)

stlankes:
👍

src/arch/x86_64/kernel/virtio.rs
tlambertz: vendor_id and device_id are currently checked in pci.rs::init_drivers(), so init_virtio_device() is only called with the correct adapters. Since there currently is only one driver, I opted for this simple approach rather than the 'loop through all drivers' linux does. Another check here wouldn't hurt if this were to ever change though.

stlankes: As fas as I know, you are able to identify with the class_id the device type (block, network, etc.)

tlambertz: I am not entirely sure what you mean. Class and subclass id from register 02 (https://wiki.osdev.org/PCI) are not mentioned in the [virtio spec](http://docs.oasis-open.org/virtio/virtio/v1.0/cs04/virtio-v1.0-cs04.html#x1-660001) as far as I can see. If you mean the Subsystem-ID and Subsystem-Vendor-ID, these are optional:

> The PCI Subsystem Vendor ID and the PCI Subsystem Device ID MAY reflect the PCI Vendor and Device ID of the environment (for informational purposes by the driver). 

I differentiate between the device types via the PCI Device ID (from register 0), which is 0x1040+virtio_id. This id is 0x1a for virtio-fs.

I suppose osdev mentiones the subsystem ID, since 
> Transitional devices MUST have the PCI Subsystem Device ID matching the Virtio Device ID, ...

Issue #30, opened 2020-04-23
hermit in uhyve can't read processor frequency


jschwe:
On my system using uhyve hermit fails to read the cpu-frequency with `detect_from_cpuid` and `detect_from_cpid_tsc_info`.

The first issue is that `detect_from_cpuid` can return a frequency of 0, which is Some, but not a valid value. This is fixed by https://github.com/hermitcore/libhermit-rs/pull/29.
The second issue I encountered is that `detect_from_cpid_tsc_info` fails with a panic, due to a division by zero here: https://github.com/gz/rust-cpuid/blob/b64cef2bbfd577b2d3e1c65b8ca76456730efa65/src/lib.rs#L3769
Is this something that should be fixed upstream or should we catch the panic on our side (I'm new to rust but I assume that's possible)?

The next method `detect_from_hypervisor` works fine for me.

stlankes:
When do you get the panic? If get_tsc_info doesn't return `None`, but the frequency isn't correctly evaluated?

jschwe:
The panic occurs here (inside the rust-cpuid module) : https://github.com/gz/rust-cpuid/blob/b64cef2bbfd577b2d3e1c65b8ca76456730efa65/src/lib.rs#L3769
This is due to a division by zero, which is because eax is apparently zero.

stlankes:
Can you specify your processor? Maybe the output of `lscpu`...

jschwe:
Sure. As you can see below, A virtual ubuntu has no problems reading the basic CPU information from kvm, so uhyve should have access to this information too.

I also want to note that `detect_from_hypervisor()` only works if uhyve is run on native ubuntu. If uhyve is run on a virtual machine (nested virtualization, see also https://github.com/hermitcore/uhyve/issues/6) this method fails, even though the information is available in the virtual machine. I suspect these two issues might be related in some way.

<details><summary>lscpu native Ubuntu</summary>
<p>

```
$ lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  1
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               94
Model name:          Intel(R) Core(TM) i5-6600K CPU @ 3.50GHz
Stepping:            3
CPU MHz:             800.017
[...]

PR #37, opened 2020-04-29
check if tsc_frequency determines a valid value

stlankes:
@jschwe Can you check this PR on your system. This should solve issue #30.


stlankes:
Thanks!

PR #47, opened 2020-05-06
Add new kernel-command-line option for setting argv

tlambertz:
I was annoyed with having to recompile my benchmark each time I wanted to change some options, so I implemented a new kernel-command-line option to set argv/c.

Previously, there were only `-freq` and `-proxy`, both set some kernel-internal stuff.

This pull request adds a new option `-args`. Since I did not want it to clash with any existing (or future) options which might get introduced, I wrote a simple "tokenizer" which can handle splitting strings at spaces while respecting quoted strings.

Exampe cmdline:
```
qemu-system-x86_64 -cpu qemu64,fsgsbase,rdtscp,xsave,fxsr,+x2apic -enable-kvm \
 -display none -smp 1 -m 1G -serial stdio -kernel LOADER -initrd APP \
 -append "-args '-a -b --opt \"some string\"'"

--> argv = ["{name}", "-a", "-b", "--opt", "some string"]
```

If you want to have a `"` inside an argument, the escaping is a bit awkward since we have a string in a string in a string. Consider calling bash-like `rusty-demo "test\" string" -v`. Starting the kernel with qemu we need to escape as follows:
- bash: `qemu-system-x86_64 .... -append "... -args '\"test\\\\\" string\" -v'"`
- qemu `argv[-2:-1]`: `[-append, ... -args '"test\\" string" -v']`
- libhermit kernel-cmdline tokens: `[..., -args, "test\" string" -v]`
- rusty-demo argv = `[test" string, -v]` 

I am not sure if uhyve currently supports setting the command line, I just had a quick look and it seems it is not supported there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,12 @@ $ qemu-system-x86_64 -display none -smp 1 -m 64M -serial stdio  -kernel path_to_
 
 It is important to enable the processor features _fsgsbase_ and _rdtscp_ because it is a prerequisite to boot RustyHermit.
 
+You can provide arguments to the application via the kernel commandline, which you can set with qemu's `-append` option. Since both the kernel and the application can have parameters, they are separated with `--`:
+
+```bash
+qemu-system-x86_64 ... -append "kernel-arguments -- application-arguments"
+```
+
 #### Using virtio-fs
 
 The Kernel has rudimentary support for the virtio-fs shared file system. Currently only files, no folders are supported. To use it, you have to run a virtio-fs daemon and start qemu as described in [Standalone virtio-fs usage](https://virtio-fs.gitlab.io/howto-qemu.html):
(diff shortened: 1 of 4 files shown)


jounathaen:
Very nice, but why do you add `-args`? Shouldn't `-append "a -b --opt \"some string\""` be sufficient?

jounathaen:
An alternative to using `-args` would be to separate libhermit options from application options by `--`. That's how it is done for example in git.

tlambertz:
I just saw that uhyve already supports arguments via the `UHYVE_PORT_CMDSIZE` hypervisor call. So this change is only relevant when running with qemu.

Splitting with `--` seems like a good idea, this way we remove one level of string-stacking.

jounathaen:
Could you also add this to the README?

tlambertz:
done.

stlankes:
👍

Issue #45, opened 2020-05-06
Libhermit's use of `halt` and the `panic_handler`


jounathaen:
When panicing, RustyHermit loops `arch::processor::halt()` infinitely. Uhyve threatens `hlt` instructions by shutting down the vm with exit code 0. (https://github.com/hermitcore/uhyve/issues/7)

Related Issue: https://github.com/hermitcore/uhyve/issues/11

 RustyHermit should not use `arch::processor::halt` when running in a VM but use the proper shutdown command.
- It should probably keep using a hlt loop when running on real hardware so that the error message can be observed on e.g. the monitor.
- It could write to the shutdown port, so that uhyve shuts down before the hlt-loop. I don't know how real Hardware handles `out` instructions to a non-existing port.

stlankes:
Doesn't the patch solve this issue?

jschwe:
Is this issue completely solved? 

I'd like to link to the [Travis CI build](https://travis-ci.com/github/hermitcore/uhyve/jobs/334213154) for commit 4a2381ccc192b3d699fd4b54fb007e297bac9e2d. In release mode it seems a race condition occured that caused a Panic due to a BorrowError: `[0][!!!PANIC!!!] /home/travis/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/src/libcore/cell.rs:798: already mutably borrowed: BorrowError`
Given the output I assume that the BorrowError occured in the kernel or demo and not in uhyve. 
The thing that makes this related to this isssue is that uhyve doesn't exit after the panic. 

I've saved the log output here so it doesn't get lost if someone reruns the build.
[log.txt](https://github.com/hermitcore/uhyve/files/4633138/log.txt)

_Originally posted by @jschwe in https://github.com/hermitcore/uhyve/issues/7#issuecomment-629103723_

stlankes:
Yes, I guess that an interrupt handler is trying to borrow something, which is already borrowed. I will check it

jschwe:
The reason I'm mentioning this here is because uhyve doesn't exit after the Panic.

stlankes:
Ah, you tested with the current patch?

jschwe:
[...]

PR #48, opened 2020-05-08
call shutdown, if a panic is triggered

stlankes:
This patch should solve #45. @jounathaen Can you test it also on your system?

### DIFF
--- a/src/arch/x86_64/kernel/processor.rs
+++ b/src/arch/x86_64/kernel/processor.rs
@@ -275,35 +275,25 @@ impl CpuFrequency {
 	}
 
 	unsafe fn detect_from_cpuid_tsc_info(&mut self, cpuid: &CpuId) -> Result<(), ()> {
-		match cpuid.get_tsc_info() {
-			Some(tsc_info) => {
-				// check if tsc_info provides a correct value
-				match tsc_info.tsc_frequency() {
-					Some(freq) => {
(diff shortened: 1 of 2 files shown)


jounathaen:
I think, this solves #45 but if executed on "Real Hardware" would result in the Device to shut down and the user not beeing able to read the panic message

src/arch/x86_64/kernel/processor.rs
```diff
@@ -288,6 +290,22 @@ impl CpuFrequency {
 		}
 	}
 
+	unsafe fn detect_from_cpuid_hypervisor_info(&mut self, cpuid: &CpuId) -> Result<(), ()> {
+		match cpuid.get_hypervisor_info() {
```
jounathaen, line 294: This is an `if let Some(...)` pattern

stlankes: Done

ColinFinck:
FYI, I recently discovered an even nicer alternative to such nested `if let` statements - one that doesn't need any nesting at all:

```
let tsc_info = cpuid.get_tsc_info().ok_or(())?;
let freq = tsc_info.tsc_frequency().ok_or(())?;
let mhz = (freq / 1000000u64) as u16;
self.set_detected_cpu_frequency(mhz, CpuFrequencySources::CpuIdTscInfo)
```

PR #49, opened 2020-05-08
Scheduler: Improve Error message when Memory allocation fails

jschwe:
This commit aims to improve the Error message when panicing after Allocation of Memory for the TaskStacks has failed. This should make it clearer to a User, that increasing the amount of memory the Machine has will solve this problem. 
The current panic message is `thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ()', src\arch\x86_64\kernel\scheduler.rs:92:25` which doesn't say much without looking at the Code.

Since `expect` is basically `unwrap` with a custom Error message, this shouldn't have any performance implications.


stlankes:
I totally agree!

PR #50, opened 2020-05-15
Cleanup of pci.rs

tlambertz:
I have refactored pci.rs a bit. Changes:

- handle both IOSpace and 32/64-bit Memory BARs
- add generic function to mm: map(), which maps a physical buffer to a newly allocated virtual address range, used for mmaping memory bars
- add function to memory map the PCI memory bars
- check pci device header-type to ensure we don't read/write garbage, warn on multifunction devices
- remove sys_pci_find_adapter, since it is not used anywhere
    - I am unsure of its original purpose, since I did not find any reference in newlib/hermit-abi. Does it have a current / planned future use?

Previously, virtio mapped the pci bars with the `Cache Disable` page-flag set. Since @stlankes introduced memory barriers, this should no longer be needed.

I have not tested these changes with virtio-net, only virtio-fs. Is there an easy way to also test virtio-net?

As an improvement we could potentially cache the result of `memory_map_bar`, but since virtual address space is huge I feel this is a non-issue. Currently virtio will map a bar of size ~5kb up to 5 times, since its capabilities are all next to each other in the same bar, but mapped separately.

### DIFF
--- a/src/arch/x86_64/kernel/pci.rs
+++ b/src/arch/x86_64/kernel/pci.rs
@@ -1,4 +1,5 @@
 // Copyright (c) 2017 Colin Finck, RWTH Aachen University
+//               2020 Thomas Lambertz, RWTH Aachen University
 //
 // Licensed under the Apache License, Version 2.0, <LICENSE-APACHE or
 // http://apache.org/licenses/LICENSE-2.0> or the MIT license <LICENSE-MIT or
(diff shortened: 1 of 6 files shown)


tlambertz:
I just noticed that the cache-disable might be a mistake. The memory barriers are only for the queues. And we mainly write to the mapped bar on initialization, so this probably doesn't even improve performance much. (We might write to the notification address once each for each buffer transmission). I'll revert that change.

stlankes:
👍

PR #52, opened 2020-05-16
CI: Test libhermit-rs on uhyve with rusty-demo

jschwe:
This commits adds testing of `libhermit-rs` for pushes and PRs. Both debug and release builds are tested with 1 and 2 cores each.
Tests are done via the rusty-hermit demo in rusty-hermit. rusty-hermit is cloned and checked out to the devel branch. The demo is then compiled with the version of libhermit-rs that triggered the build and not the version specified as a submodule in rusty-hermit.

rusty-hermit including libhermit-rs is compiled fresh every time, but cargo dependencies such as uhyve and cargo-download are cached to speed up build time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,54 @@
+language: rust
+rust:
+  - nightly
+os: linux
+dist: bionic
+cache:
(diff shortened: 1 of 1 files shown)


stlankes:
👍

PR #53, opened 2020-05-17
Log: Add option to set LevelFilter at compile time

jschwe:
Adds support to change the LevelFilter of the logger at compile time via enviroment variable `HERMIT_LOG_LEVEL_FILTER`. 
If the Environment variable is not set, or the name doesn't match, then `LevelFilter::Info` is used by default, which is the same as it was before.
The name of the Environment variable is arbitrary of course, so feel free to suggest other names.


stlankes:
Ah, `option_env` is new for my. Great feature!

PR #54, opened 2020-05-20
remove dependency to lazy_static

stlankes:
- requires PR rust-lang/rust#71839
- consequently, the latest nightly compiler has to be used


jounathaen:
Seems like you have to update the `asm!` macro to please the CI

stlankes:
The problem was in the boot loader. We used an old version the x86 crate.

PR #57, opened 2020-05-20
simplify scheduler by using const initialization of static elements

stlankes:


stlankes:
- requires PR rust-lang/rust#71839
- consequently, the latest nightly compiler has to be used

PR #59, opened 2020-05-21
Refactor libhermit-rs for rust edition 2018

jschwe:
This PR consists of two commits. The first addresses various Clippy warning.
Pinging @tlambertz: I changed occurences of `bar` into `pci_bar` or `mem_bar`, since clippy warned that bar is a black listed identifier. Please have a look and tell me if I accidentally broke something, or you want to suggest a different name.

The second commit is to compile the library with rust edition 2018. It mostly adds `crate::`. 

Please have a special look at places with `use x86` since i was unsure if this refers to the external crate or the `crate::x86`. In places where my IDE could resolve the import `use crate:x86::xxx` I chose that, otherwise I left it as `use x86:xxx`

Other than that this should be pretty straightforward.

### DIFF
--- a/src/arch/x86_64/kernel/systemtime.rs
+++ b/src/arch/x86_64/kernel/systemtime.rs
@@ -5,11 +5,11 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use arch::x86_64::kernel::irq;
-use arch::x86_64::kernel::processor;
-use arch::x86_64::kernel::BOOT_INFO;
+use crate::arch::x86_64::kernel::irq;
+use crate::arch::x86_64::kernel::processor;
+use crate::arch::x86_64::kernel::BOOT_INFO;
+use crate::environment;
 use core::sync::atomic::spin_loop_hint;
-use environment;
 use x86::io::*;
 
 const CMOS_COMMAND_PORT: u16 = 0x70;
--- a/src/arch/x86_64/kernel/vga.rs
+++ b/src/arch/x86_64/kernel/vga.rs
@@ -5,8 +5,8 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use arch::x86_64::mm::paging;
-use arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags};
+use crate::arch::x86_64::mm::paging;
+use crate::arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags};
 use x86::shared::io::*;
 
 const CRT_CONTROLLER_ADDRESS_PORT: u16 = 0x3D4;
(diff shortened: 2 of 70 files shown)


src/arch/x86_64/kernel/systemtime.rs
```diff
@@ -5,11 +5,11 @@
...
-use arch::x86_64::kernel::irq;
-use arch::x86_64::kernel::processor;
-use arch::x86_64::kernel::BOOT_INFO;
+use crate::arch::x86_64::kernel::irq;
+use crate::arch::x86_64::kernel::processor;
+use crate::arch::x86_64::kernel::BOOT_INFO;
+use crate::environment;
 use core::sync::atomic::spin_loop_hint;
-use environment;
 use x86::io::*;
```
jschwe, line 13: Have a look here if `use x86::` is correct.

src/arch/x86_64/kernel/vga.rs
```diff
@@ -5,8 +5,8 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-use arch::x86_64::mm::paging;
-use arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags};
+use crate::arch::x86_64::mm::paging;
+use crate::arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags};
 use x86::shared::io::*;
```
jschwe, line 10: Have a look here if use x86:: is correct.

src/arch/x86_64/mm/paging.rs
```diff
@@ -7,20 +7,20 @@
...
+use crate::scheduler;
 use core::marker::PhantomData;
 use core::mem;
 use core::ptr;
-use environment;
-use mm;
 use multiboot::Multiboot;
-use scheduler;
 use x86::controlregs;
 use x86::irq::PageFaultError;
```
jschwe, line 25: Have a look here if use x86:: is correct.

src/syscalls/interfaces/uhyve.rs
```diff
@@ -5,14 +5,14 @@
...
 #[cfg(feature = "newlib")]
-use syscalls::lwip::sys_lwip_get_errno;
+use crate::syscalls::lwip::sys_lwip_get_errno;
 #[cfg(feature = "newlib")]
-use syscalls::{LWIP_FD_BIT, LWIP_LOCK};
+use crate::syscalls::{LWIP_FD_BIT, LWIP_LOCK};
+use core::{mem, ptr, slice};
 
 #[cfg(target_arch = "x86_64")]
 use x86::io::*;
```
jschwe, line 18: Have a look here if use x86:: is correct.

tlambertz:
Changes look good to me :+1:

I have only reviewed the changes to the parts of the code I wrote though.

Seems like 'bar' is blacklisted only because it is often used as foo/bar temporary variable names. Oh well :smile:

stlankes:
Changes look good to me!

PR #60, opened 2020-05-21
WIP: merge all repositories to one

stlankes:
Internally we had many discussions about repositories / branch names
(see for instance hermitcore/libhermit-rs#35). This PR tries to merge
all different repositories/branches to one. If we do this, we have
to archive the RustyHermit repo and link all crates to this repo.
In addition, we have to revise our C examples in HermitPlaygroud.


hermit-sys/build.rs
```diff
@@ -0,0 +1,139 @@
+extern crate walkdir;
+
+use std::env;
+use std::process::Command;
+use walkdir::{DirEntry, WalkDir};
+
+#[cfg(all(not(feature = "rustc-dep-of-std"), not(feature = "with_submodule")))]
+fn build() {
```
jschwe, line 8: Just to clarify: We need this distinction, so that the version of hermit-sys pushed to crates.io can reliably download, build and link libos to hermit-sys. Is my understanding correct?

.github/workflows/format.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - devel
```
jschwe, line 7: I think this would be a good opportunity to also enable testing the format on all branches.

jschwe:
- The loader directory/module is still missing
- Can we talk about what master and devel are going to be used for?
- Can the feature `with_submodule` be enabled/disabled via a commandline option when building? The default should be the same for master/devel. 
- I think it would be best if the feature  `with_submodule` is off by default. Maybe we could automate the release process and have a script turn the feature on before uploading releases to crates.io. I haven't really tinkered with releases on github before, but I assume it can be automated with a workflow or something.

hermit-sys/build.rs
stlankes: Yes, hermit-sys at crates.io is only content of the directors `hermit-sys`. Consequently, a direct access isn't possible, The `hermit-sys` has to download the kernel.

stlankes:
I would remove the branch `devel`.  With cargo-option `--features` you are able specific features in the command line. The feature `with_submodule` is per default off. Only our test applications use it.

netbench/Cargo.toml
```diff
@@ -4,6 +4,7 @@ name = "rust-tcp-io-perf"
 version = "0.0.0"
 authors = ["Lorenzo Martini <lmartini@student.ethz.ch>"]
 readme = "README.md"
+pubslish = false
```
jschwe, line 7: There is a typo in publish.

stlankes: I fixed it

hermit-sys/build.rs
jschwe: Just out of curiosity: I was thinking if it might be possible to add libos as a precompiled library (rlib) to the crates.io package and specifying [an override](https://doc.rust-lang.org/cargo/reference/build-scripts.html#overriding-build-scripts) . This could be done automatically as part of a deploy step and would let us get rid of cargo download. 
In theory the libos rlib should be the same regardless of if it was built on linux, windows or macOS, right? 
I'm not sure if the linking process requires the same compiler version as the one the rlib was built with though.

jschwe:
I think this proposal addresses my issues, and it looks okay to me. I would maybe change the name of the feature `with_submodule` since libos is not a submodule anymore. I'm having trouble thinking of a good name though, maybe: `libos_from_cratesio`, `with_published_libos`.

stlankes:
Yes, the same name `with_submodule` was bad. I replaced it with `with_internal_libos `.

hermit-sys/build.rs
stlankes: Hm, maybe. I don't know how we can deliver an rlib. Currently, we rebuild everything from scratch.

stlankes:
I think about it. Currently, is my opinion to reject it because the integration into my C playground is then not so easy...

jounathaen:
Nice one :ok_hand:! I'm not super happy with the names of the folders yet.
- `img` -> `assets/img`
- `libos` -> `hermit-kernel`
- `demo` and `netbench` -> `util/demo` `util/netbench`
 
And it would be easier to review, if the netbench addition would happen in a separate branch :stuck_out_tongue:

stlankes:
I am not glad with everything in one repo. The kernel has nothing to do with hermit-sys/hermit-abi. In addition, the support of C application will be a pain.

jounathaen:
Another posibillity to "clean up" would be to give `rusty-hermit` the status of the "main repository":
- move most of the README to `rusty-hermit`. Only kernel-related infos remain.
- `libhermit-rs`' README has a link to `rusty-hermit`
- rusty-hermit contains `libhermit-rs` as submodule
- link to `rusty-hermit` in future publications

stlankes:
Good idea, I will reorganize the READMEs

stlankes:
@jschwe @jounathaen I rewrite the README. Does it solve the problem?

jounathaen:
You are talking about https://github.com/hermitcore/libhermit-rs/commit/3da7cfbbbb9931656c3189d754cdce3b5770d2b4 and
https://github.com/hermitcore/rusty-hermit/commit/26827a3cb9a2af0a584f5678f64862eee1f934b0 ?

I think there is still way too much information in `libhermit-rs`, but this is not urgent.
Let's create an Issue to track this and close the PR!

stlankes:
Yes, I close this issue.

PR #63, opened 2020-05-24
simplify the implementation of conditional variables

stlankes:
- adhere more the guideline from Andrew Birrell


stlankes:
we follow a new approach, this version isn't longer necessary

PR #84, opened 2020-06-29
introduce a new interface for smoltcp

stlankes:
- network thread is able to wakeup directly the blocked threads
- reduce the number of context switches


stlankes:
I add kernel patches, because the changes in hermit-sys depends on these changes

PR #90, opened 2020-07-08
reduce the number of interrupts during a send operation

stlankes:
By transferring a packet to Virtio, the driver will be set in polling mode. In this mode, all Virtio interrupts are disabled. Every 20ms, the driver checks, if the user space is still sending packets. If not, the polling mode will be disabled.

### DIFF
--- a/src/arch/x86_64/kernel/virtio.rs
+++ b/src/arch/x86_64/kernel/virtio.rs
@@ -54,9 +54,11 @@ pub mod consts {
 	pub const VIRTQ_DESC_F_WRITE: u16 = 2; // Buffer is device write-only (instead of read-only)
 	pub const VIRTQ_DESC_F_INDIRECT: u16 = 4; // Buffer contains list of virtq_desc
 
-	// The Guest uses this in flag to advise the Host: don't interrupt me
+	// The guest uses this in flag to advise the host: don't interrupt me
 	// when you consume a buffer.
 	pub const VRING_AVAIL_F_NO_INTERRUPT: u16 = 1;
+	/// Default behaviour, where the guest expects interrupts from the host
+	pub const VRING_AVAIL_F_DEFAULT: u16 = 0;
 }
 
 pub struct Virtq<'a> {
@@ -401,7 +403,7 @@ impl<'a> Virtq<'a> {
 		if value {
 			*vqavail.flags = VRING_AVAIL_F_NO_INTERRUPT;
 		} else {
-			*vqavail.flags = 0;
+			*vqavail.flags = VRING_AVAIL_F_DEFAULT;
 		}
 	}
 
--- a/src/arch/x86_64/kernel/virtio.rs
+++ b/src/arch/x86_64/kernel/virtio.rs
@@ -16,13 +16,11 @@ use crate::arch::x86_64::kernel::virtio_net;
 
 use crate::arch::x86_64::mm::paging;
 use crate::config::VIRTIO_MAX_QUEUE_SIZE;
-use crate::synch::spinlock::SpinlockIrqSave;
 
 use alloc::boxed::Box;
 use alloc::rc::Rc;
 use alloc::vec::Vec;
 use core::cell::RefCell;
-use core::cell::UnsafeCell;
 use core::convert::TryInto;
 use core::sync::atomic::spin_loop_hint;
 use core::sync::atomic::{fence, Ordering};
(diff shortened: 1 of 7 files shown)


src/drivers/net/mod.rs
```diff
@@ -5,22 +5,43 @@
...
+fn set_polling_mode(value: bool) {
+	// is the driver already in polling mode?
+	if POLLING.swap(value, Ordering::SeqCst) != value {
+		let irq = irq::nested_disable();
+
+		if let Some(driver) = crate::arch::kernel::pci::get_network_driver() {
+			driver.borrow_mut().set_polling_mode(value);
+		}
+
+		irq::nested_enable(irq);
```
jbreitbart, line 31: Can you use the function taking the closure from the other pull request?

src/drivers/net/mod.rs
```diff
@@ -30,32 +51,64 @@ pub fn netwait_and_wakeup(handles: &[usize], millis: Option<u64>) {
...
 
-	NET_SEM.acquire(millis);
+	let mut reset_nic = false;
+
+	// every 20ms we check if the driver should be in the polling mode
+	while POLLING.swap(false, Ordering::SeqCst) == true {
+		reset_nic = true;
+
+		let core_scheduler = core_scheduler();
+		let wakeup_time = Some(crate::arch::processor::get_timer_ticks() + 20_000);
```
jbreitbart, line 61: Can we make that 20_000 configurable? Or at least move it to a constant for now? It may be interesting to try different values for some workloads just to be sure 20ms is an optimal value. Or is there any specific reason why this has to 20ms?

src/drivers/net/mod.rs
```diff
@@ -30,32 +51,64 @@ pub fn netwait_and_wakeup(handles: &[usize], millis: Option<u64>) {
...
+	}
+
+	if reset_nic {
+		let irq = irq::nested_disable();
+
+		if let Some(driver) = crate::arch::kernel::pci::get_network_driver() {
+			driver.borrow_mut().set_polling_mode(false);
+		}
+
+		irq::nested_enable(irq);
```
jbreitbart, line 76: See above.

src/arch/x86_64/kernel/virtio.rs
```diff
@@ -396,6 +396,15 @@ impl<'a> Virtq<'a> {
 		}
 	}
 
+	pub fn set_polling_mode(&mut self, value: bool) {
+		let mut vqavail = self.avail.borrow_mut();
+		if value == true {
+			*vqavail.flags = VRING_AVAIL_F_NO_INTERRUPT;
```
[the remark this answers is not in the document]
stlankes: I derived the names from [https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.pdf](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.pdf). See section 2.6.6

I agree that flags isn't the right word. But I would prefer the wording from the standard.

src/arch/x86_64/kernel/virtio_net.rs
```diff
@@ -377,6 +377,10 @@ impl<'a> VirtioNetDriver<'a> {
 		false
 	}
 
+	pub fn set_polling_mode(&mut self, value: bool) {
+		(self.vqueues.as_deref_mut().unwrap())[0].set_polling_mode(value);
```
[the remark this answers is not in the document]
stlankes: Yes, I have to use nice names for `0`. `0` represents the input ring buffer, while `1` is the output ring buffer. For the output ring buffer, interrupts are per default deactivated.

src/drivers/net/mod.rs
stlankes: In this case, #91 must be accepted.

src/drivers/net/mod.rs
stlankes: I will do it

src/drivers/net/mod.rs
stlankes: see above :-)

src/drivers/net/mod.rs
stlankes: The last commit solve this problem. Now, I protect the network driver by a lock.

src/arch/x86_64/kernel/virtio.rs
```diff
@@ -396,6 +396,15 @@ impl<'a> Virtq<'a> {
 		}
 	}
 
+	pub fn set_polling_mode(&mut self, value: bool) {
+		let mut vqavail = self.avail.borrow_mut();
+		if value == true {
```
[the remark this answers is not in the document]
stlankes: ok, I will do it

src/arch/x86_64/kernel/virtio.rs
```diff
@@ -396,6 +396,15 @@ impl<'a> Virtq<'a> {
...
+			*vqavail.flags = VRING_AVAIL_F_NO_INTERRUPT;
+		} else {
+			*vqavail.flags = 0;
```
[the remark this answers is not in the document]
stlankes: Yes, not mentioned in the standard, but useful

PR #91, opened 2020-07-08
add helper function, which disables interrupts during the runtime of a closure

stlankes:


jounathaen:
1. I think the name is not expressive enough. I used `exclusive_section` in eduOS.
2. The closure cannot return any values. I used the following in eduOS which is able to:
```rust
	pub fn exclusive_section<F, R>(&self, closure: F) -> R
		where F: Fn (&mut T) -> R
	{
		self.obtain_lock();
		let ret = closure(unsafe {&mut *self.data.get()});
		self.dequeue.fetch_add(1, Ordering::SeqCst);
		ret
	}
```
(This is a bit specific for the spinlock but in principle)

stlankes:
Why is the name not expressive enough? I don't want to take lock. I want only to disable interrupts.

jounathaen:
My bad, I didn't looked close enough. Then it's fine. But still `irqsave` sounds a little bit cryptic. Something like `interrupt_free` could be more readable.

PR #92, opened 2020-07-12
separate virtual from physical addresses

stlankes:
- the logical address space is completly separate from the physical
- consequently, different names are used

### DIFF
--- a/src/arch/x86_64/kernel/acpi.rs
+++ b/src/arch/x86_64/kernel/acpi.rs
@@ -7,19 +7,20 @@
 
 use crate::arch::x86_64::mm::paging::{BasePageSize, PageSize, PageTableEntryFlags};
 use crate::arch::x86_64::mm::{paging, virtualmem};
+use crate::arch::x86_64::mm::{PhysAddr, VirtAddr};
 use crate::x86::io::*;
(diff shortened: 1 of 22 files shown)


src/arch/x86_64/kernel/acpi.rs
```diff
@@ -109,13 +110,13 @@ impl<'a> AcpiTable<'a> {
...
-		let physical_map_address = align_down!(physical_address, BasePageSize::SIZE);
-		let offset = physical_address - physical_map_address;
+		let physical_map_address = physical_address.align_down_to_base_page();
+		let offset: usize = (physical_address - physical_map_address).into();
 		let mut virtual_address = virtualmem::allocate(allocated_length).unwrap();
 		paging::map::<BasePageSize>(virtual_address, physical_map_address, count, flags);
 
 		// Get a pointer to the header and query the table length.
-		let mut header_ptr = (virtual_address + offset) as *const AcpiSdtHeader;
+		let mut header_ptr: *const AcpiSdtHeader = (virtual_address + offset).as_ptr();
```
jounathaen, line 119: Not 100% part of the feature, but this is no real issue.

PR #93, opened 2020-08-21
Integration tests

jschwe:
This PR aims to add integration tests to rusty-hermit. This PR consists mostly of the framework necessary, and only contains some basic tests, which realistically don't have a very high quality.  None of the rusty-demo tests have been moved here either (yet). 

The PR mainly adds hermit_test_runner.py which can run the integration tests built by `cargo test --target [...]` either on uhyve or on qemu. 

There are still a couple of open ToDos and points that might need some discussion / rework, so I'm opening this as a draft. I'll also improve the Description of the PR next week when I have a bit more time.

### Running Tests

The integration tests can executed by calling:
`cargo test --tests --no-fail-fast -Z build-std=core,alloc --target x86_64-unknown-hermit-kernel [-- [Test Runner arguments]]`
Valid Test Runner arguments currently include:
- `--bootloader_path=some/path`: This implicitly uses qemu to execute the tests. If absent uhyve will be used.
- `--uhyve_path=some/path`: Optional, use if `uhyve` is not in PATH
- `-v`: Optional, always prints the stdout/stderr of the test, regardless of test failure
- `-vv`: Optional, In addition to `-v` this also runs the test in verbose mode (i.e. `uhyve -v`)
- `--gdb`: Enables gdb on port 1234 and stops the test executable at the hermit entrypoint.
- `--num_cores`: Controls the number of CPU cores the test runner uses. Defaults to 1.

This implicitly also executes the unit-tests, however the testrunner currently skips the unit-tests with a warning, since they are currently broken and result in a bootloop. 

### Writing Tests

Integration tests are written in the top-level of the tests folder. The common `mod` provides shared functionality that the tests can `use`. This includes an exit function, that all tests should use to exit, since it ensures that an exit code is passed to the runner regardless of whether qemu or uhyve is used. If you want to use a test_runner you can use `#[test_case]` 

## Existing Tests (Host)

The existing Unit-tests can still be run on the host OS via: `cargo test --lib`. In the future the unit-tests also should be able to run on hermit, but this is currently broken, due to a bootloop, directly after entering hermit.

## Demo Tests

I added some basic tests, which demonstrate how a test could look like.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,14 @@ travis-ci = { repository = "hermitcore/libhermit-rs" }
 crate-type = ["staticlib"]
 name = "hermit"
 
+[[test]]
+name = "basic_print"
+harness = false
+
+[[test]]
+name = "basic_math"
+harness = false
+
 [features]
 default = ["pci", "acpi"]
 vga = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,10 @@ acpi = []
 [dev-dependencies]
 x86_64 = "0.11.0"
 
+[dev-dependencies.num-traits]
+version = "0.2.12"
+default-features = false
+
 [dependencies]
 bitflags = "1.2"
 #cfg-if = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,10 @@ acpi = []
 
 [dev-dependencies]
 x86_64 = "0.11.0"
+float-cmp = "0.8.0"
 
 [dev-dependencies.num-traits]
-version = "0.2.12"
+version = "0.2"
 default-features = false
 
 [dependencies]
@@ -54,7 +55,7 @@ bitflags = "1.2"
 num-derive = "0.3"
 
 [dependencies.num]
-version = "0.2"
+version = "0.3"
 default-features = false
 
 [dependencies.num-traits]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ float-cmp = "0.8.0"
 [dev-dependencies.num-traits]
 version = "0.2"
 default-features = false
+# features = ["i128"] // Blocked by: https://github.com/rust-num/num-traits/issues/177
 
 [dependencies]
 bitflags = "1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,9 @@ newlib = []
 pci = []
 acpi = []
 
+[dev-dependencies]
+x86_64 = "0.11.0"
+
 [dependencies]
 bitflags = "1.2"
 #cfg-if = "0.1"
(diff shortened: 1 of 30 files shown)


Cargo.toml
```diff
@@ -62,11 +83,12 @@ default-features = false
...
 opt-level = 1      # controls the `--opt-level` the compiler builds with
 debug = true       # controls whether the compiler passes `-C debuginfo`
-                   # a value of `true` is equivalent to `2`
+# a value of `true` is equivalent to `2`
 rpath = false      # controls whether the compiler passes `-C rpath`
 lto = false         # controls `-C lto` for binaries and staticlibs
 debug-assertions = true # controls whether debug assertions are enabled
-panic = "abort" # Call abort on panic https://github.com/rust-lang/rust/pull/32900
+# Disabled because of https://github.com/rust-lang/cargo/issues/7359
+#panic = "abort" # Call abort on panic https://github.com/rust-lang/rust/pull/32900
```
jschwe, line 91: Disabling this doesn't seem to have any negative side effects, but I wanted to point this out in case I'm overlooking something.

src/arch/x86_64/kernel/start.rs
```diff
@@ -33,6 +36,7 @@ pub unsafe extern "C" fn _start(boot_info: &'static mut BootInfo) -> ! {
 }
 
 unsafe fn pre_init(boot_info: &'static mut BootInfo) -> ! {
+	assert_eq!(boot_info.magic_number, BOOT_IMAGE_MAGIC_NUMBER);
```
jschwe, line 39: Since the Unit-tests have garbage in boot_info, I thought it would be reasonable to add an assertion here that checks the magic number.

src/lib.rs
```diff
@@ -81,20 +88,52 @@ mod collections;
 mod config;
 mod console;
 mod drivers;
-mod environment;
+pub mod environment;
```
jschwe, line 91: This needs to be public, so the test can check if it is running on uhyve when exiting. Otherwise it assumes it's on qemu and attempts to exit via the debug device. If that fails, the normal `sys_exit` is used.

src/macros.rs
```diff
@@ -22,14 +22,15 @@ macro_rules! align_up {
 ///
 /// From http://blog.phil-opp.com/rust-os/printing-to-screen.html, but tweaked
 /// for HermitCore.
+#[macro_export]
 macro_rules! print {
 	($($arg:tt)+) => ({
-		use core::fmt::Write;
-		$crate::console::CONSOLE.lock().write_fmt(format_args!($($arg)+)).unwrap();
+        $crate::_print(format_args!($($arg)*));
```
jschwe, line 28: Export the print macro, and a print function, so that the integration tests can access it.

stlankes:
@jschwe I added a workaround for uyhve. However, the number of environment variables is still limited. Does it work for you.

stlankes (approved):

Issue #94, opened 2020-08-30
Compiling fails with segfault


jschwe:
Compiling libhermit-rs currently fails with a segmentation fault. This behaviour was introduced with the `nightly-2020-08-24`. This is likely to be related to the following issue: https://github.com/rust-lang/rust/issues/75922. 
The comments there suggest that we will have to make changes on our end to the `#[naked]` functions.

stlankes:
Yes! Hm, I also create an issue rust-lang/rust#76061. I will look into rust-lang/rust#75922.

tlambertz:
I had a similar issue with naked functions recently (though this were runtime issues), where naked functions with parameters behaved weirdly. https://github.com/rust-lang/rust/pull/75417 fixed some of that, but people are unsure if naked+parameters should be supported or not. My workaround was using `global_asm!` and declaring the symbol as a `extern 'C' fn`.

stlankes:
Yes, seems to be the better solution

stlankes:
I close this issue. PR #95 solved it.

PR #95, opened 2020-08-30
remove dependencies to rust feature "naked functions"

stlankes:
- naked functions does currently not work (rust-lang/rust#75922)
- we replaced naked functions by assembly code
- using global_asm to include the assembly code


stlankes:
This PR should solve issue #94

Issue #96, opened 2020-08-31
Spinlock is Unsound


tlambertz:
RustyHermits `Spinlock<T>` implementation is unsound, since it always implements `Sync + Send`, no mater what `T` implements. This behavior was introduced in https://github.com/hermitcore/libhermit-rs/commit/cf4e955ebd5f5fda70f665f4fdb70484e4c04a16.

Contrary to the commit message, I think this behavior is unsound, since it allows the following to compile:
```rust
pub static GLOBAL_WHICH_IS_SEND_AND_SYNC: Option<Spinlock<Rc<u64>>> = None;
```
This is just an example, but there are ways to construct a non-oneliner of this without the `Option<>`. Now consider the following:
- Thread A unlocks the Spinlock, calls `Rc.clone()`, locks Spinlock again
- Thread B does the same
- now both Threads have a reference to 'the same' `Rc`, which is explicitly neither Send nor Sync (https://doc.rust-lang.org/std/rc/struct.Rc.html#impl-Send), so should not be able to exist across threads. Using it will now lead to race conditions.

Reverting the mentioned commit introduces several issues:
```
53 | static PCI_DRIVERS: SpinlockIrqSave<Vec<PciDriver>> = SpinlockIrqSave::new(Vec::new());                                                                                                                                       
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `alloc::rc::Rc<core::cell::RefCell<arch::x86_64::kernel::virtio_net::VirtioNetDriver>>` cannot be sent between threads safely         

19 | static PHYSICAL_FREE_LIST: SpinlockIrqSave<FreeList> = SpinlockIrqSave::new(FreeList::new());                                                                                                                                 
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `alloc::rc::Rc<core::cell::RefCell<collections::doublylinkedlist::Node<mm::freelist::FreeListEntry>>>` cannot be sent between th

14 | static KERNEL_FREE_LIST: SpinlockIrqSave<FreeList> = SpinlockIrqSave::new(FreeList::new());
[...]

PR #97, opened 2020-09-01
add dependency to `Send` for the implementation of spinlock

stlankes:
- first draft to solve issue #96


stlankes:
@tlambertz I revise the interface to the filesystem. Does it meet your requirements?

src/drivers/net/mod.rs
```diff
@@ -25,11 +24,9 @@ const POLL_PERIOD: u64 = 20_000;
 fn set_polling_mode(value: bool) {
 	// is the driver already in polling mode?
 	if POLLING.swap(value, Ordering::SeqCst) != value {
-		irqsave(|| {
-			if let Some(driver) = crate::arch::kernel::pci::get_network_driver() {
-				driver.borrow_mut().set_polling_mode(value);
-			}
-		});
+		if let Some(driver) = crate::arch::kernel::pci::get_network_driver() {
+			driver.lock().set_polling_mode(value);
```
tlambertz, line 28: This might deadlock. `virtio_irqhandler` also gets the network driver and calls
```rust
driver.lock().handle_interrupt()
```

The interrupt might happen while the driver is locked here.

src/syscalls/interfaces/mod.rs
```diff
@@ -117,55 +116,55 @@ pub trait SyscallInterface: Send + Sync {
...
 			_ => false,
-		})
+		}
 	}
 
 	fn get_tx_buffer(&self, len: usize) -> Result<(*mut u8, usize), ()> {
-		irqsave(|| match arch::kernel::pci::get_network_driver() {
-			Some(driver) => driver.borrow_mut().get_tx_buffer(len),
+		match arch::kernel::pci::get_network_driver() {
+			Some(driver) => driver.lock().get_tx_buffer(len),
```
tlambertz, line 141: On all of these driver.lock() calls you have the same issue as above that deadlocks might happen. A lot more likely to occur here, since these calls might happen frequently.

tlambertz (approved):
I have only taken a cursory glance and found virtio-net vulnerable to deadlocks. Though I think similar issues would have caused a panic before. My own rewrite of virtio-fs uses a different interrupt logic, that might be worth adopting. I'll merge it next week.

src/drivers/net/mod.rs
stlankes: That is the reason for SpinlockIrqSave. If a task hold this lock, all interrupt are blocked. => No interrupt during the critical section

src/syscalls/interfaces/mod.rs
stlankes: see previous comment

src/drivers/net/mod.rs
stlankes: You find more details at https://github.com/hermitcore/libhermit-rs/blob/master/src/synch/spinlock.rs#L212-L221

PR #98, opened 2020-09-11
syscall to set the seed of the random number generator

stlankes:
- using the core library to create a random number
- removing inline assembly


src/arch/x86_64/kernel/processor.rs
```diff
@@ -878,14 +878,18 @@ fn print_cpu_information() {
 }*/
 
 pub fn generate_random_number() -> Option<u32> {
-	if unsafe { SUPPORTS_RDRAND } {
-		let value: u32;
-		unsafe {
-			llvm_asm!("rdrand $0" : "=r"(value) ::: "volatile");
+	unsafe {
+		if SUPPORTS_RDRAND {
```
jbreitbart, line 882: Is there a log message in case we have fall back to pseudo random values?

src/arch/x86_64/kernel/processor.rs
```diff
@@ -878,14 +878,18 @@ fn print_cpu_information() {
 }*/
 
 pub fn generate_random_number() -> Option<u32> {
```
jbreitbart, line 880: Is generating u32 values fine? I did not check the interfaces, but how are 64bit random values generated?

jbreitbart (commented):
Related to https://github.com/hermitcore/rusty-hermit/issues/63

src/arch/x86_64/kernel/processor.rs
stlankes: I added also support for 64 numbers (see latest changes)

src/arch/x86_64/kernel/processor.rs
stlankes: I added a debug message in `src/syscalls/random.rs`

jbreitbart: Are these printed every time a random number is generated? Can we print an info message at startup saying we are using either the hardware random generator or park miller lehmer software random generator?

src/syscalls/random.rs
```diff
@@ -17,19 +17,58 @@ fn generate_park_miller_lehmer_random_number() -> u32 {
 	random
 }
 
-fn __sys_rand() -> u32 {
-	if let Some(value) = arch::processor::generate_random_number() {
+fn __sys_rand32() -> u32 {
+	if let Some(value) = arch::processor::generate_random_number32() {
 		value
 	} else {
+		debug!("Fallback to a software-based random number generator");
```
jbreitbart, line 24: You have missed this one :smile:

jbreitbart (changes requested):
Feel free to merge after you have removed the superfluous `debug!`

src/syscalls/random.rs
stlankes: Yeah, I fixed it

PR #99, opened 2020-09-11
define default target, using build-std to define the dependencies

stlankes:
After the definition of the default target and the definition of the dependencies, the crate can be build with standard command `crate build`.


stlankes:
Doesn't work in combination with unittests

PR #103, opened 2020-09-13
build per default the target x86_64-unknown-hermit-kernel

stlankes:
afterwards `cargo build` without any additional flag builds libhermit-rs for x86_64


stlankes:
bors ping

stlankes:
I delete already the source. My mistake

Issue #105, opened 2020-09-18
Scheduler does not wake sleeping threads when multiple cores are available


tlambertz:
When using multiple CPU cores, the scheduler does not wake sleeping threads once their sleep duration is over. I guess this happens because of the core-local scheduling, where the main core does not check for threads to be woken on other cores. But I have not verified this.

Simple reproducer:
```
#[cfg(target_os = "hermit")]
extern crate hermit_sys;

use std::thread;
use std::time;

fn main() {
    let mut children = vec![];

    let threadnum = 5;
    for i in 0..threadnum {
        println!("SPAWNING THREAD {}", i);
        children.push(thread::spawn(move || {
            println!("this is thread number {}", i);
            thread::sleep(time::Duration::from_millis(2000));
            println!("---------------THREAD DONE!---------- {}", i);
        }));
    }
    println!("SPAWNED THREADS");

    for child in children {
        let _ = child.join();
    }
}
```

My test project includes a cargo project, makefile and the rusty-loader: [rusty-test-sleep.tar.gz](https://github.com/hermitcore/libhermit-rs/files/5248158/rusty-test-sleep.tar.gz).
The binary is build and started with `make run`. This also sets the correct QEMU options to reproduce the bug.

stlankes:
yeah, I think that this is the reason for issue #78

stlankes:
@tlambertz I think that I solved with #106 the issue. Unfortunately, #78 has another reason.

tlambertz:
Yep, I can confirm the issue is fixed.

PR #106, opened 2020-09-19
avoid deadlock when multiple cores are available

stlankes:
- PR should solve issues #105
- add integration test to show this behavior
- make sure that the task is always added to the list of blocked task => avoid "loosing" of tasks

### DIFF
--- a/src/scheduler/task.rs
+++ b/src/scheduler/task.rs
@@ -468,15 +468,21 @@ impl Task {
 	}
 }
 
-/*impl Drop for Task {
+impl Drop for Task {
 	fn drop(&mut self) {
 		debug!("Drop task {}", self.id);
 	}
-}*/
+}
 
 struct BlockedTask {
 	task: Rc<RefCell<Task>>,
 	wakeup_time: Option<u64>,
+}
+
+impl BlockedTask {
+	pub fn new(task: Rc<RefCell<Task>>, wakeup_time: Option<u64>) -> Self {
+		Self { task, wakeup_time }
+	}
 }
 
 pub struct BlockedTaskQueue {
@@ -535,7 +541,7 @@ impl BlockedTaskQueue {
 			borrowed.status = TaskStatus::TaskBlocked;
 		}
 
-		let new_node = BlockedTask { task, wakeup_time };
+		let new_node = BlockedTask::new(task.clone(), wakeup_time);
 
 		// Shall the task automatically be woken up after a certain time?
 		if let Some(wt) = wakeup_time {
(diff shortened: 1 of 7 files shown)


stlankes:
@tlambertz The patch works on my test system. Does it work on you system? It was a stupid issues.

stlankes:
bors try

src/scheduler/task.rs
```diff
@@ -559,6 +565,15 @@ impl BlockedTaskQueue {
...
 			}
+
+			// No, then just insert it at the end of the list.
+			self.list.push_back(new_node);
+
+			// If this is the new first task in the list, update the One-Shot Timer
+			// to fire when this task shall be woken up.
+			if first_task {
+				arch::set_oneshot_timer(wakeup_time);
+			}
```
jbreitbart, line 576: Can you use scopeguards for code like this? https://docs.rs/scopeguard/1.1.0/scopeguard/ This would also get rid of the duplicated code and prevent the same issue from reappearing after a change to the function. I did not look into it in detail, not sure if it is really a good solution.

src/scheduler/task.rs
```diff
@@ -559,6 +565,15 @@ impl BlockedTaskQueue {
 				first_task = false;
 				cursor.move_next();
 			}
+
+			// No, then just insert it at the end of the list.
+			self.list.push_back(new_node);
+
+			// If this is the new first task in the list, update the One-Shot Timer
+			// to fire when this task shall be woken up.
```
jbreitbart, line 573: Can you change the comment so it is clear that you are looking at the task with the earliest deadline?

stlankes: yes, see below

src/scheduler/task.rs
```diff
@@ -468,17 +468,23 @@ impl Task {
 	}
 }
 
-/*impl Drop for Task {
+impl Drop for Task {
 	fn drop(&mut self) {
 		debug!("Drop task {}", self.id);
 	}
-}*/
+}
```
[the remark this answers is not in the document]
stlankes: You are right, I will revert this part.

src/scheduler/task.rs
stlankes: Good idea, I integrate a scopeguard

src/scheduler/task.rs
```diff
@@ -6,6 +6,8 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+#![allow(unused_assignments)]
```
jbreitbart, line 9: Why do you need this? Can you move it to he scope/line where you want to turn off the check?

stlankes: You are right, I fixed it

PR #108, opened 2020-09-27
revise allocator interface to support the latest version of libstd

stlankes:
- revise allocator interface
- reenable tests on macOS


stlankes:
Changes are minimal and required for the latest libstd. If all tests passed, I will accept the PR.

PR #109, opened 2020-09-29
port libhermit to support C-based applications

stlankes:
- using of VirtAddr also for newlib-based applications

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,69 @@
+name: Docker tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - staging
+      - trying
+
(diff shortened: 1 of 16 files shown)


tlambertz (approved):
I don't use cmake to build my applications, so I can't check the validity of all changes. The x86 hermit-kernel json target file should not be needed anymore, since the target is contained in upstream rustc?

The other changes are identical to mine and work.

stlankes:
Yes, maybe the changes aren't important. In the json-files, I define that we use a GNU linker. In case of Rust, we use `rust-ldd`. Maybe they're no differences between the linkers.

stlankes:
bors try

stlankes:
bors try

Issue #107, opened 2020-09-23
Use of newlib's memcpy/memmove functions in the kernel is unstable.


tlambertz:
When building the kernel with newlib, `rlib.rs` is not compiled. This usually provides implementation for `memcpy`, `memmove` and the like. Instead, the implementations provided by newlib are used.

This is an issue, since newlib is not build for the kernel target, but a user one. Newlib is allowed to use the floating point unit, and as such `memmove` uses SSE. Using SSE in the kernel-context is an obviously bad idea.

I can somewhat reliably reproduce a `BorrowMutError` in my filesystem benchmark code:
```
[0][!!!PANIC!!!] src/scheduler/mod.rs:326: already borrowed: BorrowMutError
```

I am not entirely sure what causes this, but have tracked it down to the SSE usage inside `memmove`. Maybe something like:
- userspace thread accesses FPU for the first time, exception is thrown and `fpu_switch()` is called.
- `fpu_switch()` borrows current owner, switches state to new owner
- `fpu_switch()` does `self.current_task.clone()`, which under certain circumstances calls `memmove`.
    - `memmove` again uses the FPU, causing another call to `fpu_switch()`, and a double-borrow occurs -> `panic()`

### Log:
I am running 6 threads on a single core. First `[ ]` column contains `rdtsc` value. Unfortunately, adding more logging makes the bug disappear.

```
[0x112aeaebc][0][INFO] FPU Switch caused by PC 0x807b1b (userspace uses `fild    qword [rbp-0x8]`)
[0x112ce00e4][0][INFO] Switching FPU owner from task 6 to 7
[0x112eda4e4][0][INFO] FPU Switch caused by PC 0x807b1b (userspace uses `fild    qword [rbp-0x8]`)
[0x1130a30c0][0][INFO] Switching FPU owner from task 7 to 5
[0x11329b468][0][INFO] FPU Switch caused by PC 0x807b1b (userspace uses `fild    qword [rbp-0x8]`)
[0x11350c854][0][INFO] Switching FPU owner from task 5 to 4
[0x1137a506c][0][INFO] FPU Switch caused by PC 0x807b1b (userspace uses `fild    qword [rbp-0x8]`)
[0x113a09900][0][INFO] Switching FPU owner from task 4 to 6
[0x113c9f9b8][0][INFO] FPU Switch caused by PC 0x807b1b (userspace uses `fild    qword [rbp-0x8]`)
[...]

PR #110, opened 2020-09-29
Various newlib fixes

tlambertz:
The current kernel does not build with `feature=newlib`. This pull request fixes various issues:

1. Kernel Heap Size is increased from 2MB to 20MB.
2. Rust userland reserves 10% of available memory for task stacks. This is previously not done with newlib, leading to crashes when many threads are spawned. This is fixed, and newlib now also reserves 10% of memory.
3. Some misc. compiler errors are fixed (extra `;`, missing conversion from VirtAddr to usize.
4. Newlib now uses the kernel-builtin (non-optimized, non SSE) `memcpy`, `memmove` etc. See issue #107 for a detailed reasoning of this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -95,7 +95,7 @@ pub mod environment;
 mod errno;
 mod kernel_message_buffer;
 mod mm;
-#[cfg(not(feature = "newlib"))]
+
 mod rlib;
 #[cfg(target_os = "hermit")]
 mod runtime_glue;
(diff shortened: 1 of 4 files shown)


stlankes:
@tlambertz Can you review #109? It is a little bit more than 93f63b4.

stlankes:
@tlambertz Do you agree to close the PR? The last open commit is merged by #127

tlambertz:
Sure, the merged changes look good.

PR #114, opened 2020-10-14
introduce __rg_oom, which is called by libstd due a oom error

stlankes:
These changes are required to support the latest nightly compiler.


stlankes:
Patch is required to support the latest nightly compiler. => accept it

PR #115, opened 2020-10-27
using compiler builtins to realize basic operations (e.g. memcpy)

stlankes:


stlankes:
I enable the nightly tests because rust-lang/rust#77610 is merged to the nightly compiler.

stlankes:
bors try

jbreitbart (commented):
The file changes themselves look fine, but I am wondering if / how this changes the performance. Have you measured it?

According to https://github.com/rust-lang/compiler-builtins/issues/339 the builtins are not (yet?) optimized, however there was a PR already merged for x86_64.

stlankes:
Do I promise better performance? :-) This is at least better than my implementation.

I am separating the operations between user- and kernel space. This will be the next PR in RustyHermit and is more critical for the performance.

stlankes:
@jbreitbart Maybe rust-lang/compiler-builtins#365 is interesting for you. I try to support it in the next pull request.

PR #116, opened 2020-11-08
build __rg_oom only for pure Rust applications

stlankes:
- avoids linker errors by building C applications
- merge PR with PR hermitcore/libhermit-rs#110

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,6 +1,6 @@
 [unstable]
 build-std = ["core", "alloc"]
-build-std-features = ["compiler-builtins-mem"]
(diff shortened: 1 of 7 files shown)


jbreitbart (approved):
The changes themselves look fine for me, but I think comments that explain the memory layout would be helpful to understand what you are actually doing in the init function 😉 

This is more of a general question, but why do you use 10% as a fixed value for the amount of memory that is reserved for the stack? Is 10% guaranteed to be big enough to allocate the stacks required for at least 1 task?

stlankes:
I revise my changes and create smaller PR

PR #117, opened 2020-11-22
fix broken CI

stlankes:
- GitHub deprecate the set-env and add-path commands
- use the [guideline](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files) to solve this issue


stlankes:
I accept these changes because they are backports from rusty-hermit.

PR #118, opened 2020-11-22
using mangled names for basic opertions to avoid linker conflicts

stlankes:
avoiding collisions of symbols by using mangled names


stlankes:
Doesn't work, I close this PR.

PR #119, opened 2020-11-22
simplify function to determine the enviornment

stlankes:
- avoid using of __sys_malloc
- remove unneeded unsafe code


src/syscalls/interfaces/uhyve.rs
```diff
@@ -215,78 +216,41 @@ impl SyscallInterface for Uhyve {
...
+		uhyve_send(UHYVE_PORT_CMDSIZE, &mut syscmdsize);
+
+		// create array to receive all arguments
+		let mut argv = Box::new(vec![ptr::null(); syscmdsize.argc as usize]);
+		let mut argv_phy = vec![ptr::null(); syscmdsize.argc as usize];
+		for i in 0..syscmdsize.argc as usize {
+			argv[i] = crate::__sys_malloc(syscmdsize.argsz[i] as usize * mem::size_of::<u8>(), 1);
+			argv_phy[i] =
+				paging::virtual_to_physical(VirtAddr(argv[i] as u64)).as_u64() as *const u8;
+		}
```
jbreitbart, line 230: How about to create the vectors with `std::Vec::with_capacity instead of filling them first with null pointers and than changing the values?

src/syscalls/interfaces/uhyve.rs
```diff
@@ -215,78 +216,41 @@ impl SyscallInterface for Uhyve {
...
-				syscmdsize.argc,
-				argv_raw as *const *const u8,
-				env_raw as *const *const u8,
-			)
+		// create array to receive the environment
+		let mut env = Box::new(vec![ptr::null(); syscmdsize.envc as usize + 1]);
+		let mut env_phy = vec![ptr::null(); syscmdsize.envc as usize + 1];
+		for i in 0..syscmdsize.envc as usize {
+			env[i] = crate::__sys_malloc(syscmdsize.envsz[i] as usize * mem::size_of::<u8>(), 1);
+			env_phy[i] = paging::virtual_to_physical(VirtAddr(env[i] as u64)).as_u64() as *const u8;
```
jbreitbart, line 237: Same as above

src/syscalls/interfaces/uhyve.rs
```diff
@@ -215,78 +216,41 @@ impl SyscallInterface for Uhyve {
...
 		}
+
+		// ask uhyve for the environment
+		let mut syscmdval = SysCmdval::new(
+			VirtAddr(argv_phy.as_ptr() as u64),
+			VirtAddr(env_phy.as_ptr() as u64),
+		);
+		uhyve_send(UHYVE_PORT_CMDVAL, &mut syscmdval);
```
jbreitbart, line 245: I guess this is a blocking send and it is ok that `argv_phy` and `env_phy` are destructed when we return from this function? Is `syscmdval` modified or why is it passed as `&mut`?

jbreitbart (changes requested):

src/syscalls/interfaces/uhyve.rs
stlankes: The hypervisor is able to modify the record. The compiler should be aware that the data can be modified.

src/syscalls/interfaces/uhyve.rs
stlankes: That is true. I fixed it.

src/syscalls/interfaces/uhyve.rs
stlankes: That is true. I fixed it.

src/syscalls/interfaces/uhyve.rs
jbreitbart: So it is more like a send_receive function?

stlankes: Yes, it is a hypercall to the VM.

jbreitbart: Ah, ok. I think the naming is not perfect in that case. Also I am wondering if &mut is fine in case the hypervisor can modify the data, should it not be marked as volatile? But let's discuss this in slack 😉 , I'll approve this PR.

jbreitbart (approved):
Looks good after you fix the format problem

PR #123, opened 2020-11-23
increasing the readability by using suitable function names

stlankes:


src/syscalls/interfaces/uhyve.rs
```diff
@@ -191,24 +191,24 @@ pub struct Uhyve;
 
 impl SyscallInterface for Uhyve {
 	fn open(&self, name: *const u8, flags: i32, mode: i32) -> i32 {
-		let mut sysopen = SysOpen::new(VirtAddr(name as u64), flags, mode);
-		uhyve_send(UHYVE_PORT_OPEN, &mut sysopen);
+		let sysopen = SysOpen::new(VirtAddr(name as u64), flags, mode);
+		hypercall(UHYVE_PORT_OPEN, &sysopen);
 
-		sysopen.ret
+		unsafe { ptr::read_unaligned(&sysopen.ret) }
```
jbreitbart, line 197: I know we started discussing this before, by I think we have not finished that discussion, let me know if I am wrong. :smile: 

Do you not have to use a volatile read aka `std::ptr::read_volatile` here? If I understood this correctly, sysopen.ret will be modified by uhyve and not by the code running within the VM. IIRC in C volatile is required if data is modified outside of the program flow of the application, which would be true here. Rust seems not to have a final definition of what volatile is and when it is required, but my understanding of `read_unaligned` is that the compiler may very well optimize it.

jbreitbart: BTW: I am not adding this comment to every `read_unaligned`, but I think it applies to all of them.

src/syscalls/interfaces/uhyve.rs
```diff
@@ -218,39 +218,44 @@ impl SyscallInterface for Uhyve {
 	#[cfg(target_os = "hermit")]
 	fn get_application_parameters(&self) -> (i32, *const *const u8, *const *const u8) {
 		// determine the number of arguments and environment variables
-		let mut syscmdsize = SysCmdsize::new();
-		uhyve_send(UHYVE_PORT_CMDSIZE, &mut syscmdsize);
+		let syscmdsize = SysCmdsize::new();
+		hypercall(UHYVE_PORT_CMDSIZE, &syscmdsize);
+
+		let argc = unsafe { ptr::read_unaligned(&syscmdsize.argc) as usize };
+		let envc = unsafe { ptr::read_unaligned(&syscmdsize.envc) as usize };
```
jbreitbart, line 225: I know I have been nagging about the `&mut`, but I am not sure if I like this solution either. Can you move the allocation of the memory inside `hypercall` and just return it from the call?

jbreitbart (changes requested):

src/syscalls/interfaces/uhyve.rs
stlankes: The data is unaligned. If you use read_volatile, the compiler assumes alignment and optimized the code. In our case, the system crashes because the assumption is in our case wrong.

src/syscalls/interfaces/uhyve.rs
stlankes: Hm, in this case you create a function for every possible hypercall. Do think this is more understandable?

src/syscalls/interfaces/uhyve.rs
jbreitbart: But why do you think you can go without a volatile? That fact that there is no unaligned  volatile read does not mean that volatile is optional 😄

src/syscalls/interfaces/uhyve.rs
jbreitbart: How about returning an enum that fits all needs?

src/syscalls/interfaces/uhyve.rs
stlankes: I checked it and take a look in the [code](http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/src/core/ptr.rs.html#453-455).

stlankes: Ok, you are right. I fixed it with the last patch.

PR #128, opened 2020-11-30
revision of the memory subsystem initialization

stlankes:
Especially with a C-base user space the subsystem doesn't align GDT/IDT correctly. In addition, the initial page table wasn't correctly initialized and triggers sometimes page faults.


stlankes:
I merged Jens' changes. He seems to agree that this PR could be merged.

Issue #78, opened 2020-06-05
SMP race condition


jschwe:
I don't think we have an issue for this yet, so I'm opening one now to track this issue. 
Currently there are issues running rusty-demo with multiple cores, where sometimes it takes a long time to complete the demo.

## Jobs failed due to SMP timeout (May 20 - June 06):
- https://github.com/hermitcore/libhermit-rs/runs/726877033
- https://github.com/hermitcore/libhermit-rs/runs/725684756
- https://github.com/hermitcore/libhermit-rs/runs/725658934
- https://github.com/hermitcore/libhermit-rs/runs/696784822
- https://github.com/hermitcore/libhermit-rs/runs/695711886
- https://github.com/hermitcore/rusty-hermit/runs/732583279

## Possibly related Job logs:
- [```thread 'main' panicked at 'Trying to get the scheduler for core 1, but it isn't available', src/scheduler/mod.rs:530:9```](https://travis-ci.com/github/hermitcore/rusty-hermit/jobs/343843527#L976) for https://github.com/hermitcore/rusty-hermit/commit/64ec710f8a683eeb0a5a6d3baa591fad794e4fe5

stlankes:
I found the race to you log and fix it with 29452bb, but their is still a race in our code.

jschwe:
I'm bumping this issue, with an update.
When running rusty-demo via ` qemu-system-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,fxsr         -display none -smp 1 -m 1G -serial stdio  -kernel loader/target/x86_64-unknown-hermit-loader/debug/rusty-loader  -initrd target/x86_64-unknown-hermit/debug/rusty_demo > log.log` with smp values 1,2,3 and 4, and ` HERMIT_LOG_LEVEL_FILTER=Debug`  The log files get increasingly larger:
1 Core: 15k lines
2 Cores: 37k lines
3 Cores:  86k lines
4 Cores:  135k lines

It looks like the tasks on the additional cpus get blocked very often. I'm not exactly sure in which part of rusty-demo this is, but this looks like it could be some sort of a synchronization problem. 
[log_single_core.log](https://github.com/hermitcore/libhermit-rs/files/5647260/log_single_core.log)
[log_2_core.log](https://github.com/hermitcore/libhermit-rs/files/5647261/log_2_core.log)
[log_3_core.log](https://github.com/hermitcore/libhermit-rs/files/5647262/log_3_core.log)
[...]

PR #129, opened 2020-12-06
make sure that the scheduler and all heap allocations are aligned to a cache line boundary

stlankes:
- an alignment avoids false sharing
- using an exponential backoff algorithm in the spinning loop of the semaphore


jschwe:
I checked out e683c86 and tested the SMP issue. The 4 core test reduced the runtime of the laplace test by 5 seconds (10%), however the runtime with 1-3 cores were more or less unchanged, so that could also be just noise (I only did one run).

stlankes:
Currently, I try to solve issue #78. I assumed that we have false sharing in our system. However, the message `Trying to get the scheduler for core 1, but it isn't available` sounds crazy fro me. Ok, the access to SCHEDULER is a little bit dirty. I know that I insert only at boot time some elements. Consequently, I don't protect the data structure. Now, I use a Vector instead of a BTreeMap. But I except that a vector is a little bit faster, but doesn't solve the problem.

src/mm/allocator.rs
```diff
@@ -11,14 +11,22 @@ use crate::mm::hole::{Hole, HoleList};
...
 const BOOTSTRAP_HEAP_SIZE: usize = 4096;
 
+#[cfg(target_arch = "x86_64")]
+const CACHE_LINE_SIZE: usize = 128;
+#[cfg(not(target_arch = "x86_64"))]
+const CACHE_LINE_SIZE: usize = 64;
+
 /// A fixed size heap backed by a linked list of free memory blocks.
+#[cfg_attr(target_arch = "x86_64", repr(align(128)))]
+#[cfg_attr(not(target_arch = "x86_64"), repr(align(64)))]
```
jbreitbart, line 29: Same as before, why do you align the struct to 128 on x86?

src/mm/allocator.rs
```diff
@@ -97,12 +104,11 @@ impl Heap {
...
-			if size < HoleList::min_size() {
-				size = HoleList::min_size();
-			}
-			let size = align_up!(size, mem::align_of::<Hole>());
-			let layout = Layout::from_size_align(size, layout.align()).unwrap();
+			let mut size = cmp::max(layout.size(), HoleList::min_size());
+			size = align_up!(size, mem::align_of::<Hole>());
+			size = align_up!(size, CACHE_LINE_SIZE);
+			let layout =
+				Layout::from_size_align(size, cmp::max(layout.align(), CACHE_LINE_SIZE)).unwrap();
```
jbreitbart, line 111: Will this force a minimum alignmend of CACH_LINE_SIZE for every allocation? Do you know how this is done on Linux or any other OS?

src/mm/test.rs
```diff
@@ -9,16 +9,23 @@
...
 	use std::prelude::v1::*;
 
 	use crate::mm::allocator::*;
 	use crate::mm::hole::*;
 
+	#[cfg(target_arch = "x86_64")]
+	const CACHE_LINE_SIZE: usize = 128;
+	#[cfg(not(target_arch = "x86_64"))]
+	const CACHE_LINE_SIZE: usize = 64;
+
```
jbreitbart, line 24: Is it possible to reuse the existing definitions?

src/mm/allocator.rs
jbreitbart: Also can you reuse the CACHE_LINE_SIZE constant here?

src/scheduler/mod.rs
```diff
@@ -49,6 +51,9 @@ impl SchedulerInput {
 		}
 	}
 }
+
+#[cfg_attr(target_arch = "x86_64", repr(align(128)))]
+#[cfg_attr(not(target_arch = "x86_64"), repr(align(64)))]
```
jbreitbart, line 56: As before. I think the size is wrong and reusing the constant would make it easier to maintain and understand

jbreitbart (changes requested):

src/mm/allocator.rs
```diff
@@ -11,14 +11,22 @@ use crate::mm::hole::{Hole, HoleList};
...
+use core::cmp;
 use core::ops::Deref;
 use core::ptr::NonNull;
 use core::{mem, ptr};
 
 /// Size of the preallocated space for the Bootstrap Allocator.
 const BOOTSTRAP_HEAP_SIZE: usize = 4096;
 
+#[cfg(target_arch = "x86_64")]
+const CACHE_LINE_SIZE: usize = 128;
```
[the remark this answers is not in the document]
stlankes: Yes and no. The prefetched loads more than one cache line (https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/cache_padded.rs#L60-L70).

src/mm/test.rs
stlankes: I don't known an existing definition.

src/scheduler/mod.rs
stlankes: See above

src/mm/allocator.rs
stlankes: Hm, normally the try to find smallest power-of-two block size big enough to hold requested size plus header.

https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/hurd/hurdmalloc.c#L213-L223

stlankes:
Now, I am using crossbeam to align data.

src/mm/allocator.rs
stlankes: I think that number in "align" must be a real number. You are not able to use a Rust constant.

src/mm/test.rs
jbreitbart: I was referring to the ones you defined before

stlankes: Ah, yes, that is true. I will fix it

src/mm/allocator.rs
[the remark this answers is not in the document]
jbreitbart: Well, no. The cache line size on x86 is 64 bytes. If you want to do padding on two cache lines that is fine, but then don't call it cache line size 😉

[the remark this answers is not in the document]
jbreitbart: Btw: polly calls it `hardware_destructive_interference_size`.

src/mm/allocator.rs
```diff
@@ -11,14 +11,25 @@ use crate::mm::hole::{Hole, HoleList};
...
 use core::ptr::NonNull;
 use core::{mem, ptr};
 
 /// Size of the preallocated space for the Bootstrap Allocator.
 const BOOTSTRAP_HEAP_SIZE: usize = 4096;
 
+#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+const CACHE_LINE_SIZE: usize = 128;
+#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+const CACHE_LINE_SIZE: usize = 64;
```
jbreitbart, line 25: As said in the other discussion, at least the wording seems wrong

src/mm/test.rs
```diff
@@ -9,16 +9,23 @@
...
 	use std::mem::{align_of, size_of};
 	use std::prelude::v1::*;
 
 	use crate::mm::allocator::*;
 	use crate::mm::hole::*;
 
+	#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+	const CACHE_LINE_SIZE: usize = 128;
+	#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+	const CACHE_LINE_SIZE: usize = 64;
```
jbreitbart, line 23: See before

jbreitbart (changes requested):

src/mm/allocator.rs
jbreitbart: You are right, didn't know that this is not possible.

src/mm/allocator.rs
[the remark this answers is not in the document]
stlankes: Ok, I renamed it

src/mm/allocator.rs
jbreitbart: Thanks for the link. That got me thinking. I am not sure why they do it they way they do it and I think jemalloc does it differently. Then I was thinking if we should try to do something that realloc can be done efficiently. After that I was worried that the implementation will get to complex and we'll have more errors than before. I guess, I am just not a fan of having to implement your own allocator, even though I understand why it is necessary at the moment.

Long story short: Feel free to ignore this comment and if you ever have good idea on how to get rid of implementing our own allocator I am all yours 😉

stlankes:
@jbreitbart Do I answer all open points?

stlankes:
bors try

stlankes:
I assume that Jens agrees after his last changes.

jbreitbart (approved):
Fine with me :wink:

PR #131, opened 2020-12-08
using a exponential backoff to realize a spinlock

stlankes:
Backing off in spin loops reduces contention and improves overall performance.


jounathaen:
Interesting.
My first thought was: If a spinlock exponentially increases the yield time and the locks are pulled frequently by different tasks, couldn't it be that one lock will start waiting incedibly long, because everytime it wakes up a newer task already took the lock?

But then I remembered: It is a ticket lock. Then this shouldn't be a problem :smiley: 

LGTM!

jounathaen (approved):

PR #140, opened 2020-12-14
add support of processors, which doesn't support FSGSBASE

stlankes:
by detecting the FSGSBASE support, the dispatcher will be patched to support it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,11 +38,12 @@ name = "measure_startup_time"
 harness = false
 
 [features]
-default = ["pci", "acpi"]
+default = ["pci", "acpi", "fsgsbase"]
 vga = []
 newlib = []
 pci = []
 acpi = []
+fsgsbase = []
 
 [dev-dependencies]
 x86 = { version = "0.34.*", default-features = false }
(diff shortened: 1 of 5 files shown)


jbreitbart (commented):
I hate to be that guy, but I am not sure if we really need support for CPUs without FSGSBASE. Intel added support with Ivy Bridge (2005) and AMD has support in their whole Zen lineup.

I am worried that this just adds complexity without any real benefit. There is no bug in a feature that is not available. Sorry. :roll_eyes:

jschwe:
@jbreitbart FYI Ivy bridge was launched in 2012 and AMD Zen in 2017. 

Of course, your argument about the added complexity still stands.

jbreitbart:
> @jbreitbart FYI Ivy bridge was launched in 2012 and AMD Zen in 2017.

Thanks, that is true. I originally wanted to write CPUs older than 5 years, but decided to check the release date of Ivy Bridge and than my mind somehow decided to fuse the infos to 2005. 🤦

stlankes:
I totally agree, but our internal CI consists of old cluster. To test the kernel with uhyve (which depends on nested virtualization), we need such a patch.

PR #147, opened 2020-12-29
Tests: Add timeout for integration tests

jschwe:
This adds the --timeout flag to `hermit_test_runner.py`. For now the value is simply a positive integer meaning a timeout in seconds.
The default timeout is 300 seconds.
This timeout applies to all integration tests (as in each integration test may take up to timeout seconds).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -63,9 +63,9 @@ test:integration:
     - kvm-ok
     - python3 --version
     - |
-        bash -c "HERMIT_LOG_LEVEL_FILTER=Debug cargo test --tests --no-fail-fast -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --no-default-features --features=pci,acpi --target x86_64-unknown-hermit-kernel -- --uhyve_path=$HOME/.cargo/bin/uhyve --veryverbose"
+        HERMIT_LOG_LEVEL_FILTER=Debug cargo test --tests --no-fail-fast -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --no-default-features --features=pci,acpi --target x86_64-unknown-hermit-kernel -- --uhyve_path=$HOME/.cargo/bin/uhyve --veryverbose
     - |
-        bash -c "HERMIT_LOG_LEVEL_FILTER=Debug cargo test --tests --no-fail-fast -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --no-default-features --features=pci,acpi --target x86_64-unknown-hermit-kernel -- --uhyve_path=$HOME/.cargo/bin/uhyve --num_cores 2 --veryverbose"
+        HERMIT_LOG_LEVEL_FILTER=Debug cargo test --tests --no-fail-fast -Z build-std=core,alloc -Z build-std-features=compiler-builtins-mem --no-default-features --features=pci,acpi --target x86_64-unknown-hermit-kernel -- --uhyve_path=$HOME/.cargo/bin/uhyve --num_cores 2 --veryverbose
   tags:
     - privileged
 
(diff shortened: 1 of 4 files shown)


stlankes:
bors try

stlankes:
I thank that you have to activate the test: https://github.com/hermitcore/libhermit-rs/blob/master/.gitlab-ci.yml#L58

stlankes:
bors try

stlankes:
Do you expect https://git.rwth-aachen.de/acs/public/hermitcore/libhermit-rs/-/jobs/1322176?

jschwe:
@stlankes I'm still somewhat surprised `Arguments` line gets printed after the exception. I guess that probably is due to something related to buffering. In the latest commit I'm catching the exception, so we should see the output (stdout, stderr) after the timeout occurs. This time I also tested it locally ;-).
I have no clue why `basic_math` is failing on gitlab, so hopefully the output will give some hints.

stlankes:
bors try

stlankes:
crazy https://git.rwth-aachen.de/acs/public/hermitcore/libhermit-rs/-/jobs/1322312

stlankes:
bors try

stlankes:
bors try

jschwe:
@stlankes  Okay, the error is apparently the following: 
`[ERROR] FSGSBASE support is enabled, but the processor doesn't support it`
see https://git.rwth-aachen.de/acs/public/hermitcore/libhermit-rs/-/jobs/1323581

This is a bit surprising, since the uhyve test runs fine. Is there more than one machine with the `privileged` tag?

The binaries themselves seem to be okay, I downloaded and ran them without problems on my local machine.

stlankes:
Ah, my mistake 😁 I will fix it

stlankes:
@jschwe I created a pull request to your branch....

stlankes:
bors try

stlankes:
Seems not to work. Why is the kernel still using fsgsbase?

jschwe:
@stlankes Surprisingly only one test is failing...
basic_math:
```
][INFO] Disable IOAPIC timer\r\n[1][!!!PANIC!!!] /root/.rustup/toolchains/nightly-2020-12-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec.rs:1217: insertion index (is 1) should be <= len (is 0)\r\n[1][INFO] Number of interrupts\r\n[1][INFO] Shutting down system\r\n'
````

Since the panic occurs after Disable IOAPIC, I think we are still in kernel code here. Which is very strange, since the other tests work fine.
Do you know if a panic flushes the buffer if something still has to be printed?

jschwe:
Another thing I noticed is that we enabled the feature acpi, but `acpi::poweroff` doesn't seem to be working in this case.
Is this a bug in uhyve or is this expected? In any case I think we should ensure that rusty-hermit exits properly if `shutdown` is called when running on uyhve and qemu.

stlankes:
bors try

jschwe:
My latest commit didn't fix anything, it only set the log level filter to debug. I'd like to find out what caused the panic in the second to last run. @stlankes Maybe you could let bors run the test a few more times to see if it occurs again?

stlankes:
bors try

stlankes:
Happy new year 😀

jschwe:
> Happy new year grinning

Happy new year to you to!

I added a few cleanup commits, that hopefully didn't break anything.

stlankes:
bors try

stlankes:
bors try

stlankes:
bors try

stlankes:
bors try

stlankes:
bors try

stlankes:
@jschwe Should we merge it and look later, why we get some times an exception. It is difficult to reproduce this failure.

jschwe:
@stlankes Yes, I think that's reasonable. 
In the future we could discuss adding a job that regularly (maybe like once a week) runs all tests multiple times (e.g. 1000x) to potentially spot flaky bugs.

PR #155, opened 2021-01-15
add option to avoid including detail pci information

stlankes:
By avoid the usage of detail PCI information, the image size of `hello_world` is reduced by 1.2 MByte


jbreitbart:
> By avoid the usage of detail PCI information, the image size of `hello_world` is reduced by 1.2 MByte

How much does this actually save? Just providing the final size isn't really telling anything 😉 Also there are no tests to check if the code actually compiles and you probably should not add any features without adding tests or some of them will break sooner or later. Arguably you may even have to test combinations of features to be sure they do not influence and depend on each other.

stlankes:
I can build the crate without `pic_ids` support, but tests requires also changes in `hermit-sys`. See also hermitcore/rusty-hermit#104

PR #156, opened 2021-01-16
add basic tests without SMP support

stlankes:


.github/workflows/nightly_nosmp.yml
```diff
@@ -0,0 +1,119 @@
+name: Tests with SMP support (nightly)
```
jbreitbart, line 1: Either the filename is wrong or the name

.github/workflows/nightly_nosmp.yml
```diff
@@ -0,0 +1,119 @@
...
+        with:
+          path: 'libhermit-rs'
+      - name: Check Cargo availability
+        run: cargo --version
+      - name: Cargo Test libhermit-rs (Unittests on Host)
+        run: cargo test --lib --target x86_64-unknown-linux-gnu
+        working-directory: libhermit-rs
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+      - name: Install qemu/nasm (apt)
+        run: sudo apt-get update --fix-missing && sudo apt-get install qemu-system-x86 nasm
```
jbreitbart, line 67: Is nasm still required?

.github/workflows/nightly_nosmp.yml
```diff
@@ -0,0 +1,119 @@
...
+      - staging
+      - trying
+
+jobs:
+  build:
+    name: Tests without SMP support
+    runs-on: ${{ matrix.os }}
+    # these tests based on the nightly compiler, which can be broken
+    # consequently, we continue on an error
+    #continue-on-error: true
```
jbreitbart, line 17: Should this be enabled? If not, can you remove it?

.github/workflows/nightly.yml
```diff
@@ -90,10 +90,10 @@ jobs:
         if: ${{ matrix.os == 'windows-latest' }}
       - name: Building dev version
         run:
-          cargo build -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit
+          cargo build -p rusty_demo -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit
       - name: Building release version
         run:
-          cargo build -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit --release
+          cargo build -p rusty_demo -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit --release
```
jbreitbart, line 96: Why are you only building rusty_demo?

jbreitbart (changes requested):

.github/workflows/nightly_nosmp.yml
stlankes: Oh, a typo... I fixed it.

.github/workflows/nightly_nosmp.yml
stlankes: `nasm` is required to build the boot loader, which we need to run the kernel within qemu. The unikernel doesn't depend on `nasm`.

.github/workflows/nightly_nosmp.yml
stlankes: I removed these lines.

.github/workflows/nightly.yml
stlankes: It is currently the only test, which is used.

jbreitbart: That is only true in this PR. The one you just merged added a standalone pi sequential test

stlankes: But the standalone pi sequential test is also included in rusty_demo

stlankes:
bors try

.github/workflows/nightly.yml
jbreitbart: > But the standalone pi sequential test is also included in the package `rusty_demo`. Consequently, we build also the standalone version of the pi sequential test.

No, you don't. If I change something in the sequential pi test that does not compile, you will not notice it with this CI pipeline. Furthermore, if you change something in the demo and not adapt the sequential pi test, you may break the standalone test without realizing it. We can remove the standalone tests again, but even than I fail to see the benefit of that change. 🤷

Issue #157, opened 2021-01-16
libhermit-rs fails to compile without feature "pci"


jschwe:
`drivers/net` contains imports / uses functions that depend on `pci` being enabled.

```
 error[E0433]: failed to resolve: use of undeclared crate or module `pci`
     --> src/drivers/net/mod.rs:149:30
      |
  149 |     let check_scheduler = match pci::get_network_driver() {
      |                                 ^^^ use of undeclared crate or module `pci`

```
I'm not really familiar with the network module, but if it depends on pci, it should be disabled if pci is not enabled. Alternatively we could remove the `pci` feature  and simply unconditionally enable it.

stlankes:
We fixed it with #160

PR #160, opened 2021-01-20
fix bug by building the minimal version of libhermit

stlankes:
This PR should solve issue #157 In addition, I add a small test to build the minimal configuration.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -89,8 +89,9 @@ jobs:
           echo "C:\Program Files\nasm" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
         if: ${{ matrix.os == 'windows-latest' }}
       - name: Building minimal kernel
+        working-directory: libhermit-rs
         run:
-          cargo build -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit --manifest-path=Cargo.toml --no-default-features
+          cargo build -Z build-std=core,alloc --target x86_64-unknown-hermit-kernel --no-default-features
       - name: Building dev version
         run:
           cargo build -Z build-std=std,core,alloc,panic_abort --target x86_64-unknown-hermit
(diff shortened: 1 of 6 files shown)


stlankes:
@jschwe It seems to work...

jschwe:
@stlankes Have you tried building and running an application? 
I get a lot of GP Exceptions when trying to run hello-world with no-default-features and this PR, but maybe I'm doing something wrong.
Steps:
1. Checkout master of rusty-hermit
2. Checkout this PR  into libhermit-rs
3. go back to rusty-hermit directory
4. ` cargo build --manifest-path examples/hello_world/Cargo.toml --target x86_64-unknown-hermit --no-default-features`
5. `cd loader && make && cd ..`
6. `qemu-system-x86_64 -display none -smp 1 -m 64M -serial stdio  -kernel loader/target/x86_64-unknown-hermit-loader/debug/rusty-loader -initrd target/x86_64-unknown-hermit/debug/pi_sequential -cpu qemu64,apic,fsgsbase,rdtscp,xsave,fxsr`

stlankes:
It works only on uhyve, QEMU depends on PCI.... That is the reason, why I don't test it within GitHub Actions. In the future, I want to support of Qemu's Micro VM (https://github.com/bonzini/qemu/blob/master/docs/microvm.rst). Currently, it doesn't run on it. I have to check, why it isn't working on this setup.

jschwe:
Ah okay, that explains it. 
I suggest to build the minimal kernel with `--target x86_64-unknown-hermit-kernel` and remove the `std` from `-Z build-std=<>`.
In practice, it's not really relevant of course, since we aren't using the binary, but I think it's better to stay consistent regarding the different parameters when building the kernel vs an application.
Other than that the PR looks good.

stlankes:
You are right, I fixed it

jschwe (approved):

stlankes:
bors retry

jschwe:
It appears that both the integration tests and the multithreaded uhyve test is failing. The multithreaded uhyve tests crashes, but still seems to exit with 0, which causes the test to appear as passed, even though it should have failed. I'll open an issue over at uhyve to track this.
The integration tests seem to be running fine, but uhyve (or possibly also libhermit) doesn't exit after they are finished. 


I also noticed, that the way I'm dumping the stdout in the integration tests does not really result in very readable output, so I'll open a PR to improve that in the next days.

PR #166, opened 2021-02-02
clearify license text, add workround to use raw-cpuid

stlankes:
The workaround should solve issue hermitcore/libhermit-rs#164 and is only necessry until gz/rust-x86#71 is accepted.


jbreitbart:
Why do we need that workaround? Just to close #164 ? Is it really worth it or should we just wait until @gz makes a new release of the x86 crate?

stlankes:
You are right, the crate will come. We are able to wait

gz:
Released 0.36, Thanks!

stlankes:
@gz I didn't want to put any pressure on you. Thanks for the fast response.

PR #180, opened 2021-04-28
Moved pipeline to new gitlab CI runners

jounathaen:


jounathaen:
It looks like the reason for the fails is the code and not the infrastructure anymore :slightly_smiling_face::+1:

stlankes:
No :-)

We need the privileged runner to test the kernel with KVM. We have currently a problem with the release version, which is reason that "user-space" isn't able to run.

jounathaen:
The gitlab runner should actually have KVM access. I have to check again...

stlankes:
Unfortunately I have closed the PR by mistake. However, your changes are already integrated in the latest version.

PR #192, opened 2021-05-19
switch to intel syntax for assembly code

stlankes:
global_asm switch per default to the intel syntax, because the behaviour is than similiar to inline assembly. libhermit-rs follow this step and switch to the same default syntax.

See also rust-lang/rust#84107


mkroening (approved):
LGTM! :+1:

PR #193, opened 2021-05-20
Update data layout to include new x86 address spaces

mkroening:
Same as https://github.com/hermitcore/rusty-loader/pull/18.


stlankes (approved):
Thanks for the improvement.

PR #197, opened 2021-05-26
new interface to the network driver

stlankes:
- the new interface is required for hermitcore/rusty-hermit#117
- in principle, it reopens #176

### DIFF
--- a/src/syscalls/tasks.rs
+++ b/src/syscalls/tasks.rs
@@ -291,7 +291,7 @@ pub extern "C" fn sys_join(id: Tid) -> i32 {
 static TASKS: SpinlockIrqSave<BTreeMap<TaskId, TaskHandle>> = SpinlockIrqSave::new(BTreeMap::new());
 
 extern "C" fn __sys_block_current_task(timeout: &Option<u64>) {
-	let wakeup_time = timeout.map(|timeout| arch::processor::get_timer_ticks() + timeout);
+	let wakeup_time = timeout.map(|timeout| arch::processor::get_timer_ticks() + timeout * 1000);
 	let core_scheduler = core_scheduler();
 	let handle = core_scheduler.get_current_task_handle();
 	let tid = core_scheduler.get_current_task_id();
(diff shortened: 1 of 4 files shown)


src/syscalls/tasks.rs
```diff
@@ -289,19 +289,25 @@ pub extern "C" fn sys_join(id: Tid) -> i32 {
...
 /// Set the current task state to `blocked`
 #[no_mangle]
 pub extern "C" fn sys_block_current_task() {
-	kernel_function!(__sys_block_current_task())
+	kernel_function!(__sys_block_current_task(None))
+}
+
+/// Set the current task state to `blocked`
+#[no_mangle]
+pub fn sys_block_current_task_with_timeout(timeout: Option<u64>) {
```
mkroening, line 309: I think, we could use `thread::sleep`, which calls `__sys_usleep`, instead of adding this function.

https://github.com/hermitcore/libhermit-rs/blob/2bf1124afb44060b539aff67df994f7df8b4536e/src/syscalls/tasks.rs#L117-L134

src/syscalls/tasks.rs
```diff
@@ -289,19 +289,25 @@ pub extern "C" fn sys_join(id: Tid) -> i32 {
...
 
-fn __sys_block_current_task() {
+fn __sys_block_current_task(timeout: Option<u64>) {
 	let core_scheduler = core_scheduler();
 	let handle = core_scheduler.get_current_task_handle();
 	let tid = core_scheduler.get_current_task_id();
 
 	TASKS.lock().insert(tid, handle);
-	core_scheduler.block_current_task(None);
+	core_scheduler.block_current_task(timeout);
```
mkroening, line 298: `block_current_task` takes an absolute timestamp (`Instant`) instead of a relative time (`Duration`), doesn't it?

So currently this would result in basically no sleeping at all. If we were to replace `sys_block_current_task_with_timeout` (see below), we could avoid this altogether.

mkroening (changes requested):
I think there is an issue with sleeping currently. This might be related to the wireguard starving, I observed, but I did not test the changes I proposed in my comments yet.

PR #204, opened 2021-05-29
Fix clippy errors

mkroening:
We currently set `clippy::all` to `warn`. Some lints from `clippy::all` are `deny` by default. This removes our relaxation and fixes all errors.

What do you think?


stlankes (approved):
👍

PR #205, opened 2021-05-29
virtqueue: Fix using send buffer as recieve buffer

mkroening:
While working on more clippy lints, I noticed something that seems to by an issue caused by copy-pasting.

We surely don't want to treat send buffers as receive buffers, right?


stlankes (approved):
How could it ever work? I have never seen any problem.

mkroening:
Resolved the merge conflicts.

It works, because these functions are never used.

PR #206, opened 2021-05-29
virtqueue: Avoid going over unsafe raw ptrs

mkroening:
We can create `&mut` to fields without having to resort to raw pointers.


stlankes (approved):
bors r+

PR #207, opened 2021-05-29
Fix clippy warnings

mkroening:
These are a bunch of fixes for clippy warnings.

In some circumstances, where fixing is not straightforward, I allowed the lints instead.

Together with the other clippy PRs, there shouldn't be any clippy warnings now, so we can enable those in CI in the future.

What do you think?


stlankes (approved):
👍

PR #208, opened 2021-05-30
RxQueues::get_next: Explicitly bind transfer

mkroening:
Workaround for #209. Introduced in https://github.com/hermitcore/libhermit-rs/commit/a9fc0530473911d00caf2357a6dad69e034e92b6.


mkroening:
Okay, this is the proper fix for this place. Rust tries to panic without explicit binding, because we are holding the `borrow_mut` until the next `borrow_mut` otherwise.

So this fix is good, but panicking here fails in debug. I'll adjust #209 accordingly.

stlankes (approved):
bors r+

PR #215, opened 2021-06-10
Rename .cargo/config to config.toml

mkroening:


mkroening:
The clippy failure has nothing to do with this PR but is introduced by better lints in newer nightlies. #216 addresses this by making sure to use the nightly specified in `toolchain.toml`.

stlankes:
@mkroening Clippy doesn't like the renaming of the toolchain file. Do you know a solution?

mkroening:
Rebased on latest `master`.

stlankes (approved):
bors r+

PR #216, opened 2021-06-10
Don't override nightly for clippy CI

mkroening:
We want to use clippy from the nightly specified in `toolchain.toml` instead of using the latest nightly.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -17,8 +17,6 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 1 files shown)


mkroening:
actions-rs does not support `rust-toolchain.toml` yet, so we have to wait for that.

mkroening:
Oh, no! Because of https://github.com/hermitcore/libhermit-rs/pull/215#issuecomment-858913335 I marked this as draft PR, thinking that bors would not merge it (https://forum.bors.tech/t/keep-track-of-unmergeable-and-draft-pull-requests/556).

This should be reverted.

PR #218, opened 2021-06-10
Strongly link panic and oom handlers

mkroening:
Depends on https://github.com/hermitcore/rusty-hermit/pull/139.

Having weak linkage on the handlers does no harm but is not necessary after the before mentioned PR.


mkroening:
Rebased on latest `master`.

stlankes:
bors try

PR #221, opened 2021-06-11
Use toolchain.toml compatible toolchain action

mkroening:


mkroening:
There already was a PR for the toolchain action to support toml files. I cleaned it up and resubmitted it in https://github.com/actions-rs/toolchain/pull/184.

Until it is supported upstream we can maintain this patch ourselves. I currently do so in my fork of the project [mkroening/toolchain@downstream](https://github.com/mkroening/toolchain/tree/downstream), but perhaps we would want this in the hermitcore organization?

stlankes (approved):
👍

PR #228, opened 2021-06-13
Replace `llvm_asm!` with `asm!`

### LINKED ISSUE #223: Migrate from `llvm_asm` to `asm`
I'll have a go at this.

mkroening:
Resolves #223.

This removes all usages of `llvm_asm!` in three steps:

1. Replace `llvm_asm!` with intrinsics where possible.
2. Replace `llvm_asm!` with wrappers provided by the `x86_64` crate.
3. Replace `llvm_asm!` with `asm!`.

@stlankes, you mentioned to wait with the assembly in `macros.rs`, but I had already done them. They are exclusively in the last commit. If you want, I can remove it again.

### DIFF
--- a/src/arch/x86_64/kernel/irq.rs
+++ b/src/arch/x86_64/kernel/irq.rs
@@ -6,8 +6,6 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
-pub use x86_64::instructions::interrupts::{disable, enable, enable_and_hlt as enable_and_wait};
-
 use crate::arch::x86_64::kernel::apic;
 use crate::arch::x86_64::kernel::idt;
 use crate::arch::x86_64::kernel::percore::*;
@@ -20,7 +18,8 @@ use crate::alloc::string::ToString;
 use alloc::collections::BTreeMap;
 use alloc::string::String;
 use core::fmt;
-use x86_64::registers::rflags::{self, RFlags};
+use x86::bits64::rflags::{self, RFlags};
+use x86::irq;
 
 static IRQ_NAMES: SpinlockIrqSave<BTreeMap<u32, String>> = SpinlockIrqSave::new(BTreeMap::new());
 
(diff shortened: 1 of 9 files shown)


stlankes:
Do we have now two x86(_64) crates in the library?

mkroening:
> Do we have now two x86(_64) crates in the library?

The `x86_64` crate is what we migrated to in eduOS. It's the crate by the rust-osdev organization. I can adapt this change to use the `x86` crate if you don't want to have the `x86_64` crate.

src/arch/x86_64/kernel/irq.rs
```diff
@@ -6,19 +6,21 @@
 // http://opensource.org/licenses/MIT>, at your option. This file may not be
 // copied, modified, or distributed except according to those terms.
 
+pub use x86_64::instructions::interrupts::{disable, enable, enable_and_hlt as enable_and_wait};
```
jounathaen, line 9: I'm only looking at the diff, but where is that import used?

src/arch/x86_64/kernel/percore.rs
```diff
@@ -115,14 +106,15 @@ impl Is32BitVariable for u32 {}
 impl<T: Is32BitVariable> PerCoreVariableMethods<T> for PerCoreVariable<T> {
 	#[inline]
 	unsafe fn get(&self) -> T {
-		let value: T;
-		llvm_asm!("movl %gs:($1), $0" : "=r"(value) : "r"(self.offset()) :: "volatile");
-		value
+		let value: u32;
```
jounathaen, line 109: is `u32` here on purpose?

src/arch/x86_64/kernel/irq.rs
jounathaen: (the `enable_and_hlt` one)

stlankes:
> > Do we have now two x86(_64) crates in the library?
> 
> The `x86_64` crate is what we migrated to in eduOS. It's the crate by the rust-osdev organization. I can adapt this change to use the `x86` crate if you don't want to have the `x86_64` crate.

I would prefer one crate. Which one is for me not so important. Both crates are good!

src/arch/x86_64/kernel/irq.rs
mkroening: `enable_and_hlt` is reexported as `enable_and_wait` and replaces our implementation of the latter. It is used by the scheduler.

src/arch/x86_64/kernel/percore.rs
mkroening: Yes, as we want to explicitly use 32-bit memory moves here.  

https://github.com/hermitcore/libhermit-rs/blob/d7ea077c0f9bc4fb7412aa3e7bb67c68a41c7267/src/arch/x86_64/kernel/percore.rs#L104-L105

mkroening:
> > > Do we have now two x86(_64) crates in the library?
> > 
> > 
> > The `x86_64` crate is what we migrated to in eduOS. It's the crate by the rust-osdev organization. I can adapt this change to use the `x86` crate if you don't want to have the `x86_64` crate.
> 
> I would prefer one crate. Which one is for me not so important. Both crates are good!

All right. Then I'll migrate the new code to `x86` for now. Evaluating a possible migration would be another issue.

mkroening:
Okay, I migrated back to `x86`.

mkroening:
Worked around an issue when migrating `kernel_function!`, see #234.

Windows CI is failing, because the NASM website is down.

mkroening:
I'll have a deeper look at early-clobber before I am confident in this again.

mkroening:
Okay, now that one output of `llvm_asm` has been identified as being wrongly specified as `"=r"` instead of `"=&r"`, I wanted to check the other places.

Specifically, I worried about marking the output in `PerCoreVariableMethods::smp` as `lateout` resulting in code like `mov rax,gs:[rax]`, but that is safe.

I further optimized `asm!`'s options and better clarified the workaround for #234.

This is ready for review again. :D

stlankes (approved):
Maybe we could later migrate to x86_64. I would avoid to use two similar crates.

PR #229, opened 2021-06-13
Enforce clippy via bors

mkroening:
Now that we have a stable Clippy CI, we can enforce Clippy lints via bors.


stlankes (approved):
bors r+

jounathaen:
The CI failure seems to be a dockerhub hickup. A rerun should probably work

stlankes:
bors retry

Issue #226, opened 2021-06-11
Migrate to upstream actions-rs/toolchain


mkroening:
Requires https://github.com/actions-rs/toolchain/issues/126

CC: https://github.com/hermitcore/libhermit-rs/pull/221

PR #230, opened 2021-06-14
CI: Rework toolchain setup

mkroening:
This avoids double-installation of rust-toolchains in CI and enables coloring in the output.

Fixes https://github.com/hermitcore/libhermit-rs/issues/226.


mkroening:
This is now ready for review. Windows CI is failing everywhere currently, because the NASM servers are down.

This started as a fix for #226, but it grew to a whole rework of the GitHub Actions workflows.

The idea regarding #266 is that rustup and cargo automatically set up the toolchain according to `rust-toolchain.toml`, so I introduced a step calling `rustup show`, which does exactly this.

I also added the `CARGO_TERM_COLOR: always` environment variable, to have nicely colored output even in CI.

We can omit the `-Z build-std=core,alloc,panic_abort -Z build-std-features=compiler-builtins-mem --target x86_64-unknown-hermit-kernel.json`, since these settings are automatically picked up by cargo nowadays.

It was also possible to streamline the test workflow, reducing the differences between the operating systems.

Note that this includes https://github.com/hermitcore/rusty-loader/pull/24 via a sed invocation.

stlankes:
bors try

stlankes (approved):
bors r+

PR #236, opened 2021-06-16
Semaphore: Drop guard explicitly for readability

mkroening:
This is a rather opinionated change regarding code readability.

In my code editor with four spaces per tab, the extra scope looks quite weird. By explicitly dropping the guard instead, I personally find this easier to read. What do you think?


stlankes (approved):
bors r+

PR #239, opened 2021-06-16
Fix `Console` unsoundness

### LINKED ISSUE #238: Unsoundness in `Stdout` implementation
In `SyscallInterface::write`, we currently call `str::from_utf8_unchecked`, but there is no guarantee that `buf` is in fact valid UTF-8 data.

This can be triggered easily:

```rust
#[cfg(target_os = "hermit")]
extern crate hermit_sys;

use std::io::{self, Write};

fn main() {
	// some invalid bytes, in a vector
	let sparkle_heart = vec![0, 159, 146, 150];
	io::stdout().write_all(&sparkle_heart).unwrap();
}
```

While this does not cause problems currently, because the underlying `fmt::Write` implementation treats the `str` as bytes again, we should get rid of this invariance violation nonetheless.

mkroening:
Fixes #238.

I also made the constructor of `Console` private. Where's the fun in locking a `Console`, when everyone can create new ones?

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -10,7 +10,7 @@ use crate::arch;
 use crate::synch::spinlock::SpinlockIrqSave;
 use core::fmt;
 
-pub struct Console;
+pub struct Console(());
 
 /// A collection of methods that are required to format
 /// a message to HermitCore's console.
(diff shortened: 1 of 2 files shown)


stlankes (approved):
bors r+

PR #240, opened 2021-06-16
Remove unintended dereference

mkroening:
87a9d3f9ef7acfd9e8f2188f27457e275ea9a925 was wrong.

Instead of calculating an address at one place in `acpi`, it read from that address.
The warning in `apic` was due to an implicit borrow and can be resolved by an explicit copy into a let binding.


stlankes (approved):
bors r+

PR #243, opened 2021-06-23
Complete migration of kernel target

mkroening:
I noticed this due to [failing docs CI](https://github.com/hermitcore/libhermit-rs/runs/2889060428).
The rename began in 18dd15f.
It seems that the repository-local json target specifications are not used, so I removed those.


stlankes (approved):
bors r+

PR #245, opened 2021-06-24
CI: fix typo in target name

stlankes:


mkroening:
This is very similar to #243, which corrects even more cases, where we use the old name.

stlankes:
Ah, I oversaw #243

PR #247, opened 2021-06-25
Fix switching to kernel stack

mkroening:
This fixes https://github.com/hermitcore/libhermit-rs/issues/234 introduced in https://github.com/hermitcore/libhermit-rs/commit/41cf625c33801690f5be566db67e3eec5b372feb.

As described in https://github.com/hermitcore/libhermit-rs/issues/234, the issue with the old implementation was inlining resulting in wrongly managing stack space after the switch, potentially resulting in a page fault, when writing before the new stack:

```rust
fn sys_foo() {
    // reserves stack space for `a` (`rsp -= 16`)

    switch_rsp();

    // use reserved stack space (`[rsp + 16]`), even though rsp is not the same anymore
    let a = [0_u8; 16];
}
```

To my knowledge, inlining can only be really prevented, by explicitly `call`ing the function in assembly.

I fixed this issue in five steps:

1. Remove `switch_to_kernel!`, `switch_to_user!`, migrate uses to `kernel_function!`
2. Fix a potential stack overflow caused by a typo (unrelated)
3. Improve consistency of kernel functions by adding the prefix `__sys_` to all functions.
4. Make all `__sys_`-functions `extern "C"` to allow calling in assembly.
5. Reimplement `kernel_function!` and call kernel function in assembly.

### DIFF
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -145,7 +145,8 @@ pub fn has_ipdevice() -> bool {
 }
 
 #[cfg(not(feature = "newlib"))]
-pub fn __sys_uhyve_get_ip() -> [u8; 4] {
+#[allow(improper_ctypes_definitions)]
+extern "C" fn __sys_uhyve_get_ip() -> [u8; 4] {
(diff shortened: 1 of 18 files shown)


stlankes (approved):
Great work!

PR #251, opened 2021-06-29
Don't switch to kernel stack for now

mkroening:
Well, turns out https://github.com/hermitcore/libhermit-rs/pull/247 broke some kernel functions, because I added the size check for arguments and return types, but did not verify these requirements on all current kernel functions. Until I have reworked those, this restores the status quo.

Specifically, kernel functions for networking can't simply return `Result<(*mut u8, usize), ()>`, since that does not fit into one register.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -40,31 +40,33 @@ macro_rules! println {
 
 #[macro_export]
 macro_rules! kernel_function {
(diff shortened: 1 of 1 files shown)


mkroening:
A proper fix is https://github.com/hermitcore/libhermit-rs/pull/252.

PR #258, opened 2021-08-07
kernel_function: Clear upper bits of register

mkroening:
If the type in a register is smaller than the register, Rust assumes the upper bits in the register to be zero.

For example, checking a `bool` might happen via `test edi, edi`, although the type only occupies `dil`.


stlankes (approved):
bors r+

stlankes:
bors retry

stlankes:
bors retry

PR #276, opened 2021-09-05
Remove Cargo.lock

mkroening:
Since we use no`Cargo.lock` in downstream builds, we should not use one in upstream CI either.


stlankes:
I don't understand this PR.

mkroening:
We agreed on closing https://github.com/hermitcore/libhermit-rs/issues/270 and continue publishing libhermit on crates.io without lockfile. Since no lockfile is present for downstream builds (downloading from crates.io), we should not maintain one in this repository as well. That way we build the same way when using this repository directly (like in CI) as when building downstream.

mkroening:
bors retry

PR #287, opened 2021-10-09
add CI tests for Qemu's microvm

stlankes:
Qemu's microvm is a machine type inspired by Firecracker and constructed after its machine model

More information are published at https://qemu.readthedocs.io/en/latest/system/i386/microvm.html

### DIFF
--- a/.github/workflows/x86.yml
+++ b/.github/workflows/x86.yml
@@ -96,10 +96,14 @@ jobs:
         run: |
           FREQ=`grep 'cpu MHz' /proc/cpuinfo | head -1 | awk -F: '{print $2}' | awk '{printf("%d\n",$1 + 0.5)}'`
           echo "FREQ = $FREQ"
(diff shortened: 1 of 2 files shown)


stlankes:
@mkroening @jbreitbart If I am using a microVM, I am not able to shutdown by ACPI. Instead ACPI, I use a crate to send an exit command to Qemu. However, I am not able to return 0 because this is a requirement of Qemu. Currently, I return 1 on success. Do you see a better solution for my test script?

mkroening:
Ah, nevermind. `qemu-exit` does some shifting, so 1 as success results in 1 as output. Unfortunately, failure also results as 1 as output, so we need to change that. Regarding the script, I can get it working, it seems.

mkroening:
https://github.com/stlankes/libhermit-rs/pull/34 should fix it.

stlankes:
@mkroening 👍

PR #288, opened 2021-10-10
Rework C string abstractions

mkroening:


stlankes:
Naive question. What is the benefit of your implementation?

mkroening:
I didn't like the c_str functions in util very much. Thus I adapted the `CStr` from std abstraction to replace them.

This resolves the `TODO`s on the original functions and avoids unnecessary allocations.

mkroening:
bors retry

mkroening:
bors retry

PR #303, opened 2021-12-01
using write_volatile to guarantee that the data is written to the I/O device

stlankes:
@simonschoening reports that the current implementation doesn't work on RISC-V. By using `write_volatile` we should get a more platform independent solution.

### DIFF
--- a/src/drivers/virtio/transport/mmio.rs
+++ b/src/drivers/virtio/transport/mmio.rs
@@ -4,10 +4,11 @@
 #![allow(dead_code)]
 
 use crate::arch::mm::PhysAddr;
-use core::arch::x86_64::_mm_mfence;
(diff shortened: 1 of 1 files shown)


mkroening:
Ah, yes. This partly fixes https://github.com/hermitcore/libhermit-rs/issues/199.

The problem is known since May, when we disabled `mutable-noalias` as a workaround.

I am still waiting on https://github.com/rust-osdev/volatile/pull/22 to do an rework on all affected drivers.

In the meantime, this is a good step in the right direction. :+1: 

bors r+

mkroening (approved):

mkroening:
bors retry

stlankes:
bors retry

PR #305, opened 2021-12-01
Revert "align TLS to 32 byte boundary"

mkroening:
This reverts commit 1f63d3fe76cc6e1a50b8d1e39868d27a624ea921.

I found `PT_TLS.p_memsz` to automatically be correctly aligned to the usual `0x8`. I don't think, explicitly aligning again to `0x8` or `PT_TLS.p_align`  is necessary.

This fixes https://github.com/hermitcore/rusty-hermit/issues/170. :tada:


mkroening:
bors retry

mkroening:
This is the right knob but not the correct solution for all cases, I'm working on a fix.

bors cancel

mkroening:
Closing in favor of a more sophisticated solution. Turns out, we'll have to pipe the alignment through...

PR #311, opened 2021-12-02
Fix thread pointer calculation, rework TaskTLS

mkroening:
Closes https://github.com/hermitcore/rusty-hermit/issues/170. :tada: 

Replaces https://github.com/hermitcore/libhermit-rs/pull/305.

Depends on:

* https://github.com/hermitcore/rusty-loader/pull/46
* https://github.com/hermitcore/uhyve/pull/241

The cause of the issue is the calculation of the thread pointer:

```python
# Current calculation (wrong)
thread_ptr = tls_block_ptr + round(tls_len, 32)

# Correct calculation
thread_ptr = tls_block_ptr + round(tls_len, tls_align)
# round(tls_len, tls_align) is called tls_offset
```

This is fixed in the first two commits of this PR.

The later commits completely rework the `TaskTLS` struct. I did this to better understand the problem. The old comments were partly outdated and wrong. Also the code was not very rusty. I replaced manual memory allocations with `Box`es and restructured and recommented the code.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,7 +62,7 @@ test:integration:
     - lscpu
     - kvm-ok
     - python3 --version
-    - cargo install uhyve --locked
+    - cargo +nightly install --git https://github.com/hermitcore/uhyve.git --locked uhyve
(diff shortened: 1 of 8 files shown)


mkroening:
bors retry

PR #314, opened 2021-12-12
determine Local APIC and IO-APIC by the MultiProcessor Specification

stlankes:
If the ACPI detections fails of is disabled, RustyHermit uses the (old) MultiProcessor Specification to detect the Local APIC and the IO-APIC.

### DIFF
--- a/src/arch/x86_64/kernel/apic.rs
+++ b/src/arch/x86_64/kernel/apic.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::unaligned_references)]
+#![allow(unaligned_references)]
 
 use crate::arch;
 #[cfg(feature = "acpi")]
--- a/src/arch/x86_64/kernel/apic.rs
+++ b/src/arch/x86_64/kernel/apic.rs
@@ -355,9 +355,7 @@ fn detect_from_mp() -> Result<PhysAddr, ()> {
 		Err(())
 	}?;
 
-	info!("Found MP config at {:#x}", unsafe {
-		ptr::read_unaligned(&mp_float.mp_config)
-	});
+	info!("Found MP config at {:#x}", { mp_float.mp_config });
 	info!(
 		"System uses Multiprocessing Specification 1.{}",
 		mp_float.version
--- a/src/arch/x86_64/kernel/apic.rs
+++ b/src/arch/x86_64/kernel/apic.rs
@@ -410,7 +410,7 @@ fn detect_from_mp() -> Result<PhysAddr, ()> {
 		addr += mem::size_of::<ApicConfigTable>();
 		for _i in 0..mp_config.entry_count {
 			match unsafe { *(addr as *const u8) } {
-				// cpu entry
+				// CPU entry
 				0 => {
 					let cpu_entry: &ApicProcessorEntry =
 						unsafe { &*(addr as *const ApicProcessorEntry) };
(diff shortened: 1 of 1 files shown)


src/arch/x86_64/kernel/apic.rs
```diff
@@ -259,6 +315,145 @@ fn detect_from_acpi() -> Result<PhysAddr, ()> {
...
+fn detect_from_mp() -> Result<PhysAddr, ()> {
+	let mp_float = if let Ok(mpf) = search_mp_floating(PhysAddr(0x9F000u64), PhysAddr(0xA0000u64)) {
+		Ok(mpf)
+	} else if let Ok(mpf) = search_mp_floating(PhysAddr(0xF0000u64), PhysAddr(0x100000u64)) {
+		Ok(mpf)
+	} else {
+		Err(())
+	}?;
+
+	info!("Found MP config at 0x{:x}", mp_float.mp_config);
```
mkroening, line 360: Instead of manually reading the unaligned field with `addr_of!` and `read_unaligned`, there is a more ergonomic solution for formatting:

Adding curly braces means that a copy of the field is made, stored in a (properly aligned) temporary, and a reference to that temporary is being formatted.

```suggestion
	info!("Found MP config at {:#x}", { mp_float.mp_config });
```

src/arch/x86_64/kernel/apic.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unaligned_references)]
+
```
mkroening, line 2: We definitely should't disable this lint.

Fields of packed structs are not properly aligned, and creating a misaligned reference is undefined behavior (even if that reference is never dereferenced).

I submitted a suggestion for the offending part of the code.

```suggestion
```

mkroening (changes requested):
Apart from allowing the `unaligned_references` lint, this looks good to me. :+1:

stlankes:
@mkroening How can I pass the clippy checks?

mkroening:
> @mkroening How can I pass the clippy checks?

I had submitted a suggestion, avoiding the creation of a reference for formatting. That avoids the `unsafe` of an explicit unaligned read. Or would you prefer the explicit read?

stlankes:
Where? Do I miss a commit?

mkroening:
In the two comments along with my review: https://github.com/hermitcore/libhermit-rs/pull/314#pullrequestreview-829610243

stlankes:
Maybe I am blind. But I integrated your suggestions and it still fails.

mkroening:
This is the comment with the specific suggestion: https://github.com/hermitcore/libhermit-rs/pull/314#discussion_r767254867

This is the proposed way of formatting the field:

```rust
info!("Found MP config at {:#x}", { mp_float.mp_config });
```

stlankes:
Naive question. Why is this not a unaligned memory access?

mkroening:
It still is a unaligned memory access. What's UB is creating an unaligned _reference_. This happens inside the macro when written like `format!(.., mp_float.mp_config)`. `format!(.., { mp_float.mp_config })` copies the value and references the copy as I already explained in https://github.com/hermitcore/libhermit-rs/pull/314#discussion_r767254867.

stlankes:
Now, I understand!

mkroening (approved):

PR #317, opened 2021-12-12
TaskTLS: Copy memory on clone

mkroening:
This makes TaskTLS' Clone implementation copy modifications to TLS instead of creating a new one from scratch (introduced in fdf8bb7).

TaskTLS' are cloned in [`Task::clone`](https://github.com/hermitcore/libhermit-rs/blob/35d73d519885e72f27e7ba9f81266571163b3ed8/src/scheduler/task.rs#L429-L447) from [`PerCoreScheduler::clone_impl`](https://github.com/hermitcore/libhermit-rs/blob/35d73d519885e72f27e7ba9f81266571163b3ed8/src/scheduler/mod.rs#L160-L210) from [`sys_clone`](https://github.com/hermitcore/libhermit-rs/blob/35d73d519885e72f27e7ba9f81266571163b3ed8/src/syscalls/tasks.rs#L166-L183).

That system calls semantic should be the same as [`clone(2)`](https://www.man7.org/linux/man-pages/man2/clone.2.html#NOTES).

### DIFF
--- a/src/arch/x86_64/kernel/scheduler.rs
+++ b/src/arch/x86_64/kernel/scheduler.rs
@@ -237,6 +237,7 @@ impl Clone for TaskStacks {
 	}
 }
 
+#[derive(Clone)]
 pub struct TaskTLS {
 	_block: Box<[u8]>,
 	thread_ptr: Box<*mut ()>,
@@ -296,12 +297,6 @@ impl TaskTLS {
 
 	fn thread_ptr(&self) -> &*mut () {
 		&*self.thread_ptr
-	}
-}
-
-impl Clone for TaskTLS {
-	fn clone(&self) -> Self {
-		Self::from_environment()
 	}
 }
 


mkroening:
@stlankes, in fdf8bb7 you explicitly say “clone creates a new TLS” in the commit message. What do you think the semantics of this clone should be?

stlankes:
Why do we copy modifications to TLS? If I understand correctly (https://github.com/hermitcore/libhermit-rs/blob/master/src/arch/x86_64/kernel/scheduler.rs#L279-L280), we copy default values to the TLS.

mkroening:
Currently, when cloning `TaskTLS` we just create a new one from scratch with default values. The question is, whether this is the correct behavior. This patch makes it so, that cloning `TaskTLS` clones the corresponding memory with any modifications done since thread creation.

stlankes:
I used this system call for the creation of pthreads (https://github.com/hermitcore/pthread-embedded/blob/2a0fbf77278915b74fd26cefe62ec6552d49afd9/platform/hermit/pte_osal.c#L265). Maybe it is the wrong name for the syscall. But in this case, we have to use the default values.

mkroening:
All right! Thanks for the clarification. I'll add a comment in a different PR.

mkroening:
Closing in favor of https://github.com/hermitcore/libhermit-rs/pull/318.

PR #323, opened 2022-01-02
avoid exporting of macros

stlankes:
- inline wrapper functions
- don't use a thread-safe version to initialize the kernel logger
  because at this stage the kernel doesn't provide threads

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,4 +1,4 @@
-use log::{set_logger, set_max_level, LevelFilter, Metadata, Record};
+use log::{set_logger_racy, set_max_level, LevelFilter, Metadata, Record};
 
 /// Data structure to filter kernel messages
 struct KernelLogger;
@@ -24,8 +24,8 @@ impl log::Log for KernelLogger {
 	}
 }
 
-pub fn init() {
-	set_logger(&KernelLogger).expect("Can't initialize logger");
+pub unsafe fn init() {
+	set_logger_racy(&KernelLogger).expect("Can't initialize logger");
 	// Determines LevelFilter at compile time
 	let log_level: Option<&'static str> = option_env!("HERMIT_LOG_LEVEL_FILTER");
 	let max_level: LevelFilter = match log_level {
@@ -56,7 +56,7 @@ macro_rules! infoheader {
 
 #[macro_export]
 macro_rules! infoentry {
-	($str:expr, $rhs:expr) => ($crate::infoentry!($str, "{}", $rhs));
+	($str:expr, $rhs:expr) => (crate::infoentry!($str, "{}", $rhs));
 	($str:expr, $($arg:tt)+) => (::log::info!("{:25}{}", concat!($str, ":"), format_args!($($arg)+)));
 }
 
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -18,16 +18,14 @@ macro_rules! align_up {
 /// for HermitCore.
 #[macro_export]
 macro_rules! print {
-	($($arg:tt)+) => ({
-		$crate::_print(format_args!($($arg)*));
-	});
+	($($arg:tt)*) => ($crate::_print(::core::format_args!($($arg)*)));
 }
 
 /// Print formatted text to our console, followed by a newline.
 #[macro_export]
 macro_rules! println {
 	() => ($crate::print!("\n"));
-	($($arg:tt)+) => ($crate::print!("{}\n", format_args!($($arg)+)));
+	($($arg:tt)*) => ($crate::print!("{}\n", ::core::format_args!($($arg)*)));
 }
 
 /// Runs `f` on the kernel stack.
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.github/workflows/aarch64.yml
+++ b/.github/workflows/aarch64.yml
@@ -23,6 +23,8 @@ jobs:
       matrix:
         os: [ubuntu-latest]
     steps:
+      - name: Install cargo-binutils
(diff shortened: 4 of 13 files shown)


src/macros.rs
```diff
@@ -16,18 +16,15 @@ macro_rules! align_up {
 ///
 /// From http://blog.phil-opp.com/rust-os/printing-to-screen.html, but tweaked
 /// for HermitCore.
-#[macro_export]
 macro_rules! print {
-	($($arg:tt)+) => ({
-		$crate::_print(format_args!($($arg)*));
-	});
+	($($arg:tt)*) => ($crate::_print(format_args!($($arg)*)));
```
mkroening, line 20: Macro hygiene:

```suggestion
	($($arg:tt)*) => ($crate::_print(::core::format_args!($($arg)*)));
```

src/macros.rs
```diff
@@ -16,18 +16,15 @@ macro_rules! align_up {
...
 }
 
 /// Print formatted text to our console, followed by a newline.
-#[macro_export]
 macro_rules! println {
-	() => ($crate::print!("\n"));
-	($($arg:tt)+) => ($crate::print!("{}\n", format_args!($($arg)+)));
+	() => (print!("\n"));
+	($s:expr) => (print!(concat!($s, "\n")));
```
mkroening, line 26: What is this specialization used for? This allows `println!(3)`, which we should discourage. `std`'s `println` requires the format argument to be a string literal.

```suggestion
```

src/macros.rs
```diff
@@ -16,18 +16,15 @@ macro_rules! align_up {
...
-	});
+	($($arg:tt)*) => ($crate::_print(format_args!($($arg)*)));
 }
 
 /// Print formatted text to our console, followed by a newline.
-#[macro_export]
 macro_rules! println {
-	() => ($crate::print!("\n"));
-	($($arg:tt)+) => ($crate::print!("{}\n", format_args!($($arg)+)));
+	() => (print!("\n"));
```
mkroening, line 25: Macro hygiene:

```suggestion
	() => ($crate::print!("\n"));
```

src/macros.rs
```diff
@@ -16,18 +16,15 @@ macro_rules! align_up {
...
+	($s:expr) => (print!(concat!($s, "\n")));
+	($($arg:tt)*) => (print!("{}\n", format_args!($($arg)*)));
```
mkroening, line 27: Macro hygiene:

```suggestion
	($($arg:tt)*) => ($crate::print!("{}\n", ::core::format_args!($($arg)*)));
```

src/logging.rs
```diff
@@ -54,13 +56,11 @@ macro_rules! infoheader {
 	}};
 }
 
-#[macro_export]
 macro_rules! infoentry {
-	($str:expr, $rhs:expr) => ($crate::infoentry!($str, "{}", $rhs));
+	($str:expr, $rhs:expr) => (infoentry!($str, "{}", $rhs));
```
mkroening, line 60: Macro hygiene:

```suggestion
	($str:expr, $rhs:expr) => ($crate::infoentry!($str, "{}", $rhs));
```

src/logging.rs
```diff
@@ -25,7 +25,9 @@ impl log::Log for KernelLogger {
 }
 
 pub fn init() {
-	set_logger(&KernelLogger).expect("Can't initialize logger");
+	unsafe {
+		set_logger_racy(&KernelLogger).expect("Can't initialize logger");
+	}
```
mkroening, line 30: This is not sound, as this does not guarantee to uphold the safety requirements.

I think it would be easiest to make `init` unsafe and have the caller uphold the requirements.

mkroening (changes requested):

mkroening:
Aah, the macro hygiene does not work as suggested, since the macro exports have been removed. :facepalm: 

I'll adjust them.

src/logging.rs
stlankes: I don't export infoentry. Consequently, the compiler don't find `$crate::infoentry`

mkroening:
Hmm. Maybe it would be best to keep `macro_export`. Why do you want to remove it?

src/logging.rs
mkroening: Yeah, I noticed as well: https://github.com/hermitcore/libhermit-rs/pull/323#issuecomment-1003723286

stlankes:
Why do we export it?

src/macros.rs
stlankes: Fun fact... Without this line, my code doesn't work on aarch64. Hm, in general does the logging macro not work. I have to check it.

src/macros.rs
stlankes: I fixed it in my PR.

mkroening:
Okay, I just read up on [macro hygiene](https://doc.rust-lang.org/reference/macros-by-example.html#hygiene) again.

`macro_export` gives the macros path-based scope. Without it, there is only textual scope: because we declare the `macro` module with `macro_use` before the other modules, the macros are in textual scope of the other modules, which can use them.

Removing `macro_export` prevents these macros to be hygienic: Inner macros cannot be referenced by path, would be expanded as it and might not necessarily refer to the intended macro in the `macro` module.

I'd prefer keeping `macro_export` and keeping these macros modular, but you can also go ahead and remove all `macro_export`s, if you prefer.

src/logging.rs
stlankes: I fixed it.

src/macros.rs
stlankes: I fixed it in my PR

stlankes:
I will reopen the PR later

PR #324, opened 2022-01-02
synchronize aarch64 port to x86 port

stlankes:
- inline wrapper functions
- don't use a thread-safe version to initialize the kernel logger
  because at this stage the kernel doesn't provide threads
- simplifies the boot process

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -347,10 +347,16 @@ fn boot_processor_main() -> ! {
 	#[cfg(target_arch = "aarch64")]
 	{
 		info!("The current hermit-kernel is only implemented up to this point on aarch64.");
-		info!("Attempting to exit via QEMU.");
-		info!("This requires that you passed the `-semihosting` option to QEMU.");
-		let exit_handler = qemu_exit::AArch64::new();
-		exit_handler.exit_success();
+		if environment::is_uhyve() {
+			syscalls::init();
+			syscalls::__sys_shutdown(0);
+		} else {
+			info!("Attempting to exit via QEMU.");
+			info!("This requires that you passed the `-semihosting` option to QEMU.");
+			let exit_handler = qemu_exit::AArch64::new();
+			exit_handler.exit_success();
+		}
+
 		loop {} /* Compiles up to here - loop prevents linker errors */
 	}
 	arch::boot_processor_init();
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,7 +104,7 @@ mod syscalls;
 mod util;
 
 #[doc(hidden)]
-pub fn _print(args: ::core::fmt::Arguments<'_>) {
+fn _print(args: ::core::fmt::Arguments<'_>) {
 	use core::fmt::Write;
 	crate::console::CONSOLE.lock().write_fmt(args).unwrap();
 }
@@ -331,7 +331,9 @@ fn synch_all_cores() {
 fn boot_processor_main() -> ! {
 	// Initialize the kernel and hardware.
 	arch::message_output_init();
-	logging::init();
+	unsafe {
+		logging::init();
+	}
 
 	info!("Welcome to HermitCore-rs {}", env!("CARGO_PKG_VERSION"));
 	info!("Kernel starts at {:#x}", environment::get_base_address());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,7 +103,6 @@ mod synch;
 mod syscalls;
 mod util;
 
-#[doc(hidden)]
 fn _print(args: ::core::fmt::Arguments<'_>) {
 	use core::fmt::Write;
 	crate::console::CONSOLE.lock().write_fmt(args).unwrap();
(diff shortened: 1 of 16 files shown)


src/lib.rs
```diff
@@ -104,7 +104,7 @@ mod syscalls;
 mod util;
 
 #[doc(hidden)]
```
mkroening, line 106: Ah, good catch. We can remove the `doc(hidden)` attribute then.

```suggestion
```

src/logging.rs
```diff
@@ -56,7 +56,7 @@ macro_rules! infoheader {
 
 #[macro_export]
 macro_rules! infoentry {
-	($str:expr, $rhs:expr) => ($crate::infoentry!($str, "{}", $rhs));
+	($str:expr, $rhs:expr) => (crate::infoentry!($str, "{}", $rhs));
```
mkroening, line 59: This should stay `$crate`, right?

```suggestion
	($str:expr, $rhs:expr) => ($crate::infoentry!($str, "{}", $rhs));
```

src/macros.rs
```diff
@@ -18,16 +18,14 @@ macro_rules! align_up {
...
-	});
+	($($arg:tt)*) => ($crate::_print(::core::format_args!($($arg)*)));
 }
 
 /// Print formatted text to our console, followed by a newline.
 #[macro_export]
 macro_rules! println {
 	() => ($crate::print!("\n"));
-	($($arg:tt)+) => ($crate::print!("{}\n", format_args!($($arg)+)));
+	($($arg:tt)*) => ($crate::print!("{}\n", ::core::format_args!($($arg)*)));
```
mkroening, line 28: As `format_args!` requires at least one argument, we should still reflect that:

```suggestion
	($($arg:tt)+) => ($crate::print!("{}\n", ::core::format_args!($($arg)+)));
```

I can follow up with another PR to make this even more concise.

mkroening (commented):
Apart from a few small comments, this looks good to me! :+1:

stlankes:
@mkroening You are right. The macros are correct. But they doesn't work on aarch64. My bad implementations seems to work. I have no idea what is happening.

mkroening:
Which change exactly is breaking aarch64? :thinking:

stlankes:
bors try

src/lib.rs
```diff
@@ -103,8 +103,7 @@ mod synch;
 mod syscalls;
 mod util;
 
-#[doc(hidden)]
-pub fn _print(args: ::core::fmt::Arguments<'_>) {
+fn _print(args: ::core::fmt::Arguments<'_>) {
```
mkroening, line 106: Since this is used in the macros, it must be accessible from other modules: https://git.rwth-aachen.de/acs/public/hermitcore/libhermit-rs/-/jobs/2199434#L260

```suggestion
pub(crate) fn _print(args: ::core::fmt::Arguments<'_>) {
```

stlankes:
@mkroening Can you re-review it? I used the PR to publish my changes to the aarch64 port

stlankes:
bors try

src/logging.rs
mkroening: Or does this not work with `$crate`?

mkroening (commented):
Apart from one comment this looks good to me. :+1:

stlankes:
I have a problem with the println. On my Ubuntu system with Qemu (and aarch64 emulation) it works. On macOS with M1 and uhyve it fails. It seems that write_fmt in console.rs doesn't work. But I don't know why...

mkroening:
Your implementation of `write_fmt` is incorrect. [`Arguments::as_str`](https://doc.rust-lang.org/std/fmt/struct.Arguments.html#method.as_str) only returns a string, if no actual formatting is happening:

```rust
assert_eq!(format_args!("hello").as_str(), Some("hello"));
assert_eq!(format_args!("").as_str(), Some(""));
assert_eq!(format_args!("{}", 1).as_str(), None);
```

So with your implementation you can only write strings without any formatting like `println!("Hello World!")`. There might be additional problems somewhere though.

stlankes:
You are right, I used normally only the default implementation and this implementation doesn't work. I replaced it by following implementation:

```
fn write_fmt(&mut self, args: ::core::fmt::Arguments<’_>) -> fmt::Result {
        if let Some(s) = args.as_str() {
            self.write_str(s)
        } else {
            let s = alloc::fmt::format(args);
            self.write_str(s.as_str())
        }
    }
```

Same behavior like before. The else-path doesn't work.

mkroening:
Do allocated `Strings` in general not work?

stlankes:
No, on my M1 system the binary doesn't call the alloctor. The same binary works in Qemu.

stlankes:
It works now on my M1 system. The target specification is wrong. I have to use following command:

`RUSTFLAGS='-C target-feature=+strict-align' cargo build -p hello_world --target aarch64-unknown-hermit`

PR #327, opened 2022-01-05
[WIP] SCHEDULERS: Fix race condition

mkroening:
Fixes https://github.com/hermitcore/libhermit-rs/issues/301.

Vec::insert does not work when initializing core schedulers in the wrong order. The mutex synchronizes accesses to the map.

### DIFF
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -478,7 +478,7 @@ unsafe fn pre_init(boot_info: &'static mut BootInfo) -> ! {
 
 	BOOT_INFO = boot_info as *mut BootInfo;
 
-	atomic::fence(Ordering::SeqCst);
+	atomic::fence(Ordering::Acquire);
 	if boot_info.cpu_online == 0 {
 		crate::boot_processor_main()
 	} else {
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -443,7 +443,6 @@ fn finish_processor_init() {
 	// to initialize the next processor.
 	unsafe {
 		let _ = intrinsics::atomic_xadd(&mut (*BOOT_INFO).cpu_online as *mut u32, 1);
-		atomic::fence(Ordering::SeqCst);
 	}
 }
 
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -1,7 +1,6 @@
 use alloc::collections::BTreeMap;
 #[cfg(feature = "newlib")]
 use core::slice;
-use core::sync::atomic::{self, Ordering};
 use core::{intrinsics, ptr};
 
 use x86::controlregs::{cr0, cr0_write, cr4, Cr0};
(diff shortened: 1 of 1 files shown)


mkroening:
bors try

mkroening:
bors try

stlankes:
How do we wake-up the cores in a wrong order?

mkroening:
> How do we wake-up the cores in a wrong order?

I am not sure, actually, but it is what happens: https://github.com/hermitcore/libhermit-rs/issues/301

We have two vCPUs. The boot vCPU might be running on a core with a high load, so maybe the second application core is faster than the first one and tries to register it's scheduler before the first one because of that. But the actual cause is just a guess.

I am pretty confident, that this fixes the issue though. Current CI runs seem to confirm that.

mkroening:
bors try

stlankes:
I used https://github.com/hermitcore/libhermit-rs/blob/master/src/arch/x86_64/kernel/mod.rs#L447-L449 to synchronize the wakeup process. At first a core muss call this function, before we wake up the next core.

mkroening:
Good point. Hmm... I'll try something else then. :thinking:

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

src/arch/x86_64/kernel/mod.rs
```diff
@@ -481,7 +481,7 @@ unsafe fn pre_init(boot_info: &'static mut BootInfo) -> ! {
 
 	BOOT_INFO = boot_info as *mut BootInfo;
 
-	if boot_info.cpu_online == 0 {
+	if intrinsics::atomic_load_acq(&(*BOOT_INFO).cpu_online) == 0 {
```
stlankes, line 484: Maybe you should use https://github.com/hermitcore/libhermit-rs/blob/master/src/arch/x86_64/kernel/mod.rs#L254

mkroening: I tried atomically loading with acquire on purpose, but this does not seem the problem. :thinking:

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
Ugh. Not being able to reproduce the problem reliably, even in CI, is a nightmare.

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try

mkroening:
bors try

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try-

mkroening:
bors try

mkroening:
bors try

mkroening:
Superseded by https://github.com/hermitcore/libhermit-rs/pull/509.

PR #335, opened 2022-01-08
add option that COMMAND_LINE_CPU_FREQUENCY is invalid

stlankes:
If `parse_command_line` isn't able to identify the processor frequency, the function `get_command_line_cpu_frequency` will return `None` instead of a invalid frequency.

In addition, the frequency measurement with PIT is able to fail. In this case, we just guess a frequency of 2Ghz. For tests is guessing a good workaround.

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -19,7 +19,7 @@ use alloc::string::String;
 use alloc::vec::Vec;
 use core::{slice, str};
 
-static mut COMMAND_LINE_CPU_FREQUENCY: u16 = 0;
+static mut COMMAND_LINE_CPU_FREQUENCY: Option<u16> = None;
 static mut IS_PROXY: bool = false;
 static mut COMMAND_LINE_APPLICATION: Option<Vec<String>> = None;
 static mut COMMAND_LINE_PATH: Option<String> = None;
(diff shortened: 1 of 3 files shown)


src/environment.rs
```diff
@@ -95,8 +95,14 @@ pub fn init() {
...
 /// CPU Frequency in MHz if given through the -freq command-line parameter, otherwise zero.
-pub fn get_command_line_cpu_frequency() -> u16 {
-	unsafe { COMMAND_LINE_CPU_FREQUENCY }
+pub fn get_command_line_cpu_frequency() -> Option<u16> {
+	unsafe {
+		if COMMAND_LINE_CPU_FREQUENCY > 0 {
+			Some(COMMAND_LINE_CPU_FREQUENCY)
+		} else {
+			None
+		}
```
mkroening, line 104: I think, we should make `COMMAND_LINE_CPU_FREQUENCY` an `Option<u16>` itself, which is initialized with `None` by default.

mkroening (commented):
Ah, nice! I like this!

Apart from one comment, this looks good to me. :+1:

src/environment.rs
stlankes: Ok, I fixed it

PR #337, opened 2022-01-10
Allow for specifying static network configuration at launch

JTS22:
This PR makes it possible to overwrite the default network parameters for IP address, gateway and mask by appending them to the list of command line arguments. This is useful when the network configuration of the corresponding tap interface is not known at build time, for example when running in a container. When one or more values for the keys `-ip`, `-gateway` and `-mask` are specified, they are provided to the application as environment variables. This way, the hermit-sys network code can read these variables and configure the network device accordingly (see https://github.com/hermitcore/rusty-hermit/pull/204).


mkroening:
bors retry

stlankes:
bors retry

PR #343, opened 2022-01-15
Cleanup entry.asm remains

mkroening:
entry.asm was replaced with _start in d36871a7e6828ad72e6b063b1fd68246ada15b81.


stlankes (approved):
bors r+

PR #360, opened 2022-02-08
Reduce crate exports

mkroening:
Unfortunately macros can't have a path-based scope without being exported from the crate, so we'll have to resort to textual scope.


stlankes (approved):
bors r+

mkroening:
bors retry

PR #365, opened 2022-02-11
use latest nightly compiler with new hermit target

stlankes:


stlankes:
bors retry

stlankes:
bors retry

PR #367, opened 2022-02-13
Remove dead code

mkroening:
Closes https://github.com/hermitcore/libhermit-rs/issues/364.

### DIFF
--- a/src/arch/x86_64/mm/paging.rs
+++ b/src/arch/x86_64/mm/paging.rs
@@ -614,6 +614,7 @@ pub fn get_page_table_entry<S: PageSize>(virtual_address: VirtAddr) -> Option<Pa
 	root_pagetable.get_page_table_entry(page)
 }
 
+#[allow(dead_code)]
 pub fn get_physical_address<S: PageSize>(virtual_address: VirtAddr) -> PhysAddr {
 	trace!("Getting physical address for {:#X}", virtual_address);
 
--- a/src/drivers/net/virtio_mmio.rs
+++ b/src/drivers/net/virtio_mmio.rs
@@ -165,7 +165,7 @@ impl VirtioNetDriver {
 			match drv.init_dev() {
 				Err(error_code) => Err(VirtioError::NetDriver(error_code)),
 				_ => {
-					//drv.print_information();
+					drv.print_information();
 					Ok(drv)
 				}
 			}
--- a/src/syscalls/fs.rs
+++ b/src/syscalls/fs.rs
@@ -35,7 +35,6 @@
 /// - FileDescriptor newtype
 use crate::env::is_uhyve;
 use crate::synch::spinlock::Spinlock;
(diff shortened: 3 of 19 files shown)


src/arch/x86_64/mm/paging.rs
```diff
@@ -616,21 +616,6 @@ pub fn get_page_table_entry<S: PageSize>(virtual_address: VirtAddr) -> Option<Pa
 	root_pagetable.get_page_table_entry(page)
 }
 
-pub fn get_physical_address<S: PageSize>(virtual_address: VirtAddr) -> PhysAddr {
```
jounathaen, line 619: This fn is really not used? I could imagine it beeing helpful in debugging contexts

src/drivers/net/virtio_mmio.rs
```diff
@@ -166,7 +166,7 @@ impl VirtioNetDriver {
 			match drv.init_dev() {
 				Err(error_code) => Err(VirtioError::NetDriver(error_code)),
 				_ => {
-					//drv.print_information();
+					drv.print_information();
```
jounathaen, line 169: `print_information` can most probably removed with a debug format print. But this doesn't look like dead code removal to me

src/syscalls/fs.rs
```diff
@@ -160,11 +159,14 @@ impl Filesystem {
 	}
 
 	/// Create new backing-fs at mountpoint mntpath
+	#[cfg(feature = "pci")]
 	pub fn mount(
 		&mut self,
 		mntpath: &str,
 		mntobj: Box<dyn PosixFileSystem + Send>,
 	) -> Result<(), ()> {
+		use alloc::borrow::ToOwned;
```
jounathaen, line 168: Is this dead code removal?

src/arch/x86_64/mm/paging.rs
mkroening: It was introduced in https://github.com/hermitcore/libhermit-rs/commit/830bbc51740f97a0dca9a50c0ea30b9bc78290fc and was never used since, as far as I can tell. Do you want me to keep it for debugging purposes though?

src/syscalls/fs.rs
mkroening: Yes. The import should be conditional. Having the import inside the function is more readable than having it conditionally at the head of the file, since this is the only use of the import.

src/drivers/net/virtio_mmio.rs
mkroening: While technically not removing dead code, it removes the dead code warning of `print_information` and `dev_status`, by making that code live.

Those functions themselves intentionally print on the `info` level, which seems a good choice to me, similar to the other info reports on detected devices on boot.

mkroening:
Rebased on latest master.

mkroening:
bors r=stlankes

PR #376, opened 2022-02-27
add system calls to change the priority of a task

stlankes:
To avoid priority inversion, I added system calls to changes the priority. They will be used by the PR https://github.com/hermitcore/rust/commit/dd7e2582764844c2e53876658732977a249b3190.

### DIFF
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -642,10 +642,5 @@ pub fn join(id: TaskId) -> Result<(), ()> {
 }
 
 fn get_task_handle(id: TaskId) -> Option<TaskHandle> {
-	let guard = TASKS.lock();
-
-	match guard.get(&id) {
-		Some((task, _)) => Some(*task),
-		_ => None,
-	}
+	TASKS.lock().get(&id).map(|(task_handle, _)| *task_handle)
 }
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -639,5 +639,5 @@ pub fn join(id: TaskId) -> Result<(), ()> {
 }
 
 fn get_task_handle(id: TaskId) -> Option<TaskHandle> {
-	TASKS.lock().get(&id).map(|task_handle| *task_handle)
+	TASKS.lock().get(&id).copied()
 }
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -338,21 +338,9 @@ impl PerCoreScheduler {
 		irqsave(|| {
 			let task = get_task_handle(id).ok_or(())?;
 
-			#[cfg(feature = "smp")]
-			if task.get_core_id() == self.core_id {
-				if self.current_task.borrow().id == task.get_id() {
-					self.current_task.borrow_mut().prio = prio;
-				} else {
-					self.ready_queue
-						.set_priority(task, prio)
-						.expect("Do not find valid task in ready queue");
-				}
-			} else {
+			if cfg!(feature = "smp") && task.get_core_id() != self.core_id {
 				warn!("Have to change the priority on another core");
-			}
-
-			#[cfg(not(feature = "smp"))]
-			if self.current_task.borrow().id == task.get_id() {
+			} else if self.current_task.borrow().id == task.get_id() {
 				self.current_task.borrow_mut().prio = prio;
 			} else {
 				self.ready_queue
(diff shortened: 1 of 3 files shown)


stlankes:
@mkroening Do you think that this is a valide implementation?

src/scheduler/mod.rs
```diff
@@ -328,6 +340,45 @@ impl PerCoreScheduler {
...
+					self.ready_queue
+						.set_priority(task, prio)
+						.expect("Do not find valid task in ready queue");
+				}
+			} else {
+				warn!("Have to change the priority on another core");
+			}
+
+			#[cfg(not(feature = "smp"))]
+			if self.current_task.borrow().id == task.get_id() {
```
mkroening, line 370: We can reduce code duplication:

```suggestion
			if cfg!(feature = "smp") && task.get_core_id() != self.core_id {
				warn!("Have to change the priority on another core");
			} else if self.current_task.borrow().id == task.get_id() {
```

src/scheduler/mod.rs
```diff
@@ -328,6 +340,45 @@ impl PerCoreScheduler {
...
+			if task.get_core_id() == self.core_id {
+				if self.current_task.borrow().id == task.get_id() {
+					self.current_task.borrow_mut().prio = prio;
+				} else {
+					self.ready_queue
+						.set_priority(task, prio)
+						.expect("Do not find valid task in ready queue");
+				}
+			} else {
+				warn!("Have to change the priority on another core");
```
mkroening, line 366: Is printing a warning and failing silently otherwise correct here? :thinking: 
Should it be possible in the future to change the priority on other cores?
The alternatives would be to panic or to return an error (results in a panic when unwrapping in the  API).

src/scheduler/mod.rs
```diff
@@ -610,3 +667,12 @@ pub fn join(id: TaskId) -> Result<(), ()> {
...
 	Ok(())
 }
+
+fn get_task_handle(id: TaskId) -> Option<TaskHandle> {
+	let guard = TASKS.lock();
+
+	match guard.get(&id) {
+		Some((task, _)) => Some(*task),
+		_ => None,
+	}
```
mkroening, line 677:
```suggestion
	TASKS.lock().get(&id).map(|(task_handle, _)| *task_handle)
```

src/scheduler/mod.rs
```diff
@@ -22,8 +22,8 @@ pub mod task;
 static NO_TASKS: AtomicU32 = AtomicU32::new(0);
 /// Map between Core ID and per-core scheduler
 static mut SCHEDULERS: Vec<&PerCoreScheduler> = Vec::new();
-/// Map between Task ID and Task Control Block
-static TASKS: SpinlockIrqSave<BTreeMap<TaskId, VecDeque<TaskHandle>>> =
+/// Map between Task ID and TaskHandle + Queue of waiting tasks
+static TASKS: SpinlockIrqSave<BTreeMap<TaskId, (TaskHandle, VecDeque<TaskHandle>)>> =
```
mkroening, line 26: Why put TaskHandle and queue in the same map instead of having two maps for each? That would avoid the inconvenient extraction of each when using the map.

src/scheduler/task.rs
```diff
@@ -334,6 +334,53 @@ impl PriorityTaskQueue {
...
+	/// Change priority of specifc task
+	pub fn set_priority(&mut self, handle: TaskHandle, prio: Priority) -> Result<(), ()> {
+		let i = handle.get_priority().into() as usize;
+		let mut pos = self.queues[i].head.as_mut().ok_or(())?;
+
+		loop {
+			if handle.id == pos.borrow().id {
+				let task = pos.clone();
+
+				{
```
mkroening, line 347: Let's add a comment for clarity:

```suggestion
				// Extract found task from queue and set new priority
				{
```

src/scheduler/task.rs
```diff
@@ -334,6 +334,53 @@ impl PriorityTaskQueue {
...
+					borrow.prev = None;
+				}
+
+				self.push(task);
+
+				return Ok(());
+			}
+
+			let ptr = pos.as_ptr();
+			pos = unsafe { (*ptr).next.as_mut().ok_or(())? };
```
mkroening, line 381: This is unsound, but we probably can't do much better with our self-made linked list. I opened https://github.com/hermitcore/libhermit-rs/issues/377 for the future, so this does not block the PR.

mkroening (changes requested):
The general approach looks good to me, but I have a few comments here and there.

src/scheduler/mod.rs
stlankes: Yes, I want also to change the priority on other cores. I am working on it.

src/scheduler/mod.rs
stlankes: You are right. I will fix it.

mkroening (approved):
LGTM :+1:

r=me if you don't want to add anything else.

stlankes:
bors retry

PR #384, opened 2022-03-11
Implement early shutdown

mkroening:
This fixes panicking (exiting in general) before syscalls::init. 

This should fix CI timeouts on early kernel panics.


stlankes:
Why does the CI fail?

mkroening:
> Why does the CI fail?

Because we did not wait for ACPI shutdown to be complete before using QEMU's debug exit, which always returns a non-zero exit code. I fixed it in https://github.com/hermitcore/libhermit-rs/pull/384/commits/91588977f51758a206cf35e9f4d42708358f8813.

stlankes (approved):
bors r+

PR #395, opened 2022-03-23
Use custom targets for kernel

mkroening:
Closes https://github.com/hermitcore/rusty-hermit/issues/197#issuecomment-1076416891.

This puts our kernel targets into json files. This is now possible since we do not distribute the kernel via crates.io anymore.


stlankes (approved):
bors r+

PR #414, opened 2022-03-27
Targets: Use none-elf instead of hermit llvm-target

mkroening:
As discussed in https://github.com/hermitcore/rusty-hermit/issues/197#issuecomment-999379314, this way the kernel's `libhermit.a` has `OS/ABI: UNIX - System V` and the resulting application has `OS/ABI: <unknown: ff>`. This works for the Rust-toolchain as well as the C-toolchain.

### DIFF
--- a/targets/aarch64-unknown-none-hermitkernel.json
+++ b/targets/aarch64-unknown-none-hermitkernel.json
@@ -7,7 +7,7 @@
     "features": "+strict-align,-neon,-fp-armv8",
     "linker": "rust-lld",
     "linker-flavor": "ld.lld",
-    "llvm-target": "aarch64-unknown-hermit",
+    "llvm-target": "aarch64-unknown-none-elf",
     "max-atomic-width": 128,
     "panic-strategy": "abort",
     "position-independent-executables": true,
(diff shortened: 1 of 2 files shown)


stlankes:
How do change the OS/ABI in the static library? Or is it not an issue that you link two different object codes with different ABIs to a specific ABI version?

mkroening:
Okay, I made a mistake in my testing of the C-toolchain. It should still be okay though. To summarize:

* rusty-hermit: Changing the OS/ABI does not seem to be necessary. It just works.
* hermit-playground: Changing the OS/ABI is done [here](https://github.com/hermitcore/hermit-playground/blob/f99489cb190bebb8c1673185d8531f647c8747d6/cmake/HermitCore-librs.cmake#L73-L75).
  Without it, we get these errors:
  ```
  [...]
  /opt/hermit/lib/gcc/x86_64-hermit/6.5.0/../../../../x86_64-hermit/bin/ld: /volume/build/local_prefix/opt/hermit/x86_64-hermit/lib/libhermit.a(hermit-5b62641ff76e024f.hermit.45c2875d-cgu.15.rcgu.o): Relocations in generic ELF (EM: 62)
  /opt/hermit/lib/gcc/x86_64-hermit/6.5.0/../../../../x86_64-hermit/bin/ld: /volume/build/local_prefix/opt/hermit/x86_64-hermit/lib/libhermit.a: error adding symbols: file in wrong format
  ```
  It would probably be easiest to just keep the conversion in that repository though. Using `${CMAKE_ELFEDIT}` is easy there, since `hermit-playground` depends on our GNU toolchain (`x86_64-hermit-elfedit`) to be available anyway.

stlankes (approved):
bors r+

stlankes:
bors retry

mkroening:
@stlankes, see https://github.com/hermitcore/libhermit-rs/pull/427.

mkroening:
bors retry

PR #421, opened 2022-03-30
Remove unused sys_uhyve_* functions

mkroening:
They don't seem to be used anywhere and are problematic for C-ABI.

### DIFF
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -135,42 +135,6 @@ pub fn has_ipdevice() -> bool {
 	!(ip[0] == 255 && ip[1] == 255 && ip[2] == 255 && ip[3] == 255)
 }
 
-#[cfg(not(feature = "newlib"))]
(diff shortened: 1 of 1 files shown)


stlankes:
Doesn't the C user space uses these functions?

mkroening:
Hmm, something is wrong here.

These functions are not compiled on newlib. Seems like between https://github.com/hermitcore/libhermit-rs/commit/2f59ff5ad2115876410e50c388d6a837ce4196f2 and https://github.com/hermitcore/LwIP/commit/6ac625c120e2062606e5a7fed905a196fb0ccfd2 something went wrong. The lwIP seems to use `uhyve_get_gateway`, which does not even exist, only `sys_uhyve_get_gateway`, but not for lwIP. Do we have integration tests for lwIP? :thinking:

mkroening:
Ah, yes. You are correct. I made sure to only remove the implementations for Rust-Userspace, since they seem to be unused.

stlankes (approved):
bors r+

PR #422, opened 2022-03-30
Condvar: Mark sys functions as extern "C"

mkroening:
They are declared in hermit-abi as extern "C" already.
So the current definition without extern "C" is wrong.


stlankes (approved):
bors r+

mkroening:
bors retry

PR #423, opened 2022-03-30
Random: Make sys_secure_* functions extern "C"

mkroening:
Note that this changes and stabilizes the API of the affected two functions.


stlankes (approved):
bors r+

PR #430, opened 2022-04-24
Migrate to x86_64-unknown-none

mkroening:
Closes https://github.com/hermitcore/libhermit-rs/issues/412.

This reduces initial build times of the kernel, since `core` and `alloc` are now downloaded instead of compiled ourselves. `x86_64-unknown-none` is tier 2 and ships artifacts, so we can omit `build-std`.


stlankes (approved):
bors r+

PR #462, opened 2022-06-12
Migrate to hermit-entry

mkroening:
This PR adapts the unified boot info API from hermit-entry 0.1 with the same semantic as before.


mkroening:
Note that this modifies the link section of `RAW_BOOT_INFO`.

PR #463, opened 2022-06-13
Adopt aarch64-unknown-none-softfloat

mkroening:


mkroening:
See 5f81570e4866da388b414fbd80f18af956ba1889 for the differences to our own target.

PR #469, opened 2022-06-21
Enable lwIP support

mkroening:
This PR enables lwIP support by using lwIP's rtl8139 driver instead of the kernel implementation.

To workaround https://github.com/hermitcore/libhermit-rs/issues/471, this disables kernel stack switching with newlib. A proper fix would require much more work and insight into our lwIP implementation.

### DIFF
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -64,13 +64,6 @@ pub fn raw_boot_info() -> &'static RawBootInfo {
 static mut COM1: SerialPort = SerialPort::new(0x3f8);
 
 #[cfg(feature = "newlib")]
-pub fn has_ipdevice() -> bool {
-	let ip = boot_info().net_info.ip;
-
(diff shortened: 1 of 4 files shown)


stlankes:
Can you remove the conflict?

You use the driver of libhermit-rs and not the old C version. But why is the stack an issue? Do you have a feeling? Is virtio support also useful?

mkroening:
This PR uses the old lwIP rtl8139 driver for the C userspace and not the Rust one in the kernel. I can't say why the stack is an issue before deep diving into and reworking our lwIP stack. Our lwIP stack has no VirtIO driver, so that would be difficult before switching to the kernel drivers. Let's talk about the lwIP situation before considering to merge this.

stlankes (approved):
bors r+

mkroening:
I did a rebase if you are sure with merging already...

stlankes:
We can do it!

mkroening:
> We can do it!

Okay then. :+1: 
bors r=stlankes

mkroening:
bors r=stlankes

PR #477, opened 2022-06-27
RTL8139: Improvement of the ring buffer management

stlankes:
The device driver RTL8139 tries to avoid unneeded copies of buffers. But in case, the ring buffer wrap around, two pieces have to copy into one continuous memory region.

=> temporary creation of a boxed buffers

### DIFF
--- a/src/drivers/net/rtl8139.rs
+++ b/src/drivers/net/rtl8139.rs
@@ -337,7 +337,7 @@ impl NetworkInterface for RTL8139Driver {
 		match self.box_map.remove_entry(&self.rxpos) {
 			Some((_, raw)) => {
 				// release temporay created boxed slice
-				let buffer: Box<[u8]> =
+				let _buffer: Box<[u8]> =
 					unsafe { Box::from_raw(slice::from_raw_parts_mut(raw.ptr, raw.len)) };
 			}
 			None => {}
--- a/src/drivers/net/rtl8139.rs
+++ b/src/drivers/net/rtl8139.rs
@@ -2,9 +2,10 @@
 
 #![allow(dead_code)]
 
-use core::mem;
-
 use alloc::boxed::Box;
+use alloc::collections::BTreeMap;
+use core::mem;
+use core::slice;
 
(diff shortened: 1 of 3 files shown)


mkroening (changes requested):
I"m not too familiar with the drivers logic, but I opened https://github.com/stlankes/libhermit-rs/pull/47 with some changes:

1. The change in `Cargo.toml` needs to be reflected in `Cargo.lock`
2. I think we can avoid `RawSlice` altogether by referencing the boxed slice directly.

The rest looks good to me. :+1:

stlankes:
@mkroening Is this a nicer solution?

mkroening:
> @mkroening Is this a nicer solution?

Yes, but we can still do better by completely avoiding reconstructing a slice as shown here: https://github.com/stlankes/libhermit-rs/pull/47

src/drivers/net/rtl8139.rs
```diff
@@ -277,10 +279,28 @@ impl NetworkInterface for RTL8139Driver {
...
+				let pos = (self.rxpos + mem::size_of::<u16>()) % RX_BUF_LEN;
+
+				// do we reach the end of the receive buffers?
+				// in this case, we have to copy the data in boxed slice
+				let buf = if pos + length as usize > RX_BUF_LEN {
+					let mut msg: Box<[u8]> = vec![0; length as usize].into_boxed_slice();
+					let limit = RX_BUF_LEN - pos;
+
+					msg[0..limit].copy_from_slice(&self.rxbuffer[pos..RX_BUF_LEN]);
+					msg[limit..].copy_from_slice(&self.rxbuffer[0..length as usize - limit]);
```
mkroening, line 291: We can also avoid to zero the buffer by using `[T]::concat`:

```suggestion
					let first = &self.rxbuffer[pos..RX_BUF_LEN];
					let second = &self.rxbuffer[..length as usize - first.len()];
					let msg = [first, second].concat().into_boxed_slice();
```

stlankes:
Thanks to @mkroening

PR #494, opened 2022-07-23
Disable exponential backoff for non-smp machines

stlankes:
Exponential backoff is only usefull for SMP machines because another
task has to release the semaphore. Without a context switch, this task
must run on another core.


mkroening (approved):
Awesome! :tada: 

bors r+

PR #495, opened 2022-07-24
add option to assign a task to the nic

stlankes:
the new syscall allows us to handle the network interrupt on the core, where the network thread (see hermit-sys) is running


mkroening (approved):
Nice!

bors r+

PR #509, opened 2022-07-26
Correctly sync all cores

mkroening:
Closes https://github.com/hermitcore/libhermit-rs/issues/506.

This requires uhyve 0.1.3 to work correctly, but is backwards compatible to the racy uhyve version.


mkroening:
bors try

PR #511, opened 2022-07-26
Remove multikernel stubs

mkroening:
Since multikernel support is nonexistent as far as I know, we might remove any remains of older multikernel plans.


mkroening:
bors r=stlankes

PR #520, opened 2022-07-30
move smoltcp support into lib hermit-rs

stlankes:
The main motivation is to simplify the timer handling.


stlankes:
In principle, it is just a back port of the hermit-sys implementation. It offers an easier handling of network timeouts.

PR #534, opened 2022-08-16
Add futex syscalls

joboet:
[Futexes](https://man7.org/linux/man-pages/man2/futex.2.html) greatly simplify the creation of synchronization primitives (see [`std`'s mutex](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/locks/futex_mutex.rs)). They are here implemented by using a hashmap, in which the handles to waiting threads are stored, keyed with the address of the futex they are waiting on. 

I prioritised readability over performance in this PR, faster implementations would use finer grained locking and would not need to allocate the queues. It should be performant enough for most purposes however.


src/synch/futex.rs
```diff
@@ -0,0 +1,106 @@
...
+		/// Use a relative timeout
+		const RELATIVE = 0b01;
+	}
+}
+
+/// If the value at address matches the expected value, park the current thread until it is either
+/// woken up with `futex_wake` (returns 0), the specified timeout elapses (returns -ETIMEDOUT) or
+/// a spurious wakeup occurs (returns -EINTR).
+///
+/// The timeout uses processor ticks. If [`Flags::RELATIVE`] is given, it is interpreted as relative
```
mkroening, line 28: Is the timeout not in microseconds as indicated in `__sys_futex_wait`?

mkroening (approved):
Looks good to me!
Thanks a lot for your effort of migrating to the generic futex-based impls in `std`. :)

mkroening:
bors try

stlankes (commented):
Excuse me for the late review. I was ill. Like @mkroening, I have the same question __sys_futex_wait.

The rest looks good for me! Thank you for your great support.

joboet:
I've now clarified the timeout behaviour in the documentation. Also, `futex_wait` now does not return spuriously in order to match the behaviour on modern Linux systems.

mkroening (approved):
Thanks a lot, @joboet! :)

bors r+

PR #539, opened 2022-09-05
Fix deadlock on spurious wakeup in `futex_wait`

joboet:
If a spurious wakeup occurs, the task was removed unconditionally from the queue, meaning it would never be woken. This is fixed if the task is only removed if a timeout occurred.

### DIFF
--- a/src/scheduler/task.rs
+++ b/src/scheduler/task.rs
@@ -164,6 +164,13 @@ impl TaskHandlePriorityQueue {
 		self.prio_bitmap == 0
 	}
 
+	/// Checks if the given task is in the queue. Returns `true` if the task
+	/// was found.
+	pub fn contains(&self, task: TaskHandle) -> bool {
+		self.queues[task.priority.into() as usize]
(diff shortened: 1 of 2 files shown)


src/scheduler/task.rs
```diff
@@ -164,6 +164,13 @@ impl TaskHandlePriorityQueue {
 		self.prio_bitmap == 0
 	}
 
+	/// Checks if the given task is in the queue. Returns `true` if the task
+	/// was found.
+	pub fn find(&self, task: TaskHandle) -> bool {
```
mkroening, line 169: `contains` might be a name more consistent with `std`. What do you think?

mkroening (changes requested):
Just a small comment on a function name. The rest looks good to me. :)

src/scheduler/task.rs
joboet: That sounds more fitting!

mkroening (approved):
bors r+

PR #586, opened 2022-11-18
Rework interrupts

mkroening:


mkroening:
Do we want to use ISTs for IRQs?

stlankes (approved):
bors r+

PR #633, opened 2022-12-18
add BSD socket layer

stlankes:
This PR realize a classical BSD socket layer. To realize a BSD socket layer, the executor has to return directly UNIX-like error number because the layer depends on a C interface and doesn't support Rust's error handling.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,7 @@ dependencies = [
  "hermit-sync",
  "include-transformed",
  "linked_list_allocator",
+ "lock_api",
  "log",
  "multiboot",
  "num",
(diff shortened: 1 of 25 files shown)


stlankes:
@mkroening A network interrupt has to wakeup all async tasks to check their tasks. Currently, I use a BtreeMap to store all blocked async tasks. The interrupts clears the map and [wakeup](https://github.com/stlankes/libhermit-rs/blob/fd/src/net/executor.rs#L31-L38) all tasks. Do you have an better idea?

mkroening (approved):
Looks good to me. 👍 

I also opened two PRs regarding `RwLock`s:
* https://github.com/cmnord/pflock/pull/9
* https://github.com/hermitcore/hermit-sync/pull/8

They don't block this PR though. We can decide between them later.

stlankes:
bors try

PR #637, opened 2022-12-20
Rewrite VGA text driver

mkroening:


mkroening:
I never finished the crate. I did some upstream work on [oem_cp](https://docs.rs/oem_cp) that did not go to waste, though.

PR #640, opened 2023-01-21
Reimplement random data generation, add `read_entropy` syscall

### LINKED ISSUE #143: Improve randomness syscalls
The RDRAND code used in `generate_random_number32/64` does not contain a workaround for the AMD glitch. You can take a look at the [`getrandom` code](https://github.com/rust-random/getrandom/blob/master/src/rdrand.rs#L30-L46) for reference.

Also it would be nice if instead of `generate_random_number32/64` you will expose a Linux-like interface, i.e. a syscall which will work on buffers.

joboet:
Fixes #143 by reimplementing random data generation using a [ChaCha-based RNG](https://docs.rs/rand_chacha/0.3.1/rand_chacha/struct.ChaCha20Rng.html) continuously reseeded using the `RDSEED` instruction. This should provide better security, as `RDRAND` is known to have hardware bugs. 

Also adds a new buffer-based syscall, `read_entropy`, which better fits the usecase of crates like [`getrandom`](https://github.com/rust-random/getrandom).

The old `secure_rand*` and `rand` syscalls should probably be removed at some point, but I do not know the Hermit stability policy and therefore have not done this in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -238,6 +238,7 @@ dependencies = [
  "num-traits",
  "pci-ids",
  "qemu-exit",
+ "rand_chacha",
  "scopeguard",
  "shell-words",
  "smoltcp",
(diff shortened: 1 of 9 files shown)


stlankes:
Thanks for your contribution! I revise the IP interface and will publish a new ABI interface. In this context I will remove the old interface.

PR #656, opened 2023-02-14
feat: remove spinlock syscalls

mkroening:
This is only used in [`hermit/mailbox.h`](https://github.com/hermit-os/hermit-playground/blob/b695beb522b6754ba8ad31b6541b57be0b6af6dd/include/hermit/mailbox.h), which is only used in [`lwip/src/include/arch/sys_arch.h`](https://github.com/hermit-os/LwIP/blob/d4d77f78a667d9773a235b010027ce1ec7973daf/src/include/arch/sys_arch.h).


mkroening:
> LGTM, since only lwip uses this feature and hermit-playground is removing lwip.

Are they already removed? We should wait until then. :)

CarlWachter (approved):
See https://github.com/hermit-os/hermit-playground/pull/27

jounathaen:
This is ready to be merged now, right?

CarlWachter:
> This is ready to be merged now, right?

Jep

mkroening:
I have created https://github.com/hermit-os/pthread-embedded/pull/7 to remove the last relevant definitions. They are unused already, though. So I'll rebase and merge.

PR #662, opened 2023-02-27
update Cargo.lock to remove Dependabot alert #1

stlankes:
See [dependabot alert](https://github.com/hermitcore/libhermit-rs/security/dependabot/1)


mkroening (approved):
bors r+

PR #664, opened 2023-03-06
Add a --randomize-layout option

mkroening:


stlankes (approved):

stlankes:
Is it still a draft?

mkroening:
> Is it still a draft?

We could add it as is, but I'd like to run some tests before merging. Marking as draft.

PR #689, opened 2023-03-22
define the same 32bit gap like uhyve

stlankes:
the gap is used to realize for memory-mapped IO (e.g. APIC)


stlankes:
bors r=mkroening

PR #694, opened 2023-03-28
removed shutdown

richiewahidin:
Fixes https://github.com/hermitcore/libhermit-rs/issues/670.

There are several different similar looking functions in the whole codebase, so I looked at issue #384 which was referenced to me when I was assigned the issue. I looked at the commits, one of them being implementing this early shutdown, which was in src/syscalls/mod.rs. I therefore removed the shutdown function from that file.


mkroening:
Ah, seems like I have to manually reapprove the CI workflow for first-time contributors. Let's see if we could change that, so your changes get tested automatically.

Issue #670, opened 2023-03-14
Remove internal `shutdown` function


mkroening:
This has been introduced because of improper initialization of `SYS`, which has since been fixed.

richiewahidin:
Hi. I am a student in UT Austin, and I am interested in working on this issue for my Virtualization project. Can you assign this issue to me?

mkroening:
Sure!

For reference:
- https://github.com/hermitcore/libhermit-rs/pull/384
- https://github.com/hermitcore/libhermit-rs/pull/600

mkroening:
Fixed by https://github.com/hermitcore/libhermit-rs/pull/695.

PR #695, opened 2023-03-28
removed all instances of shutdown

richiewahidin:
Fixes issue #670.

I removed the shutdown function in src/syscalls/mod.rs as a previous issue #384 had one of the commits implementing early shutdown which was that shutdown function in that particular file. I also removed instances when that shutdown was called, which was in the panic handler, in src/lib.rs, and replaced the shutdown syscall call with crate::__sys_shutdown(1).


mkroening (approved):
Thanks a lot; this looks good to me!

bors r+

PR #699, opened 2023-04-02
remove obsolete ist0 handling

stlankes:
IST0 isn't used by common interrupt. Just to close the session.

=> don't create IST0 for every task, just for every core


stlankes:
bors try

PR #704, opened 2023-04-09
remove workaround, which fix wrong file flags

stlankes:
- since rust-lang/rust#109368 the workaround isn't longer required


mkroening (approved):
🥳

Issue #686, opened 2023-03-21
Fix offsetof of CoreLocal


mkroening:
We are currently relying on the fact that the self-reference has the offset zero in the struct, but CoreLocal is not `repr(C)`!
We should make it `repr(C)`.

nathanwhyte:
Hi again Martin,

Can I claim this issue?

mkroening:
Sure! 👍

PR #705, opened 2023-04-09
Make CoreLocal repr(C)

nathanwhyte:
Addresses #686

Add the `repr(C)` layout specifier to the CoreLocal struct.


mkroening (approved):
Thanks a lot!

bors r+

mkroening:
Oh, no. CI currently fails because of the current, hermit-entry version bump dance. This should be fixed in the next couple of days.

CC: @stlankes

mkroening:
As a side note: in the future, we might want to explore making this `repr(rust)` again (https://github.com/hermitcore/libhermit-rs/issues/710). But this fix is definitely needed for correctness meanwhile.

mkroening:
bors retry

PR #706, opened 2023-04-12
Implements PriorityTaskQueue with std's LinkedList rather than a custom linked list implementation

michaelRichards99:
Fixes https://github.com/hermitcore/libhermit-rs/issues/377.

### DIFF
--- a/src/scheduler/task.rs
+++ b/src/scheduler/task.rs
@@ -237,33 +237,18 @@ impl TaskHandlePriorityQueue {
 	}
 }
 
-#[derive(Default)]
-struct QueueHead {
-	head: Option<Rc<RefCell<Task>>>,
-	tail: Option<Rc<RefCell<Task>>>,
-}
-
-impl QueueHead {
-	pub const fn new() -> Self {
-		Self {
-			head: None,
-			tail: None,
-		}
-	}
-}
-
 /// Realize a priority queue for tasks
 pub struct PriorityTaskQueue {
-	queues: [QueueHead; NO_PRIORITIES],
+	queues: [LinkedList<Rc<RefCell<Task>>>; NO_PRIORITIES],
 	prio_bitmap: u64,
 }
 
 impl PriorityTaskQueue {
 	/// Creates an empty priority queue for tasks
 	pub const fn new() -> PriorityTaskQueue {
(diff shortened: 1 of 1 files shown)


stlankes:
Looks good! But I will discuss it with @mkroening Just a double check.

src/scheduler/task.rs
```diff
@@ -356,49 +305,22 @@ impl PriorityTaskQueue {
...
-						self.queues[i].tail = borrow.prev.as_ref().cloned();
-					}
-
-					borrow.prio = prio;
-					borrow.next = None;
-					borrow.prev = None;
-				}
+		for (index, current_task) in self.queues[i].iter().enumerate() {
+			// Move the task from its old list to the new list.
+			if handle.id == current_task.borrow().id {
```
mkroening, line 311: This could be simplified by using `Iterator::find`.

src/scheduler/task.rs
```diff
@@ -356,49 +305,22 @@ impl PriorityTaskQueue {
...
+				// Calling remove is unstable: https://github.com/rust-lang/rust/issues/69210
+				let mut split_list = self.queues[i].split_off(index);
+				let task = split_list.pop_front().ok_or(())?;
+				self.queues[i].append(&mut split_list);
```
mkroening, line 315: We might want to extract this code into a function (be it inline or outside this method) for clarity.

mkroening (changes requested):
This looks good! Thanks a lot!

Two minor remarks, though the current code is definitely an improvement already. 👍

michaelRichards99:
Thank you for the feedback :). We've addressed those 2 changes and made sure the priority bitmap is properly set when removing a task from a list. 

Thank you @ynchen2829 for authoring these changes.

src/scheduler/task.rs
```diff
@@ -272,6 +272,30 @@ impl PriorityTaskQueue {
 		task
 	}
 
+	/// Remove the task at index from the queue and return that task,
+	/// or None if the index is out of range or the list is empty.
+	fn remove_from_queue(
```
mkroening, line 277: Ah, nice! You even made this a fully fledged method. 👍

mkroening (approved):
Looks good to me!

Thanks a lot! :)

bors r+

PR #715, opened 2023-04-20
Replace use of internal syscalls

### LINKED ISSUE #676: Don't use syscalls internally
For example, `crate::__sys_malloc` should be replaced by appropriate calls to the `alloc` crate.

https://github.com/hermitcore/libhermit-rs/blob/2203b5ae3dfc247c34098003c47ea1af2613cb66/src/syscalls/interfaces/uhyve.rs#L145

nathanwhyte:
Closes #676 

I found no internal uses of `__sys_realloc()` or `__sys_free()` anywhere.

I did find one reference to `__sys_shutdown()`:
https://github.com/hermitcore/libhermit-rs/blob/df7483de1b762c570f5d52e48fc396cdb739a12d/src/arch/x86_64/kernel/processor.rs#L832-L836

It doesn't redirect to any functions in [`lib.rs`](https://github.com/hermitcore/libhermit-rs/blob/master/src/lib.rs) like `__sys_malloc()` does. Does it need to be replaced as well?

### DIFF
--- a/src/syscalls/interfaces/uhyve.rs
+++ b/src/syscalls/interfaces/uhyve.rs
@@ -1,3 +1,4 @@
+use alloc::alloc::{alloc, Layout};
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::mem;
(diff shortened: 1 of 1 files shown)


mkroening (approved):
Thanks! This looks good to me 👍 

bors r+

Regarding shutdown: I opened https://github.com/hermitcore/libhermit-rs/issues/718 to track that.
I also opened https://github.com/hermitcore/libhermit-rs/issues/717 for moving the `sys_alloc` functions as a follow-up.

Issue #353, opened 2022-02-01
Rework rustdocs


mkroening:
* Item visibility
* Documentation of system calls

nathanwhyte:
Hi Martin,

I'm taking a Virtualization course at UT Austin and we've been asked to contribute to open-source virtualization repositories for one of our projects. Could I take a shot at resolving this issue?

mkroening:
Sure! More documentation is always welcome 👍

cagatay-y:
There are quite a few links in the docs that were broken after code reorganization as the file URLs were invalidated. We can fix at least some of these by using [intra-doc linking](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html) now that it's available. We can use [cargo-deadlinks](https://crates.io/crates/cargo-deadlinks) as a CI check to avoid such a situation in the future or at least for finding the broken links during the migration. We can also use [cargo-intraconv](https://crates.io/crates/cargo-intraconv) to convert the old style links that currently work but are susceptible to breakage in a future refactoring.

cagatay-y:
For example in this line linked struct was renamed from `SplitVqueue` to  `SplitVq`, which broke the link.
https://github.com/hermit-os/kernel/blob/a2f521fe071d11a080071ab7565a789d983212fc/src/drivers/virtio/virtqueue/mod.rs#L4
Another example is the structs getting moved to another file and the relative URL changing from `structs.PackedVq.html` to `packed/structs.PackedVq.html`
https://github.com/hermit-os/kernel/blob/a2f521fe071d11a080071ab7565a789d983212fc/src/drivers/virtio/virtqueue/mod.rs#L111

cargo-deadlinks lists 73 broken file:// URLs inside the `target/doc/hermit` directory after a `cargo doc` run with `--document-private-items`. I didn't look into all examples not to get too sidetracked.

To be clear, the two aforementioned packages would not be introduced as dependencies: I linked them partly to document them for later when I get a chance to look into the issue. Running cargo-intraconv after converting the existing links should not be necessary, as I would expect documentation added from now on to use intra-doc links from the get-go. cargo-deadlinks should also not be necessary, unless it turns out some of the file:// links cannot be converted to intra-doc links. Even in that case, it is not really necessary to make it a part of the CI workflow, since the kinds of changes that cause link breakages shouldn't be frequent.

mkroening:
Okay, makes sense.

[...]

PR #719, opened 2023-04-23
Add basic docs for timer and semaphore syscalls

nathanwhyte:
Helping with issue #353

Added _(very)_ basic documentation to syscalls in [`timer.rs`](https://github.com/hermitcore/libhermit-rs/blob/master/src/syscalls/timer.rs) and [`semaphore.rs`](https://github.com/hermitcore/libhermit-rs/blob/master/src/syscalls/semaphore.rs)

I used Linux man pages as a reference when writing these.

### DIFF
--- a/src/syscalls/fs.rs
+++ b/src/syscalls/fs.rs
@@ -141,6 +141,9 @@ impl Filesystem {
 		Ok(self.add_file(file))
 	}
 
+	/// Closes a file with given fd.
+	/// If the file is currently open, closes it
+	/// Remove the file from map of open files
 	pub fn close(&mut self, fd: u64) {
 		debug!("Closing fd {}", fd);
 		if let Some(file) = self.files.get_mut(&fd) {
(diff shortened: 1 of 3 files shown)


mkroening (changes requested):
Thanks for working on this. :)

Generally, this looks good. I would like these docs to be a bit more *rusty*, though. Specifically, an items documentation should start with a short sentence, then have an empty line, and then go into details. The details should also be relevant for API usage and not be implementation details. Comments about implementation details are welcome as well, but should be marked as documentation. Implementation details often fit better inside the function to where stuff happens, but might also be fit for placement outside the function.

Made-up example:
```rust
/// Yields to the executor.
///
/// This function can be used for ...
// We save the task handle in ... and reschedule.
extern "C" fn sys_yield () {
```

For more information see:
- https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html
- https://rust-lang.github.io/api-guidelines/documentation.html
- The standard library documentation as example.

PS: Remember punctuation at the end of sentences in the doc comments.

stlankes (approved):

nathanwhyte:
Thank you to everyone for being accommodating while I was working on these issues. Y'all saved me from a bad project grade by helping me solve them! 😆

PR #725, opened 2023-04-27
don't initialize IO-APIC for uhyve

stlankes:
uhyve doesn't support a common IO-APIC. Consequently, libhermit should not initialize the IO-APIC


jounathaen:
It might be a hotfix, but in that file there is still the `detect_from_uhyve()` fn, which probably should be thrown out as well.

jounathaen:
LGTM!

PR #729, opened 2023-04-29
remove typo

stlankes:


mkroening:
This is not a typo, though? 👀

stlankes:
:-)

Should I call it `envs`?

PR #736, opened 2023-05-01
revise TLS handling

stlankes:
- ARM used variant 1 of the TLS handling (see https://uclibc.org/docs/tls.pdf)
- the previous version doesn't work correctly
- using an unsized struct to implement TLS like a C kernel


stlankes:
@mkroening We should discuss it in our next meeting. I need this PR to extend the aarch64 support.

PR #765, opened 2023-06-15
AArch64: use a software-generated interrupt (SGI)to reschedule the system

stlankes:
On aarch64, an SGI is used to reschedule the system. However, this requires changes to schedule interface. Beside the interrupt handlers, the function `schedule` should not be called directly. Please use always `reschedule` to call the scheduler.


mkroening (approved):
LGTM! 👍

PR #769, opened 2023-06-20
`join`: Only continue if joinee finished

mkroening:


stlankes:
Why do you need loop?

mkroening:
> Why do you need a loop?

To make sure that the joinee really terminated—to avoid that we continue although the joinee did not finish yet.

This is essentially the same code as before if everything is well and an extra safeguard if something is wrong.

stlankes (approved):

PR #779, opened 2023-06-26
fix(udp): fix udp create error.

cathaysia:
fix https://github.com/hermitcore/libhermit-rs/issues/778


cathaysia:
udp has some error. this just fix udp bind error

stlankes (approved):
Thanks for the support!

PR #786, opened 2023-07-07
use write_unaligned to write data to the pci device

stlankes:
gurantees that the kernel 16-/32-bit registers to write the data to the device memory

### DIFF
--- a/src/drivers/virtio/transport/mmio.rs
+++ b/src/drivers/virtio/transport/mmio.rs
@@ -290,9 +290,16 @@ impl NotifCtrl {
 	}
 
 	pub fn notify_dev(&self, notif_data: &[u8]) {
-		let data = u32::from_ne_bytes(notif_data.try_into().unwrap());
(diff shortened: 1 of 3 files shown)


mkroening (changes requested):
I think, we might even have to use [`core::intrinsics::unaligned_volatile_store`](https://doc.rust-lang.org/nightly/core/intrinsics/fn.unaligned_volatile_store.html) here.

stlankes:
In this case, we have to enable the nightly feature _core_intrinsics_. Do we want it?

mkroening:
> In this case, we have to enable the nightly feature _core_intrinsics_. Do we want it?

We'd rather not, but I am uncertain if using inline assembly to emulate the intrinsic would be better. Maybe we should go for `core_intrinsics`. 🤔

stlankes:
Ok, I enabled the feature

mkroening (approved):

PR #799, opened 2023-07-18
revise handling of asynchronous tasks

stlankes:
- remove dependencies to async_tasks and futures_lite
- add on each core a vector with asynchrounous background tasks


src/arch/aarch64/kernel/core_local.rs
```diff
@@ -47,10 +52,19 @@ impl CoreLocal {
...
 	}
+
+	#[inline]
+	pub fn get_mut() -> &'static mut Self {
+		unsafe {
+			let raw: *mut Self;
+			asm!("mrs {}, tpidr_el1", out(reg) raw, options(nomem, nostack, preserves_flags));
+			&mut *raw
+		}
+	}
```
mkroening, line 63: This is unsound and should be removed. This allows creating multiple mutable references from thin air.

It should be sufficient to work with shared references and work with `RefCell` if you need mutable references to the insides. Note that this is then `!Sync` and can't be shared between threads, but that should be fine, right?

mkroening (changes requested):
This looks exciting! :)

src/arch/aarch64/kernel/core_local.rs
stlankes: I used it only on a core. I don't share it between core. By using this variable, I disable interrupts because the interrupt is also able to use it.

mkroening: Yes, but even on one core, you can create multiple mutable references with this. If only on one core, `RefCell` is exactly what you want for going from shared to mutable reference.

It should work if you return `RefMut<'static, Vec<AsyncTask>>` from `async_tasks` instead of `&'static mut Vec<AsyncTask>`

mkroening: Note that core_sheduler returning a mutable reference is bad as well, but I did not have time to fix that yet. (https://github.com/hermitcore/libhermit-rs/issues/622)

stlankes: I am confused. Why do I get a borrow_mut error? (see last run).

stlankes: I use `crate::executor::run()` only with disabled interrupts.

mkroening: > I am confused. Why do I get a borrow_mut error? (see last run).

Maybe `spawn` gets called by `new`? Or maybe `CoreLocal` is not working correctly? I am not sure about the specifics, but you could replace `retain_mut` with a manual loop and drop the `RefMut` before polling a task.

If you want me to look into this more deeply instead, I could do so next week. :)

mkroening: We still need to get rid of `CoreLocal::get_mut`. Using `get` should be sufficient. `RefCell` still allows mutation behind a shared reference.

stlankes: I am confused. How do I add a task (`spawn`) without a mutable reference?

mkroening: `RefCell` allows going from `&` to `RefMut`, which is a mutable reference with dynamic borrow checking. This only works on one thread, though, but this is the case for `CoreLocal`.

So just using `CoreLocal::get` in `async_tasks` should work. It still returns a `RefMut`, which can be used to mutate the `Vec`. Then we should be able to remove `CoreLocal::get_mut` again.

mkroening: I addressed it myself in https://github.com/hermitcore/libhermit-rs/pull/799/commits/342e631ed37de8b108ccc288f1dc6ca18c486149. Merging now. :)

mkroening (approved):

PR #802, opened 2023-07-22
CI: Increase sleep time from 5 to 10 seconds

mkroening:
I hope this helps with the frequent CI failures on aarch64.


stlankes:
It doesn't help :-(

mkroening:
> It doesn't help :-(

You are right. :/

It looks like Hermit gets stuck after logging the MTU... 🤔

PR #809, opened 2023-07-28
Migrate to talc memory allocator

mkroening:
Closes https://github.com/hermitcore/libhermit-rs/issues/630.

https://github.com/hermitcore/libhermit-rs/pull/803 has even further enhancements, but this might be a first step.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,7 +353,6 @@ dependencies = [
  "hermit-entry",
  "hermit-sync",
  "include-transformed",
- "linked_list_allocator",
  "lock_api",
  "log",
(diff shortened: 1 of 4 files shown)


stlankes:
Should we not remove the bump allocator?

stlankes (approved):
Should we not remove the bump allocator? But it is a good first step.

mkroening:
Yes. I was interested in the AArch64 run failure and figured a fix and simple migration might be good for comparison.

So I'll merge this as a first step. :)

Removing the bump allocator might make sense, but measuring that on its own would be interesting.

PR #810, opened 2023-07-28
network drivers create directly TxToken / RxToken for smoltcp 

stlankes:
For this purpose, it was necessary to avoid the creation of dynamic objects of the driver. This means that during compilation it must be defined which network interface is used. If the feature `rtl8139` is activated, the driver for Realtek 8139 will be used. Otherwise the Virtio network interface will be used.


mkroening (approved):

mkroening (changes requested):
This reduces read speed in our bandwidth test to 80 Mbit/s, according to @n-eiling.

mkroening:
Ah, makes sense. Still, without this PR, we can reach 191.4 MiB/s, and with it we can only reach 77.1 MiB/s.

mkroening (approved):
I could confirm that we can sometimes get good performance with this.

PR #834, opened 2023-08-10
add support of TCP segmentation offloading

stlankes:


stlankes:
This PR depends on smoltcp-rs/smoltcp#830

stlankes:
I have to revise it.

PR #845, opened 2023-08-15
avoid the usage of unneeded msr accesses

stlankes:


mkroening (approved):
Looks good to me.

I was confused, though, that MSR is emulated by KVM? Did you measure any speedup or better perf numbers with this? :)

stlankes:
Not really. :-)

Depending on access typo (and what your architecture supports), you can implicitly use an MSR registers.

mkroening (changes requested):
According to "25.4 GUEST-STATE AREA" in [Intel® 64 and IA-32 Architectures Software Developer's Manual Volume 3C: System Programming Guide, Part 3](https://cdrdv2.intel.com/v1/dl/getContent/671506), FS and GS are part of the processor state and not managed by KVM, so perhaps the msr exits are from some other source like explicit wrmsr and rdmsr? ("26.1 INSTRUCTIONS THAT CAUSE VM EXITS")

stlankes:
I know. But I don't think that we should use the GS register in case of a one-core system

mkroening (approved):

Issue #880, opened 2023-09-01
virtio-net-pci does not detect connection closed by other side


egrimley-arm:
Test program `Cargo.toml`:
```
[package]
name = "testtcp"
version = "0.1.0"
edition = "2021"

[dependencies]

[target.'cfg(target_os = "hermit")'.dependencies.hermit]
version = "0.6.0"
```

Test program `src/main.rs`:
```
#[cfg(target_os = "hermit")]
use hermit as _;

use std::io::Read;
use std::net::TcpListener;

fn main() {
    let listener = TcpListener::bind("127.0.0.1:8080").unwrap();
    let (mut socket, _) = listener.accept().unwrap();
    let mut buf = [0u8; 1000];
    loop {
        println!("about to read");
        match socket.read(&mut buf) {
            Err(e) => {
                println!("read err {e:?}");
                break;
            }
            Ok(n) => {
                println!("read {n}");
                if n == 0 {
                    break;
                }
            }
        }
    }
}
```

Build and run without Hermit:
```
cargo run
```

Test from another terminal (exit socat and close connection with C-c/C-d):
```
$ socat - TCP:localhost:8080
a
bb
$
```

Output:

```
about to read
read 2
about to read
read 3
about to read
read 0
$
```

Build and run with Hermit:
```
cd rusty-loader
cargo xtask build --target x86_64 --release
cd ../testtcp
cargo build --target x86_64-unknown-hermit --release \
  -Z build-std=std,core,alloc,panic_abort \
  -Z build-std-features=compiler-builtins-mem

qemu-system-x86_64 \
    -cpu qemu64,apic,fsgsbase,fxsr,rdrand,rdtscp,xsave,xsaveopt \
    -smp 1 -m 512M \
    -device isa-debug-exit,iobase=0xf4,iosize=0x04 \
    -display none -serial stdio \
    -kernel ../rusty-loader/target/x86_64/release/hermit-loader \
    -initrd target/x86_64-unknown-hermit/release/testtcp \
    -netdev user,id=u1,hostfwd=tcp::8080-:8080 \
    -device virtio-net-pci,netdev=u1,disable-legacy=on
```

Same test. Output:
```
about to read
read 2
about to read
read 3
about to read
```
(und es bleibt hängen!)

I am supposing that this is a problem in Rusty Hermit rather than in QEMU or the system I am testing this on. Any evidence for or against that supposition would be helpful.

egrimley-arm:
[...]

PR #881, opened 2023-09-02
return error also in the closing state

stlankes:
If smoltcp receives a FIN message, the socket changes the state to the `closing`, but the socket is still active. To close the socket, the kernel returns a error if the socket is in the closing state.

The PR should solve issue hermit-os/kernel#880

### DIFF
--- a/src/fd/socket/tcp.rs
+++ b/src/fd/socket/tcp.rs
@@ -72,11 +72,12 @@ impl<T> Socket<T> {
 	async fn async_read(&self, buffer: &mut [u8]) -> Result<isize, i32> {
 		future::poll_fn(|cx| {
 			self.with(|socket| match socket.state() {
+				tcp::State::Closed | tcp::State::Closing | tcp::State::CloseWait => {
+					Poll::Ready(Ok(0))
+				}
 				tcp::State::FinWait1
 				| tcp::State::FinWait2
-				| tcp::State::Closed
-				| tcp::State::Closing
-				| tcp::State::CloseWait
+				| tcp::State::Listen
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
I think we should return `Ok(0)` in both cases.

From [`Read::read`](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read):

> If n is 0, then it can indicate one of two scenarios:
>
> 1.   This reader has reached its “end of file” and will likely no longer be able to produce bytes. Note that this does not mean that the reader will always no longer be able to produce bytes. As an example, on Linux, this method will call the recv syscall for a [TcpStream](https://doc.rust-lang.org/nightly/std/net/struct.TcpStream.html), where returning zero indicates the connection was shut down correctly. While for [File](https://doc.rust-lang.org/nightly/std/fs/struct.File.html), it is possible to reach the end of file and get zero as result, but if more data is appended to the file, future calls to read will return more data.
> 2.    The buffer specified was 0 bytes in length.

From [`Write::write`](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.write):

> A return value of Ok(0) typically means that the underlying object is no longer able to accept bytes and will likely not be able to in the future as well, or that the buffer provided is empty.

Note that the example from the issue does not result in an error on Linux either.

stlankes:
You are right. I fixed if.

mkroening (changes requested):
Should we really return an error on FIN-WAIT-1 etc.? It might be better to not match against the TCP state directly and instead use `may_send` and `may_receive`. What do you think?

https://github.com/smoltcp-rs/smoltcp/blob/main/examples/httpclient.rs#L84-L119

mkroening (approved):

PR #895, opened 2023-09-05
add KVM tests with uhyve, qemu and firecracker

stlankes:
Closes https://github.com/hermit-os/kernel/pull/868


stlankes:
@mkroening Do you know, how can I determine the version of firecracker. Currently, I use the command "firecracker-v1.4.1-x86_64" to start firecracker, which is bad if the version increased.

mkroening:
> @mkroening Do you know, how can I determine the version of firecracker. Currently, I use the command "firecracker-v1.4.1-x86_64" to start firecracker, which is bad if the version increased.

I think for now this is fine. There are options such as [`gh release list`](https://cli.github.com/manual/gh_release_list), but I am not certain how to integrate them best. Maybe an explicit version is good for now?

mkroening (approved):

PR #904, opened 2023-09-09
add test for the network interface of qemu micro vm

stlankes:


stlankes:
It is already part of the current CI

PR #907, opened 2023-09-11
ci: install kernel toolchain too

mkroening:


mkroening:
This was hidden by the internal `cargo xtask` before.

PR #811, opened 2023-07-30
Extend the FUSE support

simonschoening:
We added a few new system calls (`get_errno, set_errno, mkdir, rmdir, lstat, fstat, opendir, readdir`) and also modified the `stat` system call.

PR #917, opened 2023-09-17
Extend the FUSE support

stlankes:
We added a few new system calls (mkdir, rmdir, lstat, fstat, opendir, readdir) and also modified the stat system call.

Rebase and replace #811


mkroening (approved):
Looks good to me! 👍

PR #924, opened 2023-09-20
Identity-mapped pagetable support (UEFI boot)

sarahspberrypi:
OG: 9b8db5193

### DIFF
--- a/src/arch/x86_64/kernel/mod.rs
+++ b/src/arch/x86_64/kernel/mod.rs
@@ -80,6 +80,7 @@ pub fn is_uefi() -> bool {
 	}
 }
 
+// RSDP has to be located in the bootloader for UEFI systems, so we can retrieve the address from boot info
 pub fn get_rsdp_addr() -> u64 {
 	match boot_info().platform_info {
 		PlatformInfo::Uefi { rsdp_addr } => rsdp_addr,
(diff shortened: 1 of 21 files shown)


src/arch/x86_64/kernel/mod.rs
```diff
@@ -67,12 +67,24 @@ pub fn get_limit() -> usize {
 	boot_info().hardware_info.phys_addr_range.end as usize
 }
 
+pub fn get_start() -> usize {
+	boot_info().hardware_info.phys_addr_range.start as usize
+}
+
+pub fn is_uefi() -> Result<(), ()> {
```
jounathaen, line 74: Why you don't just return a bool?

src/mm/mod.rs
```diff
@@ -87,22 +87,45 @@ pub(crate) fn init() {
...
+		{
+			panic!("Not enough memory available!");
+		}
+
+		(total_memory_size()
+			- (kernel_end_address().as_usize() - env::get_ram_address().as_usize())
+			- reserved_space)
+			.align_down(LargePageSize::SIZE as usize)
+	};
+	// if !crate::arch::x86_64::kernel::is_uefi() {
```
jounathaen, line 111: This probably needs to be removed :slightly_smiling_face:

src/arch/x86_64/mm/paging.rs
```diff
@@ -64,6 +70,16 @@ unsafe fn recursive_page_table() -> RecursivePageTable<'static> {
 	}
 }
 
+unsafe fn identity_mapped_page_table() -> OffsetPageTable<'static> {
```
jounathaen, line 73: I'd appreciate a short doc comment, as I think it is not 100% obvious what the function is doing.

PR #926, opened 2023-09-21
feat: remove NASM dependency

mkroening:
This migrates the x86-64 application processor boot code from NASM to GNU assembly language.
    
`boot.asm` is replaced by `boot.s`.
Instead of NASM directives, their GNU assembler (`as`) equivalents are used.
This assembly file is embedded in an LLVM assembly language file with only one module level (global) assembly statement.
Rust bundles `llvm-as`, which we use to translate the LLVM assembly to LLVM bitcode.
The resulting file is linked to binary using the bundled `rust-lld`.
The final binary is exactly the same as before, bit for bit.
    
This finally removes NASM as the last non-Rust toolchain dependency.

Closes https://github.com/hermit-os/kernel/issues/914


stlankes:
You want that I am not longer able to extend the code? :-)

stlankes:
Ok, you use "intel syntax"...

stlankes (approved):

PR #940, opened 2023-10-04
ci: use a matrix for test configurations

cagatay-y:
Use a test configuration matrix to increase architecture testing parity, decrease CI configuration duplication and run tests at parallel, at the cost of increased compute and storage usage.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,12 +108,58 @@ jobs:
         env:
           RUSTFLAGS:
 
-  run-x86_64:
-    name: Run Hermit for Rust (x86_64)
+  run-hermit:
+    name: Run Hermit for Rust
     runs-on: ubuntu-latest
     defaults:
       run:
         working-directory: kernel
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+        profile: [dev, release]
+        package: [rusty_demo, httpd, testudp, hello_world]
+        netdev: [none, virtio-net-pci, rtl8139]
+        exclude:
+          # microvm (Firecracker) test does not run on aarch64
+          - package: hello_world
+            arch: aarch64
+          # rtl8139 support does not build on aarch64
+          - arch: aarch64
+            netdev: rtl8139
+          # Following packages do not depend on the netdev
+          - package: rusty_demo
+            netdev: virtio-net-pci
+          - package: rusty_demo
+            netdev: rtl8139
+          - package: hello_world
+            netdev: virtio-net-pci
+          - package: hello_world
+            netdev: rtl8139
+          # Following packages require a netdev
+          - package: httpd
+            netdev: none
+          - package: testudp
+            netdev: none
+        include:
+          - arch: x86_64
+            packages: qemu-system-x86 libcap-ng-dev libseccomp-dev socat
+          - arch: aarch64
+            packages: qemu-system-aarch64
+
+          - arch: x86_64
+            package: rusty_demo
+            flags: --features pci-ids --virtiofsd
+          - arch: aarch64
+            package: rusty_demo
+            flags: --features pci-ids
+          - package: httpd
+            flags: --features ci,dhcpv4
+          - package: testudp
+            flags: --features udp,dhcpv4
+          - package: hello_world
+            flags: --no-default-features --microvm
+
     steps:
       - name: Checkout hermit-rs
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
Thanks a lot for the PR! The CI times look very promising. :)

I added some comments.

Also, I'd like the cache to be restricted to main before merging this.

Nit: I think we can remove the parentheses from most literals.

.github/workflows/ci.yml
```diff
@@ -98,12 +98,49 @@ jobs:
         env:
           RUSTFLAGS:
 
-  run-x86_64:
-    name: Run Hermit for Rust (x86_64)
+  run-hermit:
```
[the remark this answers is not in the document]
cagatay-y: I avoided using 'run' as the name as I thought it may be confusing to have a job with the same name as one of the Actions keywords, even if it can be told apart from the context.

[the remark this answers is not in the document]
mkroening: I see. Makes sense. 👍

cagatay-y:
After configuring with a single matrix, I think it may make more sense to create a basis action for the setup and use it with two separate matrices for the networked and not networked jobs.

cagatay-y:
I think the cache change works because the last cache item from this pull request on the list was added right before [the commit that disabled caching](https://github.com/hermit-os/kernel/pull/940/commits/a17a31aeb7e669f03c7669d469f971a20d453357). I normally wouldn't mix the two changes as they are not really related but I forgot that I was on the branch with the cache change.

cagatay-y:
> How about removing the netdev dimension, but explicitly adding it in the required cases? That would be inverting the description and work similar to what you are doing with flags. I am not sure, if that is possible, though.

I tried that but the configurations that are added with include

    think the cache change works

Looks good! 👍

    I normally wouldn't mix the two changes as they are not really related but I forgot that I was on the branch with the cache change.

Feel free to wildly rebase and force-push on PRs. You could also extract that into another PR, so we can merge that already. :)
In the end, you can also squash most PRs. If you need any assistance with that, I am happy to help.

PS: You can re-request a review from me in the top right, when you want me to review again.

cagatay-y:
> How about removing the `netdev` dimension, but explicitly adding it in the required cases? That would be inverting the description and work similar to what you are doing with flags. I am not sure, if that is possible, though.

I tried that but the configurations added via `include` overrode each other.

I opened [a separate PR](https://github.com/hermit-os/kernel/pull/947)  for the cache issue. After it is merged I will rebase this one on it and request a review.

.github/workflows/ci.yml
```diff
@@ -108,12 +108,58 @@ jobs:
...
+    name: Run Hermit for Rust
```
mkroening, line 112:
```suggestion
    name: Run
```

If we shorten this, the matrix instances will be easier to on the overview: https://github.com/hermit-os/kernel/actions/runs/6434190933?pr=940

mkroening (changes requested):
This looks great. Only one more nit. Thereafter, I'll merge this. 👍

cagatay-y:
> This looks great. Only one more nit. Thereafter, I'll merge this. 👍

Makes a lot of sense, I had to resize the left pane via inspect element during testing 😂

.github/workflows/ci.yml
```diff
@@ -108,12 +108,58 @@ jobs:
...
+    name: Run
     runs-on: ubuntu-latest
     defaults:
       run:
         working-directory: kernel
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+        profile: [dev, release]
+        package: [rusty_demo, httpd, testudp, hello_world]
```
mkroening, line 121:
```suggestion
        package: [rusty_demo, httpd, testudp, hello_world]
        profile: [dev, release]
```

Looking at this again, it might also make sense to transpose the matrices. What do you think?

cagatay-y: You are right. That would bring the order in which the tests appear more in line with the hand-written cases from earlier. It may also make sense to have the netdev variable above the profile variable, so that the tree goes `arch → package → netdev → profile`. Which one would you prefer?

mkroening: Yeah, I think that makes sense. 👍

mkroening (approved):

PR #941, opened 2023-10-06
fix(core_local): put scheduler into InterruptRefCell

mkroening:


mkroening:
I am not sure whether this works with the current structure, since I believe we currently switch tasks while holding a mutable reference. Without restructuring this, a RefCell would just panic there.

I did create the [interrupt-ref-cell](https://docs.rs/interrupt-ref-cell) back then as a foundation, though, so any future PR that aims to reattempt this can build on that.

Issue #899, opened 2023-09-06
Memory leak: network buffers are never deallocated


egrimley-arm:
I get this with the same `testtcp` program as in #846, but by running the release build for about 15 minutes on Intel. The error is:
```
about to accept
[0][PANIC] panicked at library/alloc/src/alloc.rs:411:13:
memory allocation of 65535 bytes failed
[0][INFO] Number of interrupts
[0][INFO] [0][FPU]: 1
[0][INFO] [0][virtio_net]: 17103
[0][INFO] [0][Timer]: 6812
[0][INFO] Shutting down system
```

To build and run:
```
cargo build --target x86_64-unknown-hermit --release \
  -Z build-std=std,core,alloc,panic_abort \
  -Z build-std-features=compiler-builtins-mem

qemu-system-x86_64 \
    -cpu qemu64,apic,fsgsbase,fxsr,rdrand,rdtscp,xsave,xsaveopt \
    -smp 1 -m 512M \
    -device isa-debug-exit,iobase=0xf4,iosize=0x04 \
    -display none -serial stdio \
    -kernel ../rusty-loader/target/x86_64/release/hermit-loader \
    -initrd target/x86_64-unknown-hermit/release/testtcp \
    -netdev user,id=u1,hostfwd=tcp::8080-:8080 \
    -device virtio-net-pci,netdev=u1,disable-legacy=on
```

To test, from a different terminal:
```
rm -f log ; date ; while socat -u TCP:localhost:8080 CREATE:out ; do date >> log ; done &
```

The error happens after exactly 3443 connections have been made, which takes about 14-15 minutes on my system. So I suspect a memory leak.

rusty-hermit is at 5051aeed6c01845962973b0e907b366ed8fc0163
rusty-hermit/kernel is at 5f1ccd640f54de829c3aac6c88c2d5b90ddd0d7e

(I've not seen this on AArch64 but quite possibly just because on AArch64 a different error happens first: #898.)

egrimley-arm:
If I replace `let n = 300` with `let n = 0` in the test program the error happens after about 1 minutes and 3600 connections.

mkroening:
I opened a PR that fixes this: https://github.com/hermit-os/kernel/pull/943

When opening and closing thousands of connections with that PR, memory usage stays flat. 🥳

PR #943, opened 2023-10-06
fix(net): deallocate network buffers

mkroening:
Previously, network buffers were never deallocated again, resulting in quick out-of-memory situations if many connections are created.

Closes https://github.com/hermit-os/kernel/issues/899


stlankes (approved):
A stupid failure from my side. :-(

PR #956, opened 2023-10-15
chore: upgrade to nightly-2023-10-15

mkroening:


mkroening:
I opened https://github.com/firecracker-microvm/firecracker/issues/4176 for the CI failure with Firecracker. Maybe we should pin Firecracker 1.4.1 for the time being?

Issue #793, opened 2023-07-14
Consider adopting `ptr_from_ref`


mkroening:
https://github.com/rust-lang/rust/issues/106116

PR #964, opened 2023-10-24
Minimize casts to pointers

cagatay-y:
Minimize casts to pointers by using:
- pointer formatting,
- from_exposed_addr{,_mut},
- from_{ref,mut} (solves #793),
- as_str
- and some other methods.

### DIFF
--- a/src/arch/aarch64/kernel/interrupts.rs
+++ b/src/arch/aarch64/kernel/interrupts.rs
@@ -190,7 +190,7 @@ pub(crate) extern "C" fn do_sync(state: &State) {
 
 			// add page fault handler
 
-			error!("Current stack pointer {:#x}", state as *const _ as u64);
+			error!("Current stack pointer {state:p}");
 			error!("Unable to handle page fault at {:#x}", far);
 			error!("Exception return address {:#x}", ELR_EL1.get());
 			error!("Thread ID register {:#x}", TPIDR_EL0.get());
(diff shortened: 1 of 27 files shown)


mkroening:
CI failure is being fixed in https://github.com/hermit-os/kernel/pull/965.

mkroening:
Thanks a lot! :)

Could you rebase on main instead of merging main? I'd like to avoid merge commits in feature branches.

Also, it would be great if you could have the CI run all of these commits. I usually do that by force pushing the first commit of the feature branch and then pushing commit by commit. Something like `git push origin HEAD~10:cast-elimination --force` should work (`--force` can be omitted on all but the first push).

I can also recomment `git-interactive-rebase-tool`, but any editor of your choice works just as well for rebases.

mkroening:
Awesome. You also don't need to be up to date with main and rebase on every change, but having a green CI everywhere would be great. :)

cagatay-y:
> As a side note, you don't have to wait for CI to finish before pushing the next commit. :)

I completely misread this message the first time and thought that you said I _have to_ wait :) Thanks for the tip!

mkroening:
I did cancel the CI runs starting from the first unsuccessful one, so this does not block the CI. I think, you should be able to do that too if rights are configured correctly: https://github.com/hermit-os/kernel/actions

mkroening (approved):
Perfect! 👌 

Thanks a lot! :)

PR #968, opened 2023-10-29
collect configuration of mmio devices

stlankes:
Firecracker pass the configuration of all mmio device by a Linux kernel parameter This parameter is defined as followd:

> virtio_mmio.device=
>    [VMMIO] Memory mapped virtio (platform) device.
> 
>       <size>@<baseaddr>:<irq>[:<id>]
>    where:
>       <size>     := size (can use standard suffixes like K, M and G)
>       <baseaddr> := physical base address
>       <irq>      := interrupt number (as passed to request_irq())
>       <id>       := (optional) platform device id
>    example:
>                  virtio_mmio.device=1K@0x100b0000:48:7
>    Can be used multiple times for multiple devices.

This patch parse the command line and store the string `<size>@<baseaddr>:<irq>[:<id>]` in a vector, which is part of CLI. The vector is used to find the network device on a microVM.

In addition, the network driver support devices, which doesn't support VIRTIO_NET_F_STATUS.

This PR based on preliminary work from @duanyu-yu

### DIFF
--- a/src/arch/x86_64/kernel/mmio.rs
+++ b/src/arch/x86_64/kernel/mmio.rs
@@ -1,3 +1,4 @@
+use alloc::string::String;
 use alloc::vec::Vec;
 use core::{ptr, str};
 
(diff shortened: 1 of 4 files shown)


mkroening (approved):
Looks good to me! 👍 

Waiting on the CI on each commit before merging.

PR #972, opened 2023-11-02
replace self-baked Pinned with std::pin

### LINKED ISSUE #668: Replace self-baked Pinned with std::pin
https://github.com/hermitcore/libhermit-rs/blob/8802a3fc0a55c12d668618f4cd641421fa9d4228/src/drivers/virtio/virtqueue/mod.rs#L2374-L2389

[std::pin](https://doc.rust-lang.org/stable/std/pin/index.html)

cagatay-y:
Resolves #668.


mkroening:
Oh no, this feels like miscompilations due to too strong aliasing guarantees.
I guess, this would already be the case for unpinned boxes.
We'll have to make sure that all reads and writes are volatile.

Looking into the generated assembly using `rust-objdump -Cdl <HERMIT_BINARY>` might provide some insights.

PR #983, opened 2023-11-23
Remove Pinned

cagatay-y:
Pinned was used to keep track of if a raw pointer to the object was handed out via the use of a boolean and free or not free accordingly. The initial commit replaces it with Rc to keep track of the sharing in a standard and safer way. The second commit eliminates the need to keep track of shares completely by removing the sharing of `TransferToken`s and adapting the only instance of the use of the returned `TransferToken`.


cagatay-y:
I think it is possible that I introduced a memory leak. Can we profile the memory consumption of `main` and this branch to check?

mkroening:
We'll sure, you could accumulate the sizes of all allocs minus all deallocs somewhere and compare.
I did something similar for debugging purposes, hacked into the allocator once using an `AtomicUsize` for https://github.com/hermit-os/kernel/issues/899.
You could also start VMs with minimal memory and create/destroy connections a lot and see if we run out of memory.

cagatay-y:
I've tested the branch with `siege` using 10 concurrent connections for 115 seconds (just under the xtask timeout). It handled 34460 requests in total without error. The QEMU process also seems to have a stable memory usage during the test and returns the memory at the end of the test.

mkroening:
Thanks for testing! How much memory did the VM get?

cagatay-y:
I am not on my computer right now, so I cannot check to be sure but if I recall correctly, ~95 MB during the test and ~80 MB at idle. The VM limit was the xtask default (128 MB (?)).

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -704,13 +670,12 @@ impl Transfer {
...
-						.unwrap()
-						.buff_tkn
-						.as_mut()
-						.unwrap();
+					} = unsafe {
+						Rc::get_mut_unchecked(self.transfer_tkn.as_mut().unwrap())
+							.buff_tkn
+							.as_mut()
+							.unwrap()
+					};
```
mkroening, line 678: `Rc::get_mut` does not work here, right? We can't justify this unsafe, can we? We should mark this as `FIXME`, stating that this should be reworked in the future. It is a step in the right direction, having these invariants spread around instead of collected in `Pinned` and collectively dismissed.

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -135,47 +135,38 @@ impl DescriptorRing {
...
 			// all operations on the reference.
-			let tkn = unsafe { &mut *(tkn) };
-
-			match tkn.await_queue {
-				Some(_) => {
-					tkn.state = TransferState::Finished;
-					let queue = tkn.await_queue.take().unwrap();
 
+			unsafe {
+				let tkn_ref = Rc::get_mut_unchecked(&mut tkn);
```
mkroening, line 146: Same as above. Also, we should keep the `unsafe` block as small as possible.

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -472,26 +467,22 @@ struct ReadCtrl<'a> {
...
+				.take()
+				.expect("TransferToken at position does not exist");
 
 			let (send_buff, recv_buff) = {
 				let BufferToken {
 					send_buff,
 					recv_buff,
 					..
-				} = tkn.buff_tkn.as_mut().unwrap();
+				} = unsafe { Rc::get_mut_unchecked(&mut tkn).buff_tkn.as_mut().unwrap() };
```
mkroening, line 485: Same as above.

src/drivers/virtio/virtqueue/split.rs
```diff
@@ -189,41 +189,48 @@ impl DescrRing {
...
-			let tkn = unsafe { &mut *(self.ref_ring[used_elem.id as usize]) };
+			let mut tkn = self.ref_ring[used_elem.id as usize]
+				.take()
+				.expect("Invalid TransferToken reference in reference ring");
 
-			if tkn.buff_tkn.as_ref().unwrap().recv_buff.as_ref().is_some() {
-				tkn.buff_tkn
-					.as_mut()
+			unsafe {
+				let tkn_ref = Rc::get_mut_unchecked(&mut tkn);
```
mkroening, line 212: Same as above. Also, we should keep the `unsafe` block as small as possible.

mkroening (changes requested):
Thanks, this looks great! :)

Sorry for taking so long with the review.

Just a few remarks regarding the introduction of additional `unsafe` blocks.

src/drivers/virtio/virtqueue/split.rs
mkroening:
```rust
let tkn_ref = unsafe { Rc::get_mut_unchecked(&mut tkn) };
```

cagatay-y:
Looking closer at the uses of unsafe, I think they are fundamentally not thread-safe. It is possible to use synchronization to make it safe, but I realized that the only uses of shared Transfers are inside the module itself to check for Transfer status and they can be avoided. The latest commit does that. 

If it is preferable to keep the possibility to check status through the Transfer for future use, we can also switch to using synchronization. I kept the initial commit (the one with reference counting) to make it more convenient to review the changes to the original PR and to switch back to the Rc approach if necessary. The two commits can be squashed if the approach described above is acceptable.

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -82,13 +83,9 @@ impl WrapCount {
...
 
 /// Structure which allows to control raw ring and operate easily on it
-///
-/// WARN: NEVER PUSH TO THE RING AFTER DESCRIPTORRING HAS BEEN INITIALIZED AS THIS WILL PROBABLY RESULT IN A
-/// RELOCATION OF THE VECTOR AND HENCE THE DEVICE WILL NO LONGER NO THE RINGS ADDRESS!
 struct DescriptorRing {
 	ring: &'static mut [Descriptor],
-	//ring: Pinned<Vec<Descriptor>>,
-	tkn_ref_ring: Box<[*mut TransferToken]>,
+	tkn_ref_ring: Box<[MaybeUninit<Box<TransferToken>>]>,
```
mkroening, line 88: Would it make sense to replace `MaybeUninit<Box<T>>` with `Option<Box<T>>`?
It is the same size due to niches.
It would incur some overhead due to checking when accessing, but properly ensuring the soundness of accessing `MaybeUninit` might be too difficult, no?

mkroening (commented):
Thanks a lot! This looks great.

Just one question about `MaybeUninit` inline.

Another question: Is the PR description still up to date?

You mention one-bit refcounts. What do you mean with that? I mean a refcounter that can only count to one is exactly the same as just having a single owner.

The description also talks about replacing `Pinned<T>` with `Rc<T>`, but to me, the code reads more like replacing it with `Box<T>` and `T`, which is good, I was just wondering about the description. :)

cagatay-y:
I've updated the PR description and implemented the change from MaybeUninit to Option.

mkroening (approved):
Thanks a lot! :)

PR #999, opened 2023-12-05
feat: add `sys_image_start_address`

mkroening:


stlankes:
Should we add feature to enable these functions. Normally, the memory layout should be randomized and these function shouldn't exists.

mkroening:
Maybe? In that case, we would want to add a function to get the randomized offset of individual symbols to still be able to debug, though, right? As far as I understand it, it's not a secret, it's just random.

PR #1013, opened 2023-12-25
add basic support of a virtual filesystem

stlankes:
Add basic support of a virtual filesystem (including an in-memory fs). If the kernel is running on top of uhyve, the host system will be mounted at mount point, which is defined by the kernel argument `-mount`. If this argument is not used, the kernel uses `/host` as default mount point. If the kernel detects FUSE, the kernel mount the filesystem at the mount point, which is specified by FUSE's `tag`. In case of Qemu, the `tag` is specified by the argument `-device ` (e.g. `-device vhost-user-fs-pci,queue-size=1024,chardev=char0,tag=root`). In this example, FUSE will be mount at `/root`.

In addition, the interface for file descriptors is revised and move the usage of raw pointers to the system call implementations.


mkroening (approved):
Looks good to me! 👍

Issue #743, opened 2023-05-08
cargo-careful support


mkroening:
[cargo-careful](https://github.com/RalfJung/cargo-careful)

PR #1045, opened 2024-01-31
perf: forward default allocator methods and remove memory overprovisioning

SFBdragon:
Thought I'd check up on how people were using talc in the wild :)

There are two ideas here:
1. Use `talc`'s reallocation mechanisms. This is recommended for faster in-place reallocation (avoiding the much-more-expensive memcpy that default-reallocation _always_ invokes) and also allows for interleaving some allocator operations as introduced in v4.2 to somewhat improve allocation concurrency.
2. Avoid padding the size of allocations. Increasing the alignment of an allocation is sufficient to avoid false sharing. Talc needs a little bit of metadata next to each allocation, and this would create `N-8` sized holes in between every allocation due to the demands placed on it. By removing the size requirement, false sharing avoidance is unhindered, but this allows Talc to minimize the overhead of high-alignments. (Talc is designed to automatically place the metadata in a way that minimized false sharing, but this only helps if Talc can sneak it in within the same cache line, which is impossible if Talc thinks you want the whole cache line for user data.)

Word salad aside, the changes are quite trivial.

Closes https://github.com/hermit-os/kernel/issues/1059

### DIFF
--- a/src/mm/allocator.rs
+++ b/src/mm/allocator.rs
@@ -45,6 +45,11 @@ unsafe impl GlobalAlloc for LockedAllocator {
 		let layout = Self::align_layout(layout);
 		unsafe { self.0.dealloc(ptr, layout) }
 	}
+
+	unsafe fn alloc_zeroed(&self, layout: Layout) -> *mut u8 {
+		let layout = Self::align_layout(layout);
+		unsafe { self.0.alloc_zeroed(layout) }
+	}
 }
 
 #[cfg(all(test, not(target_os = "none")))]
--- a/src/mm/allocator.rs
+++ b/src/mm/allocator.rs
@@ -50,6 +50,12 @@ unsafe impl GlobalAlloc for LockedAllocator {
 		let layout = Self::align_layout(layout);
 		unsafe { self.0.alloc_zeroed(layout) }
 	}
+
+	unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
+		let layout = Self::align_layout(layout);
+		let new_size = new_size.align_up(core::mem::size_of::<crossbeam_utils::CachePadded<u8>>());
+		unsafe { self.0.realloc(ptr, layout, new_size) }
+	}
 }
 
 #[cfg(all(test, not(target_os = "none")))]
--- a/src/mm/allocator.rs
+++ b/src/mm/allocator.rs
@@ -3,7 +3,6 @@
 
 use core::alloc::{GlobalAlloc, Layout};
 
-use align_address::Align;
 use hermit_sync::RawInterruptTicketMutex;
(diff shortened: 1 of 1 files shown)


SFBdragon:
Hmm..
```
thread '<unnamed>' panicked at /home/runner/.rustup/toolchains/nightly-2024-01-15-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/local.rs:262:26:
cannot access a Thread Local Storage value during or after destruction: AccessError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Just to confirm, this was introduced by this PR? 

I'm going to submit a PR with only the size round-up, no reallocation invocation, to A/B test this.

SFBdragon:
Hmm. My only guess at the moment is there's a memory management bug in some platform-specific codepath in the kernel, or the size of some structure is meaningfully different? (That is avoided by allocating/deallocating more than needed? Or perhaps the rounding might be homogenizing the size of a deallocation of an allocation with a different size sizes which... is related how? I don't know.) I don't have aarch64 hardware nor the familiarity with the codebase to meaningfully assist there though.

I can change the PR to just enable reallocation, and a separate issue could be opened to address this, or just keep the PR blocked like this? Completely up to you.

mkroening:
Thanks for opening this! Both changes are wonderful. I opened https://github.com/hermit-os/kernel/issues/1059 to track avoiding padding the size of allocations. I'll have to dig a bit into this, to locate the underlying issue.

Changing this PR to resolve the reallocation issue would be great. Then we can move forward with that change while I dig into the other issue.

mkroening (approved):

SFBdragon:
I'm causing a mess, heh, sorry. I'll need to check this out properly when I have time. Hopefully soon.

mkroening:
Interesting issues! Thanks a lot for your help! :)

I'll look at https://github.com/hermit-os/kernel/issues/743. Maybe that will help us locate any issues on our side.

mkroening (approved):
Hi, @SFBdragon, as you have already noticed, I finally located the issue that was causing the AArch64 CI to fail (https://github.com/hermit-os/kernel/pull/1155). We had allocated too little memory for a struct, which then became corrupted by the allocator metadata.

I am relieved, that we can merge this now. Hopefully, we will find these cases as soon as we introduce them from now on.

Thanks a lot for your contribution! :)

SFBdragon:
Well spotted! This class of bug is a nightmare..

I've been thinking about the feasibility of talc providing a configuration that performs more thorough integrity checks, with extra memory overhead for debug data to help with catching issues like overwriting allocator metadata. I can't promise anything in the near future but I think it would be quite useful to embedded projects where unsafe code is common. (So you'd run your tests and CI in this configuration but leave it off in production, typically. Although I'd rather not rely on a system allocator necessarily.) Idle thoughts for now though.

PR #1047, opened 2024-02-03
Refactor FUSE

cagatay-y:
This PR aims to use safer standard library functionality for allocation for FUSE, deduplicate what is in common between different operations and specify what is necessarily varying between them in a easier-to-read manner. 
Another goal is to set up the module so that in the future we can access the `Rsp` layout without creating a skeleton `Rsp` object, in order to eliminate copying. However, that requires changes in virtqueue API and is not a part of this PR.


cagatay-y:
Some of the commits that were automatically rebased do not build. I thought it may make more sense to get feedback first and fix the commits later, incorporating the feedback. I also could not figure out why Clippy thinks that ROOT_ID is not used.

src/fs/mod.rs
```diff
@@ -1,5 +1,6 @@
 #[cfg(all(feature = "fuse", feature = "pci"))]
 pub(crate) mod fuse;
+mod fuse_abi;
```
mkroening, line 3: The issue with Clippy is that with `--no-default-features`, the constant is unused.

Would something like this work?

```suggestion
#[cfg(all(feature = "fuse", feature = "pci"))]
mod fuse_abi;
```

mkroening (changes requested):

src/lib.rs
```diff
@@ -15,13 +15,15 @@
...
 #![feature(asm_const)]
+#![feature(associated_type_defaults)]
 #![feature(exposed_provenance)]
 #![feature(linked_list_cursors)]
 #![feature(maybe_uninit_slice)]
 #![feature(naked_functions)]
 #![feature(new_uninit)]
 #![feature(noop_waker)]
 #![feature(pointer_is_aligned)]
+#![feature(ptr_metadata)]
```
mkroening, line 26: Can we avoid depending on these unstable features? We would like to decrease the dependence on unstable features if possible.

src/fs/fuse.rs
```diff
@@ -38,1190 +38,557 @@ const U64_SIZE: usize = ::core::mem::size_of::<u64>();
...
+		let cmd = Self::new_cmd(
+			fuse_abi::ROOT_ID,
+			fuse_abi::InitIn {
+				major: 7,
+				minor: 31,
+				max_readahead: 0,
+				flags: 0,
+			},
+		);
+		let rsp = unsafe { Box::new_uninit().assume_init() };
```
mkroening, line 117: This is immediate undefined behavior. You need to either initialize `rsp` here, or return `MaybeUninit`.

This applies to all other instances of `new_uninit().assume_init()` too (except for producing, say `[MaybeUninit<u8>; N]` or other uninitialized structs). See [`std::mem::uninitialized`](https://doc.rust-lang.org/nightly/std/mem/fn.uninitialized.html) for details on why this is bad.

src/fs/fuse.rs
```diff
@@ -38,1190 +38,557 @@ const U64_SIZE: usize = ::core::mem::size_of::<u64>();
...
+	type OutStruct: FuseOut + core::fmt::Debug;
+	type OutPayload: ?Sized;
+
+	type Cmd: ?Sized + AsSliceU8 = ReqPart<fuse_abi::InHeader, Self::InStruct, Self::InPayload>;
+	type Rsp: ?Sized + AsSliceU8 =
+		ReqPart<fuse_abi::OutHeader, MaybeUninit<Self::OutStruct>, Self::OutPayload>;
 }
 
-unsafe impl FuseIn for fuse_init_in {}
+pub(crate) struct Op<const CODE: u32>;
```
mkroening, line 97: Is there any benefit of having this be a const generic instead of having distinct types that implement `OpTrait`? I found that `Op<{ fuse_abi::Opcode::Init as u32 }>` reads much more difficult than, just `ops::Init`. Similar for where clauses where `cmd: &T::Cmd where T: OpTrait` would be much easier to read than `cmd: &<fuse::Op<CODE> as fuse::OpTrait>::Cmd where fuse::Op<CODE>: fuse::OpTrait`.

src/fs/fuse.rs
```diff
@@ -38,1190 +38,557 @@ const U64_SIZE: usize = ::core::mem::size_of::<u64>();
...
-	pub len: u32,
-	pub error: i32,
-	pub unique: u64,
-}
+/// Marker trait, which signals that a struct is a valid Fuse command.
+/// Struct has to be repr(C)!
+pub(crate) unsafe trait FuseIn {}
+/// Marker trait, which signals that a struct is a valid Fuse response.
+/// Struct has to be repr(C)!
+pub(crate) unsafe trait FuseOut {}
```
mkroening, line 57: We'd need a safety comment on these, explaining exactly what is structurally depended upon. Is there any overlap with [`zerocopy::FromBytes`](https://docs.rs/zerocopy/latest/zerocopy/trait.FromBytes.html)?

src/drivers/fs/virtio_fs.rs
```diff
@@ -153,17 +153,21 @@ impl VirtioFsDriver {
...
 	{
 		if let Some(mut buff_tkn) = self.ready_queue.pop() {
 			let cmd_len = Some(cmd.len());
 			let rsp_len = Some(rsp.len());
 			buff_tkn.restr_size(cmd_len, rsp_len).unwrap();
 
-			let transfer_tkn = buff_tkn.write(Some(cmd), Some(rsp)).unwrap();
+			let transfer_tkn = buff_tkn
+				.write(Some(cmd), None::<&<fuse::Op<CODE> as fuse::OpTrait>::Rsp>)
+				.unwrap();
```
mkroening, line 170: Did you change `Some(rsp)` to `None` purposefully?

mkroening (changes requested):
Thanks a lot, @cagatay-y!

This is looking great, though I have a few requests. Please don't hesitate to reach out, if you want to talk about them.

src/fs/fuse.rs
cagatay-y: You are right. I started writing the code thinking that C-style enums can be const generic parameters and once it didn't work as I hoped I worked around the limitation instead of changing it to empty structs.

src/fs/fuse.rs
cagatay-y: I think we can also simply remove them. They were there in the previous version of the code and I kept them since I thought they did not hurt but I don't think they are all that useful anymore since we clarify the input/output relationship with the associated types.

src/drivers/fs/virtio_fs.rs
cagatay-y: Yes. As far as I understand, FUSE does not need the memory allocated for the response to be initialized at all. It reads the size of the writable section from the command struct. `BufferToken::write` copies the the argument when it is Some and since we don't initialize the response, I bypassed the copying.

src/fs/fuse.rs
cagatay-y: I tried using MaybeUninit but it did not work because `Rsp` is a dynamically-sized type (at least in the cases where the output payload is not `()`). What would be an appropriate alternative here?

mkroening: For `OutPayload = ()`, we could `impl zerocopy::FromZeroes for Op` and use `FromZeros::new_box_zeroed()`. For `OutPayload = [MaybeUninit<u8>]` we have to at least initialize the non-`MaybeUninit` part.

BTW: In the future, [`DynSized`](https://github.com/rust-lang/rfcs/pull/3536) hopefully improves on all of this. :D

src/fs/fuse.rs
```diff
@@ -38,1190 +38,557 @@ const U64_SIZE: usize = ::core::mem::size_of::<u64>();
...
-struct fuse_read_in {
-	pub fh: u64,
-	pub offset: u64,
-	pub size: u32,
-	pub read_flags: u32,
-	pub lock_owner: u64,
-	pub flags: u32,
-	pub padding: u32,
+impl Op<{ fuse_abi::Opcode::Init as u32 }> {
+	fn create() -> (Box<<Self as OpTrait>::Cmd>, Box<<Self as OpTrait>::Rsp>) {
```
mkroening, line 107: For `Sized` return types, we might want to avoid boxing if possible. Though that might be better off as a separate PR, since boxing was the previous behavior too, I think.

src/fs/fuse.rs
cagatay-y: Would it be safe to `assume_init` the struct if the fields inside the struct are MaybeUninit? I would assume it would be since it is not possible to access the fields, but I don't know for sure.

Otherwise, I think init the struct with fields set to `MaybeUninit::uninit()`s, but I think the first option is preferable since it saves us from writing a bunch of struct literals with `MaybeUninit::uninit()`s.

mkroening: > Would it be safe to `assume_init` the struct if the fields inside the struct are MaybeUninit?

Yeah, that would correspond to the `[MaybeUninit<u8>; N]` case above.

src/fs/fuse.rs
cagatay-y: By the way, I removed the traits in 62f08f61548a6aa7216324b871f1ee67ffbfd1e1.

src/fs/fuse.rs
cagatay-y: The new definition for Rsp declares each field as MaybeUninit (or (), in the case of payloadless Rsps).

src/fs/fuse.rs
cagatay-y: Solved with 695a46766e5e2947b3b71af8d65f6740035e27ee.

src/drivers/fs/virtio_fs.rs
cagatay-y: The part of the request structure that corresponds to our `Rsp` is [referred](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-4670001) to as a “[d]evice-writable part” in the file system device section of the Virtio specification and the Virtqueue section [specifies](https://docs.oasis-open.org/virtio/virtio/v1.2/virtio-v1.2.html#x1-440001) that “a device SHOULD NOT read a device-writable buffer (it MAY do so for debugging or diagnostic purposes).”

cagatay-y: Since this is true for Virtqueue's in general, I think we can also remove receive buffer copying from the Virtqueue interface (unless we want to keep it for “debugging or diagnostic purposes,” as mentioned in the specification), though it should probably be a separate PR since it touches parts of the code base outside FUSE.

mkroening: All right, sounds good! 👍

mkroening (approved):
Thanks a lot, this looks great! Could you rebase so CI passes? :)

Issue #418, opened 2022-03-27
Rework integration tests


mkroening:
- [ ] Only use public functions.
- [ ] Consume the kernel as static library.
- [ ] Integrate the runtime into xtask.

PR #1052, opened 2024-02-08
add basic support a linux-like syscall `evendfd`

stlankes:
`eventfd` creates an linux-like "eventfd object" that can be used as an event wait/notify mechanism by user-space applications, and by the kernel to notify user-space applications of events. The object contains an unsigned 64-bit integer counter that is maintained by the kernel. This counter is initialized with the value specified in the argument `initval`.

As its return value, `eventfd` returns a new file descriptor that can be used to refer to the eventfd object.

The following values may be bitwise set in flags to change the behavior of `eventfd`:
- `EFD_NONBLOCK`: Set the file descriptor in non-blocking mode
- `EFD_SEMAPHORE`: Provide semaphore-like semantics for reads from the new file descriptor.

In addition, the PR includes some fine-tuning of the IP stack and the behavior of the system call `poll`. This should increase the compatibility to Unix.

### DIFF
--- a/src/fd/eventfd.rs
+++ b/src/fd/eventfd.rs
@@ -12,14 +12,16 @@ use crate::fd::{block_on, EventFlags, IoError, ObjectInterface, PollEvent};
 #[derive(Debug)]
 struct EventState {
 	pub counter: u64,
-	pub queue: VecDeque<Waker>,
+	pub read_queue: VecDeque<Waker>,
+	pub write_queue: VecDeque<Waker>,
 }
 
 impl EventState {
 	pub fn new(counter: u64) -> Self {
 		Self {
 			counter,
-			queue: VecDeque::new(),
+			read_queue: VecDeque::new(),
+			write_queue: VecDeque::new(),
 		}
 	}
 }
@@ -67,9 +69,12 @@ impl ObjectInterface for EventFd {
 						guard.counter -= 1;
 						let tmp = u64::to_ne_bytes(1);
 						buf[..len].copy_from_slice(&tmp);
+						if let Some(cx) = guard.write_queue.pop_front() {
+							cx.wake_by_ref();
+						}
 						Poll::Ready(Ok(len))
 					} else {
-						guard.queue.push_back(cx.waker().clone());
+						guard.read_queue.push_back(cx.waker().clone());
 						Poll::Pending
 					}
 				} else {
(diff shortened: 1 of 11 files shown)


src/syscalls/net.rs
```diff
@@ -265,22 +267,30 @@ extern "C" fn __sys_socket(domain: i32, type_: i32, protocol: i32) -> i32 {
 			let fd = FD_COUNTER.fetch_add(1, Ordering::SeqCst);
 
 			#[cfg(feature = "udp")]
-			if type_ == SOCK_DGRAM {
+			if type_ & SOCK_DGRAM == SOCK_DGRAM {
```
jounathaen, line 270: Not super relevant for now, but I think it'd be nicer if `type_` was a [_bitflag_](https://docs.rs/bitflags/latest/bitflags/index.html) or something similar.

jounathaen:
Can we add some kind of test for this functionality?

mkroening:
> Can we add some kind of test for this functionality?

There will be one on the user space side of things.

Revamping integration tests in this repo is another thing that I should move up on my list. Then, we could test those things easily here as well: https://github.com/hermit-os/kernel/issues/418

src/syscalls/net.rs
stlankes: This is the C user space interface. In "high-level" representation I use normally bitflags

stlankes: I revised the code. I use now a bitflag type.

src/fd/eventfd.rs
```diff
@@ -0,0 +1,202 @@
...
+							cx.wake_by_ref();
+						}
+						Poll::Ready(Ok(len))
+					} else {
+						guard.read_queue.push_back(cx.waker().clone());
+						Poll::Pending
+					}
+				} else {
+					Poll::Pending
+				}
```
mkroening, line 82: We can extract `Poll::Ready` with [`core::task::ready`](https://doc.rust-lang.org/stable/core/task/macro.ready.html):

```suggestion
				let mut guard = ready!(pinned.as_mut().poll(cx));
				if guard.counter > 0 {
					guard.counter -= 1;
					let tmp = u64::to_ne_bytes(1);
					buf[..len].copy_from_slice(&tmp);
					if let Some(cx) = guard.write_queue.pop_front() {
						cx.wake_by_ref();
					}
					Poll::Ready(Ok(len))
				} else {
					guard.read_queue.push_back(cx.waker().clone());
					Poll::Pending
				}
```

If you like, I can do this after merging this PR or change this PR, however you prefer.

src/fd/eventfd.rs
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	async fn poll(&self, event: PollEvent) -> Result<PollEvent, IoError> {
+		let mut result: PollEvent = PollEvent::empty();
+		let guard = self.state.lock().await;
+
+		if guard.counter < u64::MAX - 1 {
+			if event.contains(PollEvent::POLLOUT) {
+				result.insert(PollEvent::POLLOUT);
+			}
```
mkroening, line 151: This could be heavily deduplicated by using a closure. Again, I can do this after merging this PR or before.

mkroening (approved):
Looks good to me!

I have two comments that I will address after merging.
I also opened https://github.com/hermit-os/kernel/issues/1060 since we don't need `async` locks in numerous instances.

stlankes:
I merge this PR. Just to release a new abi and to publish some ci test.

Issue #989, opened 2023-11-30
virtio: adapt `enum_dispatch`


mkroening:
https://docs.rs/enum_dispatch

cagatay-y:
I think we can close this issue since using a trait and trait objects worked out for `SplitVq` and `PackedVq`s.

mkroening:
> I think we can close this issue since using a trait and trait objects worked out for `SplitVq` and `PackedVq`s.

This issue is about potential performance improvements, not about whether we need it to work at all. I have reworked the description.

This issue might be interesting for @CarlWachter, who will be working on benchmarking.

CC: @jounathaen

PR #1053, opened 2024-02-08
Use a trait for the common Virtq interface

cagatay-y:
The initial version simply comments out a method that would be more difficult to adapt, as I wanted to see if the rest works on CI before dealing with it. Should also solve #989 (by removing the need for it).


cagatay-y:
a0c1b5a1e125954956135a297a3202ee8995ee35 is a bit outside the scope of the PR but I pushed it into this branch rather than making it a separate PR because it depends on the older commits. If preferred, we can merge only the commits before it and I can open it as a separate PR.

cagatay-y:
FYI: `git diff --color-moved` provides much better output for the last commit.

mkroening (approved):
Thanks a lot! :)

PR #1061, opened 2024-02-13
add CI tests for the MIO support

stlankes:


xtask/src/ci/qemu.rs
```diff
@@ -279,6 +283,36 @@ fn test_testudp() -> Result<()> {
...
+	stream.write_all(buf.as_bytes())?;
+
+	let mut buf = vec![];
+	let received = stream.read_to_end(&mut buf)?;
+	eprintln!("[CI] receive: {}", from_utf8(&buf[..received])?);
+
+	Ok(())
+}
+
+fn test_mioudp() -> Result<()> {
```
mkroening, line 308: You added `socket.recv` to `test_testudp` but not to `test_mioudp`. Is that on purpose?

stlankes: I forgot it and fixed it.

mkroening (approved):

PR #1074, opened 2024-02-19
Detect memory from fdt

duanyu-yu:
Add option to detect memory regions from fdt


mkroening:
CC: @sarahspberrypi

src/arch/x86_64/kernel/mod.rs
```diff
@@ -82,6 +82,10 @@ pub fn get_mbinfo() -> VirtAddr {
 	}
 }
 
+pub fn get_fdt() -> usize {
+	boot_info().hardware_info.device_tree.unwrap().get() as usize
+}
```
mkroening, line 87: This needs to return an `Option`, since this is not available on Uhyve (hence the CI failure).

src/arch/x86_64/mm/physicalmem.rs
```diff
@@ -19,6 +19,47 @@ const KVM_32BIT_MAX_MEM_SIZE: usize = 1 << 32;
 const KVM_32BIT_GAP_SIZE: usize = 768 << 20;
 const KVM_32BIT_GAP_START: usize = KVM_32BIT_MAX_MEM_SIZE - KVM_32BIT_GAP_SIZE;
 
+fn detect_from_fdt() -> Result<(), ()> {
```
mkroening, line 22: Since this is fallible (returns `Result`), this should also be used.

Ideally, things that can reasonably fail should return an error instead of panicking. Then, the next method for detection is used.
For example, we can use the `Option` from `get_fdt` with
```rust
let fdt = get_fdt().ok_or(())?;
```

mkroening (changes requested):
Thanks! :)

mkroening (approved):
Thanks! :)

PR #1082, opened 2024-02-28
add error number in thread local storage

stlankes:
Keep a copy of the the error number in TLS


stlankes:
Thanks @mkroening for your changes. I will merge it.

PR #1084, opened 2024-02-29
fuse: reduce copying and handle errors

cagatay-y:
Instead of copying the commands into Virtqueue buffers and the buffers back into responses, directly provide the commands and responses as references to the send_command method.


stlankes (approved):

mkroening (approved):
Thanks! :)

PR #1096, opened 2024-03-13
add option to start a simple shell as async task

stlankes:
Currently, the shell is only able to print the number of received interrupts. Type `h` to show the supported commands.


mkroening (approved):
Looks good to me! :)

PR #1099, opened 2024-03-16
RFC: add async interrupt safe synchronization primitives

stlankes:
An asynchronous interface has been added to the
interrupt-safe mutex so that interrupt handlers and asynchronous tasks can be synchronized. The mutex was demonstrated by using the shell as an example, where the asynchronous task communicates with the serial interface and its interrupt handler.


mkroening (changes requested):
Do you still want to pursue this?

PR #1105, opened 2024-03-22
feat: implement system calls with attribute proc macro

mkroening:
The purpose of this PR is the last commit: Implementing system calls via an attribute proc macro to reduce the boilerplate code from having both `sys_` and `__sys_` functions everywhere.

Along the way, I found and resolved many issues and inconsistencies with the system calls.

This is best reviewed commit by commit.


stlankes (approved):
Cool!

Issue #1110, opened 2024-03-26
Panic in FUSE


mkroening:
This program

```rust
fn main() {
    std::fs::write("/root/test.txt", "important content").unwrap();
}
```

has an out-of-bounds panic since https://github.com/hermit-os/kernel/pull/1047:

```console
[0][PANIC] panicked at src/fs/fuse.rs:528:50:
range end index 10 out of range for slice of length 0
```

Bisecting gives 93fc6cbb2ab2432b3d895a28fd1507226823734c as reason:

```git
93fc6cbb2ab2432b3d895a28fd1507226823734c is the first bad commit
commit 93fc6cbb2ab2432b3d895a28fd1507226823734c
Author: Çağatay Yiğit Şahin <cyigitsahin@outlook.com>
Date:   Wed Feb 14 15:54:59 2024 +0100

    fuse: remove dependency on unstable ptr_metadata

 src/fs/fuse.rs | 28 +++++++++++-----------------
 src/lib.rs     |  1 -
 2 files changed, 11 insertions(+), 18 deletions(-)
```

CC: @cagatay-y 

PS: I should really fix https://github.com/hermit-os/kernel/issues/620.

cagatay-y:
I am confused about how this works. When I extract the expression with `slice_from_raw_parts_mut` into a variable, rust-analyzer reports the type as `*mut [u8]`. `len` is not the length of the entire struct, rather the payload, so the slice we create here does not cover the whole struct. I validated this by checking the length of the pointed slice.

The reason I originally used 0 as the slice length was because that was how it was in the previous version of the code, but I think reverting 93fc6cb and using ptr_metadata here may be worth the unstable feature. I included a commit with the reversion in my PR (#1113) for #1111, but I can remove it if you prefer.

PR #1112, opened 2024-03-26
fix(fuse): fix fat pointer metadata

mkroening:
Closes https://github.com/hermit-os/kernel/issues/1110.


mkroening:
CC: @cagatay-y

PR #1113, opened 2024-03-26
fuse: fix incorrect len field in command headers

### LINKED ISSUE #1111: Extra zeros at the end of files with FUSE
This program

```rust
fn main() {
    std::fs::write("/root/test.txt", "important content").unwrap();
}
```

writes extra zeros to the end of the file:

```console
$ od -t x1 foo/test.txt 
0000000 69 6d 70 6f 72 74 61 6e 74 20 63 6f 6e 74 65 6e
0000020 74 00 00 00 00 00 00 00
0000030
```

expected:

```console
$ od -t x1 foo/test.txt 
0000000 69 6d 70 6f 72 74 61 6e 74 20 63 6f 6e 74 65 6e
0000020 74
0000021
```

Bisecting fails since not all of https://github.com/hermit-os/kernel/pull/1047 's commits compile!

```
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
a2212c010250c69ce581a9459683ae14a7b963ef
fbb5789508286aa5a019f8b9e5371c3ecbf108d2
24c425b5e825ca5e2c91bdd863df9b5cdf33717f
3a93d4a843fa9e9af340ff4432f461265d90f508
f13bf48f06817a2a44f5dc903f4cd8ac8163fe2a
We cannot bisect more!
```

- a2212c010250c69ce581a9459683ae14a7b963ef
- fbb5789508286aa5a019f8b9e5371c3ecbf108d2
- 24c425b5e825ca5e2c91bdd863df9b5cdf33717f
- 3a93d4a843fa9e9af340ff4432f461265d90f508
- f13bf48f06817a2a44f5dc903f4cd8ac8163fe2a

This can be replicated by editing `hello_world` in hermit-rs and running

```console
cargo xtask ci qemu --arch x86_64 --profile dev --package hello_world --virtiofsd
```

CC: @cagatay-y

cagatay-y:
FUSE seems to expect in the length field of the header the total length of the command without the padding. This PR fixes #1111.

### DIFF
--- a/src/fs/fuse.rs
+++ b/src/fs/fuse.rs
@@ -482,7 +482,9 @@ where
 	fn new(nodeid: u64, op_header: O::InStruct) -> Box<Self> {
 		Box::new(Cmd {
 			in_header: fuse_abi::InHeader {
-				len: Layout::new::<Self>().size() as u32,
+				// The length we need the provide in the header is not the same as the size of the struct because of padding, so we need to calculate it manually.
+				len: (core::mem::size_of::<fuse_abi::InHeader>()
+					+ core::mem::size_of::<O::InStruct>()) as u32,
 				opcode: O::OP_CODE as u32,
 				nodeid,
 				unique: 1,
@@ -498,7 +500,10 @@ impl<O: ops::Op> Cmd<O> {
 	fn with_capacity(nodeid: u64, op_header: O::InStruct, len: usize) -> Box<UninitCmd<O>> {
 		let mut cmd = unsafe { Self::new_uninit(len) };
 		cmd.in_header = MaybeUninit::new(fuse_abi::InHeader {
-			len: core::mem::size_of_val(cmd.as_ref())
+			// The length we need the provide in the header is not the same as the size of the struct because of padding, so we need to calculate it manually.
+			len: (core::mem::size_of::<fuse_abi::InHeader>()
+				+ core::mem::size_of::<O::InStruct>()
+				+ len)
 				.try_into()
 				.expect("The command is too large"),
 			opcode: O::OP_CODE as u32,


mkroening (changes requested):
Thanks for the swift fix!

> I am confused about how this works. When I extract the expression with `slice_from_raw_parts_mut` into a variable, rust-analyzer reports the type as `*mut [u8]`. `len` is not the length of the entire struct, rather the payload, so the slice we create here does not cover the whole struct. I validated this by checking the length of the pointed slice.
>
> The reason I originally used 0 as the slice length was because that was how it was in the previous version of the code, but I think reverting 93fc6cb and using ptr_metadata here may be worth the unstable feature. I included a commit with the reversion in my PR (#1113) for #1111, but I can remove it if you prefer.

`len` never refers to the length of the entire struct but to the number of elements. For structs with DSTs as the last field, the metadata of the last field is used. So when whe create a pointer with the correct metadata via `ptr::slice_from_raw_parts_mut` everything should be fine, although it might be confusing at first. References:

- [Dynamically Sized Types - The Rust Reference](https://doc.rust-lang.org/stable/reference/dynamically-sized-types.html)
- [Slice DST pointer to pointer cast - Operator expressions - The Rust Reference](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#slice-dst-pointer-to-pointer-cast)

I still think `ptr_metadata` does not improve the situation here, aside from the fact that it will probably take ages to stabilize still. Please remove that commit and then I can merge this. :)

cagatay-y:
No problem, sorry for breaking the code earlier 😅

mkroening:
> No problem, sorry for breaking the code earlier 😅

No worries, it happens to everybody. :)

Issue #199, opened 2021-05-28
Use volatile memory access for memory-mapped I/O


mkroening:
Currently, we access memory-mapped I/O in a non-volatile fashion, which results in the compiler optimizing the access away under certain circumstances. We should make access to memory-mapped I/O volatile.

Instead of using [`pointer::read_volatile`](https://doc.rust-lang.org/std/primitive.pointer.html#method.read_volatile) and [`pointer::write_volatile`](https://doc.rust-lang.org/std/primitive.pointer.html#method.write_volatile) we should use an abstraction like [`volatile::Volatile`](https://docs.rs/volatile/0.4/volatile/struct.Volatile.html) by rust-osdev to enforce volatile memory access via the type system.

mkroening:
We'll want to wait for https://github.com/rust-osdev/volatile/pull/22 making the `Volatile` abstraction correct.

mkroening:
I think this is now done.

PR #1141, opened 2024-04-17
virtq: specify endianness

cagatay-y:
The Virtio specification specifies little-endian as the correct endianness for its structs. Although it matches the native endianness of the platforms we support, this change allows us to match the specification explicitly.


mkroening:
I am running into the same error with some completely unrelated changes, so this probably has nothing to do with this PR specifically.

mkroening:
The CI failure is https://github.com/hermit-os/kernel/issues/199 again. I'll come up with a proposal tomorrow. :)

PR #1148, opened 2024-04-21
chore: upgrade to nightly-2024-04-15

stlankes:
In addition, this PR rename ptr::from_exposed_addr to ptr::with_exposed_provenance, which is required by rust-lang/rust#122935


stlankes:
@mkroening With this PR, rust_demo should be able to read the directory entries.

mkroening (approved):

PR #1149, opened 2024-04-22
feat: add `virtio-def` crate for virtio definitions

mkroening:


stlankes (approved):

src/drivers/virtio/transport/mmio.rs
```diff
@@ -141,13 +142,13 @@ impl ComCfg {
 
 	/// Returns the device status field.
 	pub fn dev_status(&self) -> u8 {
-		unsafe { read_volatile(&self.com_cfg.status).try_into().unwrap() }
+		unsafe { read_volatile(&self.com_cfg.status).bits() }
```
cagatay-y, line 145: Is there a reason to return the underlying u8 instead of a DeviceStatus?

src/drivers/virtio/transport/pci.rs
```diff
@@ -219,6 +220,36 @@ impl PciCap {
 	pub fn dev_id(&self) -> u16 {
 		self.origin.dev_id
 	}
+
+	/// Returns a reference to the actual structure inside the PCI devices memory space.
+	fn map_common_cfg(&self) -> Option<VolatileRef<'static, CommonCfg>> {
+		if self.bar.length < u64::from(self.length + self.offset) {
+			error!("Common config of with id {} of device {:x}, does not fit into memory specified by bar {:x}!",
```
cagatay-y, line 227: The error message seems to be missing a word. "Common config of the capability with id {} of device {:x} does not fit into memory specified by bar {:x}!" maybe?

cagatay-y: It is not really the topic of this PR but thought we may want to fix it while we are at it.

src/drivers/virtio/transport/pci.rs
```diff
@@ -219,6 +220,36 @@ impl PciCap {
...
+		if self.bar.length < u64::from(self.length + self.offset) {
+			error!("Common config of with id {} of device {:x}, does not fit into memory specified by bar {:x}!", 
+			self.id,
+			self.origin.dev_id,
+			self.bar.index
+            );
+			return None;
+		}
+
+		// Using "as u32" is safe here as CommonCfg has a defined size smaller 2^31-1
```
cagatay-y, line 235: The comment seems to be outdated.

src/drivers/virtio/transport/pci.rs
mkroening: Done! 👍

src/drivers/virtio/transport/pci.rs
mkroening: Removed ✅

cagatay-y (approved):

PR #1150, opened 2024-04-22
virtq: use a struct with the actual layout rather than a struct of references

cagatay-y:
The code is not ready as I haven't yet figured out how the existing code ensures that the whole allocation is on the same page. This PR needs to be modified to ensure that as well. It is also based on #1141.

### DIFF
--- a/src/drivers/virtio/virtqueue/split.rs
+++ b/src/drivers/virtio/virtqueue/split.rs
@@ -11,7 +11,8 @@ use core::cell::RefCell;
 use core::mem::{size_of, MaybeUninit};
 use core::ptr;
 
-use zerocopy::{little_endian, FromBytes, FromZeroes};
+use virtio_def::num::{le16, le32, le64};
+use zerocopy::{FromBytes, FromZeroes};
 
 #[cfg(not(feature = "pci"))]
 use super::super::transport::mmio::{ComCfg, NotifCfg, NotifCtrl};
@@ -29,10 +30,10 @@ use crate::mm::device_alloc::DeviceAlloc;
 #[repr(C)]
 #[derive(Copy, Clone)]
 struct Descriptor {
-	address: little_endian::U64,
-	len: little_endian::U32,
-	flags: little_endian::U16,
-	next: little_endian::U16,
+	address: le64,
+	len: le32,
+	flags: le16,
+	next: le16,
 }
 
 impl Descriptor {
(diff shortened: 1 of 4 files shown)


cagatay-y:
I am not sure if the use of `Pin` here is correct and I need to check it but I think the PR can be reviewed since any changes related to `Pin` should not have a big effect for the majority of the changes.

src/drivers/virtio/virtqueue/split.rs
```diff
@@ -45,40 +48,113 @@ impl Descriptor {
...
-	used_ring: UsedRing,
+	token_ring: Box<[Option<Box<TransferToken>>]>,
+
+	// The following fields should not be used directly, as they do not ensure volatile access.
+	#[deprecated = "_*_cell fields are only for keeping the objects around. All access should be done through the respective *_ref methods."]
+	_descr_table_cell: Box<UnsafeCell<[MaybeUninit<Descriptor>]>, DeviceAlloc>,
+	#[deprecated = "_*_cell fields are only for keeping the objects around. All access should be done through the respective *_ref methods."]
+	_avail_ring_cell: Box<UnsafeCell<AvailRing>, DeviceAlloc>,
+	#[deprecated = "_*_cell fields are only for keeping the objects around. All access should be done through the respective *_ref methods."]
+	_used_ring_cell: Box<UnsafeCell<UsedRing>, DeviceAlloc>,
```
mkroening, line 135: I think we don't benefit from the noise of having `_` prefixes and deprecation attributes. Any access would have to be `unsafe` anyway, with the invariant being that access has to be volatile, which is conveniently provided by methods of the same name. The `_` prefix should only be used when we want to silence dead-code warnings but need the value around for side effects. Deprecation attributes should be used for things that are slated for removal, but these are not. Instead, it is very valid to use these fields when performing volatile accesses, which the wrapper method guarantees. We should change these attributes to be safety comments. Something like this:

```suggestion
    /// Descriptor Tables
    ///
    /// # Safety
    ///
    /// These tables may only be accessed via volatile operations.
    /// See the corresponding method for a safe wrapper.
	descr_table: Box<UnsafeCell<[MaybeUninit<Descriptor>]>, DeviceAlloc>,
	
	/// ...
	avail_ring: Box<UnsafeCell<AvailRing>, DeviceAlloc>,
	used_ring: Box<UnsafeCell<UsedRing>, DeviceAlloc>,
```

cagatay-y:
FYI, in addition to the deprecation change, I also switched a `ptr::from_mut` call to `ptr::from_ref` and changed the mutability of the variable declaration accordingly. It is not easily visible because I amended the last commit, so I am noting it here. 
https://github.com/hermit-os/kernel/blob/8e446e291ba25678ce769ef3fb7d3a851213311f/src/drivers/virtio/virtqueue/split.rs#L477-L485

mkroening (approved):
Thanks! :)

PR #1160, opened 2024-04-30
feat(mm): add `DeviceAlloc` for communicating with devices

mkroening:


src/arch/x86_64/kernel/gdt.rs
```diff
@@ -47,8 +49,10 @@ pub fn add_current_core() {
 			BasePageSize::SIZE as usize
 		};
 
-		let ist = crate::mm::allocate(sz, true);
-		let ist_start = ist.as_u64() + sz as u64 - TaskStacks::MARKER_SIZE as u64;
+		let layout = Layout::from_size_align(sz, BasePageSize::SIZE as usize).unwrap();
+		let ist = PhysAlloc.allocate(layout).unwrap();
+		let ist: u64 = ist.as_ptr().expose_provenance().try_into().unwrap();
+		let ist_start = ist + sz as u64 - TaskStacks::MARKER_SIZE as u64;
```
cagatay-y, line 55: Would offset methods on pointers be suitable here? Not a substantial change but I think it would reflect the meaning of what is happening here better. Otherwise, I think we can get the total first and then cast to u64 rather than casting each summand to `u64` and adding, since all are usize.

src/mm/mod.rs
```diff
@@ -297,7 +299,7 @@ pub(crate) fn print_information() {
 	arch::mm::virtualmem::print_information();
 }
 
-#[allow(dead_code)]
+/// Soft-deprecated in favor of `PhysAlloc`
```
cagatay-y, line 302: Maybe we can note that this deprecation does not apply for executable allocations.

src/mm/phys_alloc.rs
```diff
@@ -0,0 +1,29 @@
...
+/// Allocates whole pages of memory that correspond to a contigous range of physical frames.
+///
+/// This is useful for communicating with devices via physical memory.
+/// In this case, a struct must be contiguous in physical memory.
+pub struct PhysAlloc;
+
+unsafe impl Allocator for PhysAlloc {
+	fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, AllocError> {
+		assert!(layout.align() <= BasePageSize::SIZE as usize);
+		let size = layout.size().align_up(BasePageSize::SIZE as usize);
```
cagatay-y, line 17: Do we need to align_up, since `allocate` and `deallocate` already do?

src/mm/mod.rs
mkroening: There are no such allocations right now, so I think we can remove those functions entirely in the future. I think the last such allocation was removed in https://github.com/hermit-os/kernel/commit/544c6e5b44fcfc42fb890737fb1bd4b091722a19#diff-499078c6ea49b14787bc425331ad2b552b919d74ac6add023d7c16e261a3af37R488.

src/mm/phys_alloc.rs
mkroening: Yes, to return a slice with the bigger size. Once we merge `allocate` and `deallocate` into `PhysAlloc`, this will be deduplicated.

src/mm/phys_alloc.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use align_address::Align;
+
+use crate::arch::mm::paging::{BasePageSize, PageSize};
+
+/// Allocates whole pages of memory that correspond to a contigous range of physical frames.
+///
+/// This is useful for communicating with devices via physical memory.
+/// In this case, a struct must be contiguous in physical memory.
+pub struct PhysAlloc;
```
mkroening, line 12:
```suggestion
pub struct DeviceAlloc;
```

src/mm/mod.rs
mkroening: I have moved all non-device allocations (kernel stacks) to the heap.

PR #1165, opened 2024-05-02
feat(newlib): remove lwIP support

mkroening:
These functions can be removed once lwIP is removed from C user space.


stlankes:
Do we still the feature `newlib`?

mkroening:
> Do we still the feature `newlib`?

For now yes. But I am already looking for a way to remove it. For example:

- https://github.com/hermit-os/kernel/pull/656
- https://github.com/hermit-os/kernel/pull/1166

stlankes (approved):

PR #1166, opened 2024-05-02
feat(newlib): remove recmutex system calls

mkroening:
We should rework our C user space mutexes to use futexes, instead.


mkroening:
This is used in [pthread-embedded/platform/hermit/pte_osal.c#L600-L624](https://github.com/hermit-os/pthread-embedded/blob/e376740eacceb5e4ee092fa9a4d1c296bffa2521/platform/hermit/pte_osal.c#L600-L624). Removing `__malloc_lock()` and `__malloc_unlock()` are fine to remove since I removed Newlib's `sbrk()`-based allocator in https://github.com/hermit-os/newlib/pull/6 and we now use the kernel allocator. `__env_lock()` and `__env_unlock()` are still used in Newlib, though. They are exposed as `ENV_LOCK` and `ENV_UNLOCK` in [newlib/libc/include/envlock.h#L9-L10](https://github.com/hermit-os/newlib/blob/newlib-4.5/newlib/libc/include/envlock.h#L9-L10), which are used in getenv_r.c and setenv_r.c.

mkroening:
I am not sure if we can just remove getenv_r and setenv_r, but I believe that recursive mutexes are generally useful. They are also used in AML interpreters as well as parts of the pthread API, though they are not currently exposed. We could reimplement these primitives around `lock_api` in the future, but that is not urgent.

Closing this PR.

PR #1171, opened 2024-05-04
new: Add colored log levels

### LINKED ISSUE #905: Add colors to log levels

camnwalter:
Closes #905 
Also see https://github.com/hermit-os/loader/pull/336

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -1,4 +1,7 @@
-use log::{LevelFilter, Metadata, Record};
+use core::fmt;
+
+use anstyle::AnsiColor;
+use log::{Level, LevelFilter, Metadata, Record};
 
 /// Data structure to filter kernel messages
 struct KernelLogger;
@@ -17,11 +20,38 @@ impl log::Log for KernelLogger {
 			println!(
 				"[{}][{}] {}",
 				crate::arch::core_local::core_id(),
-				record.level(),
+				ColorLevel(record.level()),
 				record.args()
 			);
 		}
 	}
+}
+
+struct ColorLevel(Level);
+
+impl fmt::Display for ColorLevel {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		let level = self.0;
+
+		if no_color() {
+			write!(f, "{level}")
+		} else {
+			let color = match level {
+				Level::Trace => AnsiColor::Magenta,
+				Level::Debug => AnsiColor::Yellow,
+				Level::Info => AnsiColor::Green,
+				Level::Warn => AnsiColor::Blue,
+				Level::Error => AnsiColor::Red,
+			};
+
+			let style = anstyle::Style::new().fg_color(Some(color.into()));
+			write!(f, "{style}{level}{style:#}")
+		}
+	}
+}
+
+fn no_color() -> bool {
+	option_env!("NO_COLOR").is_some_and(|val| !val.is_empty())
 }
 
 pub unsafe fn init() {
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -38,9 +38,9 @@ impl fmt::Display for ColorLevel {
 		} else {
 			let color = match level {
 				Level::Trace => AnsiColor::Magenta,
-				Level::Debug => AnsiColor::Yellow,
+				Level::Debug => AnsiColor::Blue,
 				Level::Info => AnsiColor::Green,
-				Level::Warn => AnsiColor::Blue,
+				Level::Warn => AnsiColor::Yellow,
 				Level::Error => AnsiColor::Red,
 			};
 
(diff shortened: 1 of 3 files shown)


src/logging.rs
```diff
@@ -14,8 +14,18 @@ impl log::Log for KernelLogger {
 
 	fn log(&self, record: &Record<'_>) {
 		if self.enabled(record.metadata()) {
+			let color_code: &str = match record.level() {
+				log::Level::Error => "\x1b[31m", // red
+				log::Level::Debug => "\x1b[32m", // green
+				log::Level::Trace => "\x1b[35m", // magenta
+				log::Level::Warn => "\x1b[33m",  // yellow
+				log::Level::Info => "\x1b[34m",  // blue
```
mkroening, line 22: Please import `log::Level` in cases like this.

mkroening (changes requested):
Thanks a lot for the PR! :)

This is great, but I have a few visual and technical requests:

Visual requests:

1. I think we should only color the level (e.g., `WARN` in `[WARN]`) and not the whole line.
2. For color choice I would prefer if we would follow what tracing-subscriber is doing: [tracing-subscriber/src/fmt/format/mod.rs#L1504-L1510](https://github.com/tokio-rs/tracing/blob/tracing-subscriber-0.3.18/tracing-subscriber/src/fmt/format/mod.rs#L1504-L1510)
3. We should respect the [`NO_COLOR`](https://no-color.org/) environment variable at build time to disable any colors.

Technical requests:

I think it would be good, if we introduced a newtype `ColorLevel(Level)` that implements a colored 'Display' via [`anstyle`](/latest/anstyle/struct.Style.html).

camnwalter:
> I think it would be good, if we introduced a newtype `ColorLevel(Level)` that implements a colored 'Display' via [`anstyle`](/latest/anstyle/struct.Style.html).

anstyle comes with the stdlib, so do you want me to re-create what they do?

mkroening:
> anstyle comes with the stdlib, so do you want me to re-create what they do?

If you disable the default `std` feature, it should work:

```toml
anstyle = { version = "1", default-features = false }
```

mkroening (changes requested):
> anstyle comes with the stdlib, so do you want me to re-create what they do?

If you disable the default `std` feature, it should work:

```toml
anstyle = { version = "1", default-features = false }
```

src/logging.rs
```diff
@@ -17,13 +20,37 @@ impl log::Log for KernelLogger {
...
+
+impl fmt::Display for ColorLevel {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		let level: Level = self.0;
+		let color: AnsiColor = match level {
+			Level::Error => AnsiColor::Red,
+			Level::Debug => AnsiColor::Yellow,
+			Level::Trace => AnsiColor::Magenta,
+			Level::Info => AnsiColor::Green,
+			Level::Warn => AnsiColor::Blue,
```
mkroening, line 40:
```suggestion
			Level::Trace => AnsiColor::Magenta,
			Level::Debug => AnsiColor::Blue,
			Level::Info => AnsiColor::Green,
			Level::Warn => AnsiColor::Yellow,
			Level::Error => AnsiColor::Red,
```

src/logging.rs
```diff
@@ -17,13 +20,37 @@ impl log::Log for KernelLogger {
...
+		};
+
+		let no_color: Option<&'static str> = option_env!("NO_COLOR");
+
+		if no_color.is_none() || no_color.unwrap().is_empty() {
```
mkroening, line 45: Let's extract this into a function:
```rust
fn color() -> bool {
    option_env!("NO_COLOR").is_some_and(|val| !val.is_empty())
}
```

mkroening (changes requested):
Thanks, this is looking great. Could you remove the explicit types from the `let` bindings unless necessary? :)

src/logging.rs
```diff
@@ -17,13 +20,40 @@ impl log::Log for KernelLogger {
...
+
+		if no_color() {
+			write!(f, "{level}")
+		} else {
+			let color = match level {
+				Level::Trace => AnsiColor::Magenta,
+				Level::Debug => AnsiColor::Yellow,
+				Level::Info => AnsiColor::Green,
+				Level::Warn => AnsiColor::Blue,
+				Level::Error => AnsiColor::Red,
```
mkroening, line 44:
```suggestion
				Level::Trace => AnsiColor::Magenta,
				Level::Debug => AnsiColor::Blue,
				Level::Info => AnsiColor::Green,
				Level::Warn => AnsiColor::Yellow,
				Level::Error => AnsiColor::Red,
```

mkroening (changes requested):

src/logging.rs
camnwalter: Oops I don't know what happened there!

mkroening (approved):
Thanks! :)

camnwalter:
I'll do similar changes in the other pull request in a little bit

PR #1184, opened 2024-05-13
feat(sys): add `sys_available_parallelism`

mkroening:
We could deprecate `sys_get_processor_count` in favor of `sys_available_parallelism`.

### DIFF
--- a/src/syscalls/processor.rs
+++ b/src/syscalls/processor.rs
@@ -6,6 +6,11 @@ pub extern "C" fn sys_get_processor_count() -> usize {
 	get_processor_count().try_into().unwrap()
 }
 
+#[hermit_macro::system]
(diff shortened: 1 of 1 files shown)


stlankes:
Why is this name more useful? I like the naming from [num_cpus](https://github.com/seanmonstar/num_cpus/blob/master/src/lib.rs#L431C1-L434C2).

mkroening:
I don't feel strong on this, but the reasoning behind Rust naming it like this (https://github.com/rust-lang/rust/issues/74479#issuecomment-910143466, https://github.com/rust-lang/rust/pull/89324) makes sense to me:

For the user, it is important to know how many threads they can spawn to make use of the available concurrency of the system. It is not relevant to how many physical CPU cores or hyperthreads or hardware threads this maps.

stlankes (approved):

PR #1204, opened 2024-05-17
fuse: provide header and payloads separately

cagatay-y:
Providing headers and payloads separately allows us to avoid the complexity associated with dynamically-sized types and in some cases copying.


cagatay-y:
813ad4359e3623ebade1bdf3596cc32719984bb0 does not really fit into the theme of this PR as originally the branch was intended to be a more general Virtio work and only later I realized the parts related to FUSE can be merged independently. I can remove the commit, if you prefer.

PR #1207, opened 2024-05-19
fix conversion from IpAddress::Ipv4 to in_addr

stlankes:
IpAddress::Ipv4 contains alreads the IP address in big-endian formt. => no byte swapping needed

With these changes, the demo httpd shows the correct address of the client.


mkroening (approved):
Ah, that's right. 👍

PR #1216, opened 2024-05-21
Added no_mangle to all extern C functions

CarlWachter:
Extern C functions were not externally available due to a missing no_mangle.

### DIFF
--- a/hermit-macro/src/system.rs
+++ b/hermit-macro/src/system.rs
@@ -59,13 +59,27 @@ fn parse_sig(sig: &Signature) -> Result<ParsedSig> {
 }
 
 fn validate_attrs(attrs: &[Attribute]) -> Result<()> {
+	let mut no_mangle_found = false;
 	for attr in attrs {
-		if !attr.path().is_ident("cfg") && !attr.path().is_ident("doc") {
+		if !attr.path().is_ident("cfg")
+			&& !attr.path().is_ident("doc")
+			&& !attr.path().is_ident("no_mangle")
+		{
 			bail!(
 				attr,
-				"#[system] functions may only have `#[doc]` and `#[cfg]` attributes"
+				"#[system] functions may only have `#[doc]`, `#[no_mangle]` and `#[cfg]` attributes"
 			);
 		}
+		if attr.path().is_ident("no_mangle") {
+			no_mangle_found = true;
+		}
+	}
+
+	if !no_mangle_found {
+		bail!(
+			attrs.first(),
+			"#[system] functions must have `#[no_mangle]` attribute"
+		);
 	}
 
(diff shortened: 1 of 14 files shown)


hermit-macro/src/system.rs
```diff
@@ -60,10 +60,13 @@ fn parse_sig(sig: &Signature) -> Result<ParsedSig> {
 
 fn validate_attrs(attrs: &[Attribute]) -> Result<()> {
```
mkroening, line 61: Can you also add a check that `#[no_mangle]` must be present? Also, this macro should not add `#[no_mangle]` anymore.

mkroening (changes requested):

hermit-macro/src/system.rs
CarlWachter: Added

CarlWachter: Added

mkroening (approved):
Thanks! :)

hermit-macro/src/system.rs
```diff
@@ -59,13 +59,27 @@ fn parse_sig(sig: &Signature) -> Result<ParsedSig> {
...
+		{
 			bail!(
 				attr,
-				"#[system] functions may only have `#[doc]` and `#[cfg]` attributes"
+				"#[system] functions may only have `#[doc]`, `#[no_mangle]` and `#[cfg]` attributes"
 			);
 		}
+		if attr.path().is_ident("no_mangle") {
+			no_mangle_found = true;
+		}
```
jounathaen, line 75: You could write this as `no_mangle_found = attr.path().is_ident("no_mangle");`

mkroening: No, we can't, this is a loop over multiple attrs.

hermit-macro/src/system.rs
```diff
@@ -59,13 +59,27 @@ fn parse_sig(sig: &Signature) -> Result<ParsedSig> {
...
+	}
+
+	if !no_mangle_found {
+		bail!(
+			attrs.first(),
+			"#[system] functions must have `#[no_mangle]` attribute"
+		);
 	}
```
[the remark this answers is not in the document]
mkroening: Not sure where if the best place to document the system calls would be here or in hermit-abi or somewhere else entirely. It might be a good idea eventually, but this is out of scope for this PR.

hermit-macro/src/system.rs
jounathaen: Yes, we can. Obviously, this is part of the loop.

jounathaen: ah, now I see. Nevermind.

PR #1218, opened 2024-05-21
Virtio bugfix

jounathaen:
These are two Bugs, where I am wondering how Hermit has event worked before... :thinking:


src/drivers/virtio/transport/pci.rs
```diff
@@ -702,9 +702,9 @@ impl NotifCfg {
 
 		// Assumes the cap_len is a multiple of 8
 		// This read MIGHT be slow, as it does NOT ensure 32 bit alignment.
-		let notify_off_multiplier = cap.device.read_register(
-			u16::try_from(cap.origin.cfg_ptr).unwrap() + u16::from(cap.origin.cap_struct.cap_len),
-		);
+		let notify_off_multiplier = cap
+			.device
+			.read_register(u16::try_from(cap.origin.cfg_ptr).unwrap() + 16);
```
stlankes, line 707: Why do is 16 correct and cap.origin.cap_struct.cap_len wrong? 

Is the following line also wrong?

https://github.com/hermit-os/kernel/blob/main/src/drivers/virtio/transport/pci.rs#L950-L952

jounathaen: This needs to read the last 4 bytes of the struct. Adding `cap_len` results in reading the first byte of the next struct. Likely that the other calculation is also incorrect.

mkroening (changes requested):
Can we squash the commits if they don't pass CI individually? :)

jounathaen:
Before merging, can sbdy test if this breaks something with QEMU? I don't have a test setup at hand

mkroening:
> Before merging, can sbdy test if this breaks something with QEMU? I don't have a test setup at hand

CI passes and we test virtio-net-pci there, so everything should be fine, I think?

src/drivers/virtio/transport/pci.rs
mkroening: I have fixed the other two uses of `cfg_ptr` too in the latest commits. Could you take a look?

jounathaen: A little verbose but looks good :+1:

mkroening (approved):

PR #1219, opened 2024-05-21
add system calls `readv` and `writev`

stlankes:
the offers the possibility to read or to write data into multiple buffers

See https://github.com/hermit-os/hermit-rs/pull/580.

### DIFF
--- a/src/syscalls/mod.rs
+++ b/src/syscalls/mod.rs
@@ -72,7 +72,7 @@ struct iovec {
 	pub iov_len: usize,
 }
 
-const IOV_MAX: usize = 1024;
+const IOV_MAX: i32 = 1024;
 
 pub(crate) fn init() {
 	Lazy::force(&SYS);
@@ -433,13 +433,14 @@ pub unsafe extern "C" fn sys_read(fd: FileDescriptor, buf: *mut u8, len: usize) 
 }
 
 #[hermit_macro::system]
-pub unsafe extern "C" fn sys_readv(fd: i32, iov: *const iovec, iovcnt: usize) -> isize {
+#[no_mangle]
+pub unsafe extern "C" fn sys_readv(fd: i32, iov: *const iovec, iovcnt: i32) -> isize {
 	if iovcnt > IOV_MAX {
 		return (-crate::errno::EINVAL).try_into().unwrap();
 	}
 
 	let mut count: isize = 0;
-	let slice = unsafe { core::slice::from_raw_parts(iov, iovcnt) };
+	let slice = unsafe { core::slice::from_raw_parts(iov, iovcnt.try_into().unwrap()) };
 
 	for i in slice {
 		let buf = unsafe { core::slice::from_raw_parts_mut(i.iov_base, i.iov_len) };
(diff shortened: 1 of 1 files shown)


src/syscalls/mod.rs
```diff
@@ -417,6 +459,37 @@ pub unsafe extern "C" fn sys_write(fd: FileDescriptor, buf: *const u8, len: usiz
 	unsafe { write(fd, buf, len) }
 }
 
+#[hermit_macro::system]
+pub unsafe extern "C" fn sys_writev(fd: FileDescriptor, iov: *const iovec, iovcnt: i32) -> isize {
+	if iovcnt < 0 || iovcnt > IOV_MAX {
+		return (-crate::errno::EINVAL).try_into().unwrap();
```
jounathaen, line 465: why is negative `EINVAL` returned?

src/syscalls/mod.rs
```diff
@@ -417,6 +459,37 @@ pub unsafe extern "C" fn sys_write(fd: FileDescriptor, buf: *const u8, len: usiz
 	unsafe { write(fd, buf, len) }
 }
 
+#[hermit_macro::system]
+pub unsafe extern "C" fn sys_writev(fd: FileDescriptor, iov: *const iovec, iovcnt: i32) -> isize {
```
jounathaen, line 463: It is not done for the other functions, but it might be nice to add a brief doc comment to the external ABI, specifically if it is an `unsafe` fn and has some safety requirements (`iovec` needs to be a valid buffer of size `iovcnt` here)

jounathaen, line 463:
```suggestion
/// Read `iovvcnt` bytes from `fd` into a buffer `iovec` .
///
/// # Safety:
///
/// - `iovec` must be safely writeable.
/// - `iovec` must be at least `iovcnt` bytes long 
pub unsafe extern "C" fn sys_writev(fd: FileDescriptor, iov: *const iovec, iovcnt: i32) -> isize {
```

src/syscalls/mod.rs
```diff
@@ -417,6 +459,37 @@ pub unsafe extern "C" fn sys_write(fd: FileDescriptor, buf: *const u8, len: usiz
...
+	if iovcnt < 0 || iovcnt > IOV_MAX {
+		return (-crate::errno::EINVAL).try_into().unwrap();
+	}
+
+	let mut count: isize = 0;
+	let slice = unsafe { core::slice::from_raw_parts(iov, iovcnt.try_into().unwrap()) };
```
jounathaen, line 469: Isn't `iov` a struct that contains a pointer to the buffer? So we should rather read the first element from it and also add a boundary check against the contained `iov_len`?

jounathaen: Nevermind, I overlooked that `iov` is an array of `iovec`.

src/syscalls/mod.rs
```diff
@@ -417,6 +459,37 @@ pub unsafe extern "C" fn sys_write(fd: FileDescriptor, buf: *const u8, len: usiz
...
+
+	for i in slice {
```
jounathaen, line 471: Nitpick: I'd suggest renaming `i` to `buf` and `slice` to `buffers`

src/syscalls/mod.rs
```diff
@@ -63,6 +63,17 @@ pub(crate) static SYS: Lazy<&'static dyn SyscallInterface> = Lazy::new(|| {
 	}
 });
 
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+struct iovec {
```
jounathaen, line 68:
```suggestion
/// Describes  a  region  of  memory, beginning at `iov_base` address and with the size of `iov_len` bytes.
struct iovec {
```

src/syscalls/mod.rs
stlankes: I added a description

stlankes: I added already a similar text

src/syscalls/mod.rs
stlankes: I fix it (see last commit)

src/syscalls/mod.rs
```diff
@@ -417,6 +459,37 @@ pub unsafe extern "C" fn sys_write(fd: FileDescriptor, buf: *const u8, len: usiz
...
+		return (-crate::errno::EINVAL).try_into().unwrap();
+	}
+
+	let mut count: isize = 0;
+	let slice = unsafe { core::slice::from_raw_parts(iov, iovcnt.try_into().unwrap()) };
+
+	for i in slice {
+		let buf = unsafe { core::slice::from_raw_parts(i.iov_base, i.iov_len) };
+
+		let len = crate::fd::write(fd, buf).map_or_else(
```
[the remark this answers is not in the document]
stlankes: I fix it (see last commit)

jounathaen (approved):

PR #1241, opened 2024-05-30
refactor(virtqueue): make several types smaller

mkroening:


cagatay-y (approved):
I don't see any issue regarding the first commit. For the second one, I don't think any of the two states of the wrap counter are more true or false than the other but if it has an performance advantage I don't see an issue with it too.

mkroening:
> For the second one, I don't think any of the two states of the wrap counter are more true or false than the other

I changed it to a `u8`. 👍

PR #1243, opened 2024-06-01
add option to use a busy waiting within the idle loop

stlankes:
Instead of the `hlt` instruction, the idle loop use the `pause` instruction. => lower latency to wake up the core, but higher energy consumption.

In addition, this PR support `monitor`/`mwait` instructions. In this, the kernel hasn't to send an IPI to wakeup a core.  The second optimization is that the kernels sends only an IPI, if the core is halted.

### DIFF
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -943,7 +943,7 @@ pub(crate) fn add_current_core() {
 
 #[inline]
 #[cfg(all(target_arch = "x86_64", feature = "smp"))]
-pub(crate) fn get_core_hlt_state(core_id: CoreId) -> bool {
+pub(crate) fn take_core_hlt_state(core_id: CoreId) -> bool {
 	CORE_HLT_STATE.lock()[usize::try_from(core_id).unwrap()].swap(false, Ordering::Acquire)
 }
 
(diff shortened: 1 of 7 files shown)


src/scheduler/mod.rs
```diff
@@ -920,9 +934,19 @@ pub(crate) fn add_current_core() {
...
+		#[cfg(target_arch = "x86_64")]
+		CORE_HLT_STATE
+			.lock()
+			.insert(core_id.try_into().unwrap(), &CoreLocal::get().hlt);
 	}
 }
 
+#[inline]
+#[cfg(all(target_arch = "x86_64", feature = "smp"))]
+pub(crate) fn get_core_hlt_state(core_id: CoreId) -> bool {
```
mkroening, line 946:
```suggestion
pub(crate) fn take_core_hlt_state(core_id: CoreId) -> bool {
```

Since this is where `hlt` is set to `false` again, which is stronger than just getting.

mkroening (approved):
Looks good to me aside from one nit. I'll run CI on the in-between commits too, to make sure there's no breakage.

src/scheduler/mod.rs
mkroening: I applied the change.

Issue #1260, opened 2024-06-06
Reading uninitialized memory from FUSE


mkroening:
https://github.com/hermit-os/kernel/pull/1248 has mare it mode likely that we end up reading uninitialized memory for FUSE. I suspect this is coming from subtle codegen changes (aka we are doing something wrong with uninit memory or references).

```
$ cargo xtask ci qemu --arch x86_64 --profile release --package rftrace-example --virtiofsd
...
Saving traces to disk...!
0xc1b000, Events { ptr: 0xc1b000, len: 2000, cap: 2000 }
  Parsing TID 1...!
  Writing to disk: 8 events, 128 bytes (/root/tracedir/1.dat)
  Parsed all events!
Creating fake uftrace data dir at /root/tracedir..
  Creating ./info
    feats = TASK_SESSION | SYM_REL_ADDR
    info = CMDLINE | TASKINFO
    cmdline = 'fakeuftrace'
    tid = [1]
  Creating ./task.txt
[0][PANIC] panicked at src/fs/fuse.rs:479:47:
called `Result::unwrap()` on an `Err` value: TryFromIntError(())
```

This happens while converting a `u64` into an `i64`, so this is most definitely not a real value.

https://github.com/hermit-os/kernel/blob/a723af0ae2de31d04474e290b2e33d253a3cabfe/src/fs/fuse.rs#L479

I'll disable the test in CI for now.

mkroening:
Closing as of https://github.com/hermit-os/kernel/pull/1285

PR #1261, opened 2024-06-06
ci: allow failing rftrace in release mode

mkroening:
See https://github.com/hermit-os/kernel/issues/1260.
CC: @cagatay-y


mkroening:
Part of https://github.com/hermit-os/kernel/pull/1259.

PR #1274, opened 2024-06-12
fix(virtqueue): fix driver notifications

mkroening:
I found these while working on https://github.com/hermit-os/kernel/pull/1273.

Best reviewed commit by commit.

### DIFF
--- a/src/drivers/virtio/virtqueue/packed.rs
+++ b/src/drivers/virtio/virtqueue/packed.rs
@@ -881,10 +881,7 @@ impl DrvNotif {
 		// Check if VIRTIO_F_RING_EVENT_IDX has been negotiated
 		if self.f_notif_idx {
 			self.raw.flags |= 1 << 1;
-			// Reset event fields
-			self.raw.event = 0;
-			self.raw.event = at_offset;
-			self.raw.event |= (at_wrap as u16) << 15;
+			self.raw.event = at_offset | (at_wrap as u16) << 15;
 		}
 	}
 }
--- a/src/drivers/virtio/virtqueue/packed.rs
+++ b/src/drivers/virtio/virtqueue/packed.rs
@@ -11,6 +11,7 @@ use core::ptr;
 use core::sync::atomic::{fence, Ordering};
 
 use align_address::Align;
+use virtio::pci::NotificationData;
 use zerocopy::little_endian;
 
 #[cfg(not(feature = "pci"))]
@@ -969,8 +970,11 @@ impl Virtq for PackedVq {
 		}
 
 		if self.dev_event.is_notif() | self.dev_event.is_notif_specfic(next_off, next_wrap) {
-			self.notif_ctrl
-				.notify_dev(self.index.0, next_off, next_wrap);
+			let notification_data = NotificationData::new()
+				.with_vqn(self.index.0)
+				.with_next_off(next_off)
+				.with_next_wrap(next_wrap);
+			self.notif_ctrl.notify_dev(notification_data);
 		}
 	}
 
@@ -997,8 +1001,11 @@ impl Virtq for PackedVq {
 		}
 
 		if self.dev_event.is_notif() {
-			self.notif_ctrl
-				.notify_dev(self.index.0, next_off, next_wrap);
+			let notification_data = NotificationData::new()
+				.with_vqn(self.index.0)
+				.with_next_off(next_off)
+				.with_next_wrap(next_wrap);
+			self.notif_ctrl.notify_dev(notification_data);
 		}
 	}
 
@@ -1012,8 +1019,11 @@ impl Virtq for PackedVq {
 		}
 
 		if self.dev_event.is_notif() {
-			self.notif_ctrl
-				.notify_dev(self.index.0, next_off, next_wrap);
+			let notification_data = NotificationData::new()
+				.with_vqn(self.index.0)
+				.with_next_off(next_off)
+				.with_next_wrap(next_wrap);
+			self.notif_ctrl.notify_dev(notification_data);
 		}
 	}
 
--- a/src/drivers/virtio/virtqueue/packed.rs
+++ b/src/drivers/virtio/virtqueue/packed.rs
@@ -5,10 +5,10 @@
 use alloc::boxed::Box;
 use alloc::rc::Rc;
 use alloc::vec::Vec;
-use core::cell::RefCell;
+use core::cell::{Cell, RefCell};
 use core::mem::MaybeUninit;
-use core::ptr;
 use core::sync::atomic::{fence, Ordering};
+use core::{ops, ptr};
 
 use align_address::Align;
 use virtio::pci::NotificationData;
@@ -26,10 +26,29 @@ use super::{
 use crate::arch::mm::paging::{BasePageSize, PageSize};
 use crate::arch::mm::{paging, VirtAddr};
 
-#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+#[derive(Default, PartialEq, Eq, Clone, Copy, Debug)]
 struct RingIdx {
 	off: u16,
 	wrap: u8,
+}
+
+trait RingIndexRange {
+	fn wrapping_contains(&self, item: &RingIdx) -> bool;
+}
+
+impl RingIndexRange for ops::Range<RingIdx> {
+	fn wrapping_contains(&self, item: &RingIdx) -> bool {
+		let ops::Range { start, end } = self;
+
+		if start.wrap == end.wrap {
+			item.wrap == start.wrap && start.off <= item.off && item.off < end.off
+		} else if item.wrap == start.wrap {
+			start.off <= item.off
+		} else {
+			debug_assert!(item.wrap == end.wrap);
+			item.off < end.off
+		}
+	}
 }
 
 /// A newtype of bool used for convenience in context with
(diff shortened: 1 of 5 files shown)


src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -872,17 +872,17 @@ impl DrvNotif {
 	/// Disables notifications by unsetting the LSB.
```
cagatay-y, line 872: The comment refers to the previous incorrect behaviour.

mkroening: Fixed. 👍

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -863,26 +863,26 @@ impl EventSuppr {
...
 	}
 
 	/// Enables a notification by the device for a specific descriptor.
 	fn enable_specific(&mut self, at_offset: u16, at_wrap: u8) {
 		// Check if VIRTIO_F_RING_EVENT_IDX has been negotiated
 		if self.f_notif_idx {
-			self.raw.flags |= 1 << 1;
+			self.raw.flags = 2;
 			// Reset event fields
 			self.raw.event = 0;
```
cagatay-y, line 884: Since we specifically unset the wrap bit now, do we still need this line?

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -897,12 +897,12 @@ impl DevNotif {
 	/// Reads notification bit (i.e. LSB) and returns value.
 	/// If notifications are enabled returns true, else false.
 	fn is_notif(&self) -> bool {
-		self.raw.flags & (1 << 0) == 0
+		self.raw.flags & 0b11 == 0
```
cagatay-y, line 900: I think at least one of the uses of this method is written with the assumption that it checks only for general notifications: 
https://github.com/hermit-os/kernel/blob/737b1388d9d7a62c01a61499c71564d4f70abb5d/src/drivers/virtio/virtqueue/packed.rs#L973

If I understand correctly, for that one and the remaining two uses 
https://github.com/hermit-os/kernel/blob/737b1388d9d7a62c01a61499c71564d4f70abb5d/src/drivers/virtio/virtqueue/packed.rs#L1001 
https://github.com/hermit-os/kernel/blob/737b1388d9d7a62c01a61499c71564d4f70abb5d/src/drivers/virtio/virtqueue/packed.rs#L1016
the behaviour is not correct either way, as there is no check for descriptor specific notification. I guess falling back to always notifying even when VIRTIO_F_EVENT_IDX had been negotiated is the better option, though it may be best to stop advertising support for it.

src/drivers/virtio/virtqueue/packed.rs
mkroening: True. It was the same case before, but I will fix it. 👍

src/drivers/virtio/virtqueue/packed.rs
mkroening: So what you are saying is that the latter two uses should also check for `is_notif_specific`? Or what kind of descriptor-specific notification check is missing?

cagatay-y: Sorry, I missed that the offset condition is checked inside `is_notif_specific`.

I think ideally, we would check specifically for the LSB (the current behaviour) in `is_notif` (maybe we can change its name to make it more clear, something like `is_generic_notif_enabled`), and change the if statement to be like the one in the first example. This way, the driver would not send a notification if the device asked for notification for a specific descriptor.

However, I think the way `is_notif_specific` is currently implemented, this would not work for batch transfers. If we push a batch past the descriptor the device specified, we will fail to send a notification since the specific notification condition check is at the end and is a equality check, not a greater-equals one.

For this reason, I think we can continue with the new implementation of `is_notif` in the PR, since sending a notification for each dispatch rather than a specific one is a "SHOULD NOT" and not a "MUST NOT", and will guarantee that we do not skip sending a notification when we need to. However, since this means we effectively do not support descriptor specific notifications, it may be preferable not to advertise support for `VIRTIO_F_EVENT_IDX`, too. We can also remove the second part of the "or" expression in the first example, since the new implementation means that if the first operand is false, the second operand is guaranteed to be false, too.

A future PR that correctly implements the descriptor specific notification also for batch dispatches can reenable support for `VIRTIO_F_EVENT_IDX` and implement `is_generic_notif_enabled`, along with other necessary changes.
  
For reference, the relevant section of the specification (VIRTIO Spec. v1.2 § 2.7.10.1 (the section number is probably different for version 1.1)):
> if the VIRTIO_F_EVENT_IDX feature bit is negotiated:  • The driver MUST ignore the lower bit of flags.  • After the driver writes a descriptor index into the available ring:  – If the idx field in the available ring (which determined where that descriptor index was placed) was equal to avail_event, the driver MUST send a notification.  – Otherwise the driver SHOULD NOT send a notification.

(Note: the quote is from the split queue section, as the packed queue section does not go into as much detail. I would assume that they should behave similarly in this specific issue.)

### Alternative that may be outside the scope of this PR
We can rename the existing function, add the implementation in the PR as a new function, use the new function in the `dispatch_batch{,_await}` functions and change the if condition inside the non-batch dispatch to be a or statement like the one in the first example above.

cagatay-y: I should note that I don't have a lot of confidence in my understanding of the notification part of the specification as I previously focused mostly on the handling of the descriptors.

mkroening: I have implemented it the way we talked about.

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -963,17 +989,25 @@ impl Virtq for PackedVq {
...
 
-		if self.dev_event.is_notif() | self.dev_event.is_notif_specfic(next_off, next_wrap) {
-			self.notif_ctrl
-				.notify_dev(self.index.0, next_off, next_wrap);
+		let range = self.last_notify.get()..next_idx;
+		let notif_specfic = self
+			.dev_event
+			.notif_specfic()
+			.map_or(false, |idx| range.wrapping_contains(&idx));
+		self.last_notify.set(next_idx);
```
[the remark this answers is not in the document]
mkroening: Good question!

I think we should only notify for any transfers just performed. I don't think a device would except a notification for an old transfer. That would also potentially create problems with wrapping around.

What do you think?

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -898,24 +925,22 @@ impl DevNotif {
...
 	}
 
-	fn is_notif_specfic(&self, next_off: u16, next_wrap: u8) -> bool {
-		if self.f_notif_idx {
-			if self.raw.flags & 1 << 1 == 2 {
-				// as u16 is okay for usize, as size of queue is restricted to 2^15
-				// it is also okay to just loose the upper 8 bits, as we only check the LSB in second clause.
-				let desc_event_off = self.raw.event & !(1 << 15);
-				let desc_event_wrap = (self.raw.event >> 15) as u8;
+	fn notif_specfic(&self) -> Option<RingIdx> {
```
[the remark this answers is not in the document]
mkroening: Fixed 👍

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -940,6 +965,7 @@ pub struct PackedVq {
 	/// The virtqueues index. This identifies the virtqueue to the
 	/// device and is unique on a per device basis.
 	index: VqIndex,
+	last_notify: Cell<RingIdx>,
```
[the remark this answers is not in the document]
mkroening: True. 👍

virtio-spec/src/pci.rs
```diff
@@ -471,3 +471,52 @@ pub struct NotificationData {
     #[bits(1)]
     pub next_wrap: u8,
 }
+
+impl NotificationData {
```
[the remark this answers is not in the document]
mkroening: It is already reexported from `mmio`. We can't make it global, since it has a different definition for Virtio Over Channel I/O.

src/drivers/virtio/virtqueue/packed.rs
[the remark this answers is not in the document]
cagatay-y: I think that would depend on the device checking to see if the specific descriptor it wants to request a notification for has already been written. If it does not, it would not be aware of the fact that it is an old dispatch.

What would be the problem with wrap-arounds? The driver cannot wrap around more than once before sending a notification to the device since the slots on the ring would be exhausted, if that is relevant.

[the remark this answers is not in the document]
mkroening: Good point. I changed it. 👍

cagatay-y (approved):

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -278,11 +304,13 @@ impl DescriptorRing {
 		self.ring[usize::from(first_ctrl_settings.0)].flags |=
 			first_ctrl_settings.2.as_flags_avail().into();
 
-		// Converting a boolean as u8 is fine
-		(first_ctrl_settings.0, first_ctrl_settings.2 .0 as u8)
+		RingIdx {
```
cagatay-y, line 307: We also need to update this one.

cagatay-y (changes requested):

src/drivers/virtio/virtqueue/packed.rs
mkroening: Fixed 👍

src/drivers/virtio/virtqueue/packed.rs
```diff
@@ -286,7 +286,7 @@ impl DescriptorRing {
 			}
 
 			if i == 0 {
-				first_ctrl_settings = (ctrl.start, ctrl.buff_id, ctrl.wrap_at_init);
+				first_ctrl_settings = (ctrl.position, ctrl.buff_id, self.drv_wc);
```
cagatay-y, line 289: I think we actually need the start index and the initial wrap counter state here. For multi-descriptor transfers, we delay updating the information for the first descriptor until all the other descriptors are written, so that the device never observes a partial chain. Normally, that is handled by the `WriteCtrl` and  `make_avail`. However, because in batch push we want to delay the update of the flags of the first descriptor not only until the rest of the chain is written but the whole batch is written, we need to keep the `WriteCtrl` state for the first descriptor around.

mkroening: Makes sense. 👍

cagatay-y (approved):

PR #1284, opened 2024-06-19
virtq: various simplifications

cagatay-y:


src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -2084,9 +2081,9 @@ impl MemDescr {
 	fn into_vec(mut self) -> Vec<u8> {
 		// Prevent double frees, as ownership will be tracked by
 		// Box from now on.
-		self.dealloc = Dealloc::Not;
+		self.dealloc = false;
 
-		unsafe { Vec::from_raw_parts(self.ptr, self._mem_len, 0) }
+		unsafe { Vec::from_raw_parts(self.ptr, self.len, self._init_len) }
 	}
```
mkroening, line 1709: This needs to return a `Vec<u8, DeviceAlloc>` to not get memory corruption. It would be good if we also allocate the memory as `Vec` to begin with and use `Vec::into_raw_with_alloc` for symmetry.

mkroening (changes requested):
Thanks!

Except for the first commit, this looks good to me.

Regarding the last commit: Does descriptor ID exhaustion happen? What motivates this change?

mkroening (approved):
Thanks!

Issue #1291, opened 2024-06-24
Rename `env.rs` to `args.rs` or `cli.rs`.


jounathaen:
My association of the name `env` would be _environment variables_. However, `env.rs` is only about kernel parameters/the CLI.

I suggest renaming `env.rs` to `args.rs` or `cli.rs`

mkroening:
This currently follows [`std::env`](https://doc.rust-lang.org/stable/std/env/index.html).

The module is about inspection of the kernel's environment, be it via kernel arguments or environment variables or boot information or similar.

mkroening:
> But `std::env` is about _Environment_ variables ("Userspace"). Here we are talking about the kernel "runtime" parameters.

Exactly, `std::env` is about the environment of the application and `hermit::env` is about the environment of the kernel, which is a superset of the environment of the application. 

`Cli` parses kernel arguments and provides them as kernel environment variables to other components of the kernel (`hermit::env::vars` correspond to `std::env::vars`). The fact that these are provided via kernel arguments instead of any other mechanism is an implementation detail.

> The struct `cli` contains a variable `env_vars` which is filled with four possible environment variables from the kernel command line (in `arch::init`)
> 
>     * -> These are only the "environment variables" relevant for the kernel (these actually aren't environment variables at all).
> 
>     * An application can't read them

Being a top-level module in does not imply exposition to the user.

> * The "real"/application relevant environment variables are retrieved via systemcall

Yes, most system calls live in `hermit::syscalls`.

PR #1292, opened 2024-06-24
Renaming of `env.rs` to `runtime_params.rs`

jounathaen:
And some minor related non-functional code improvements

fixes: https://github.com/hermit-os/kernel/issues/1291


mkroening (changes requested):
I'm opposed to this change as explained in https://github.com/hermit-os/kernel/issues/1291, but let's talk about this in person.

PR #1300, opened 2024-06-28
fix(virtqueue/split): fix off-by-one error

mkroening:


src/drivers/virtio/virtqueue/split.rs
```diff
@@ -60,7 +60,7 @@ impl AvailRing {
 		volatile_self: VolatilePtr<'_, Self, A>,
 	) -> VolatilePtr<'_, MaybeUninit<le16>, A> {
 		let ring_and_event_ptr = map_field!(volatile_self.ring_and_event);
-		ring_and_event_ptr.index(ring_and_event_ptr.len() - 2)
+		ring_and_event_ptr.index(ring_and_event_ptr.len() - 1)
```
cagatay-y, line 63: This reverses the change in the other commit, right? If so, I think we should amend the original commit.

src/drivers/virtio/virtqueue/split.rs
```diff
@@ -429,7 +429,7 @@ impl Virtq for SplitVq {
 				Box::from_raw_in(
 					core::ptr::slice_from_raw_parts_mut(
 						allocation.as_mut_ptr(),
-						ring_and_event_layout.size(),
+						usize::from(size) + 1,
```
cagatay-y, line 432: I am confused about this one. `UsedRing` stores the last field as a [u8] because the ring elements and the event flag are not the same size. Shouldn't we give their combined size as the slice length? We are casting from `*mut [u8]` to `*mut UnsafeCell<UsedRing>` (effectively `*mut UnsafeCell<[u8]>`) and according to the Rust Reference

> If T and U are both unsized, the pointer is also returned unchanged. In particular, the metadata is preserved exactly.
> 
> For instance, a cast from *const [T] to *const [U] preserves the number of elements. Note that, as a consequence, such casts do not necessarily preserve the size of the pointer's referent (e.g., casting *const [u16] to *const [u8] will result in a raw pointer which refers to an object of half the size of the original). The same holds for str and any compound type whose unsized tail is a slice type, such as struct Foo(i32, [u8]) or (u64, Foo).

, so I think we should give the length in bytes here.

src/drivers/virtio/virtqueue/split.rs
mkroening: Oops

src/drivers/virtio/virtqueue/split.rs
mkroening: Ah, that's right. I got confused, since Avail Ring is [`le16`] and Used Ring is `[u8]`.

cagatay-y (approved):

PR #1301, opened 2024-06-28
refactor(virtqueue/split): migrate to `virtq::Avail` and `virtq::Used`

mkroening:


virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        if !ptr.cast::<UsedElem>().is_aligned() {
+            return None;
+        }
+
+        if len % mem::size_of::<le16>() != 0 {
+            return None;
+        }
+
+        let len = len / mem::size_of::<le16>() - 2 - usize::from(has_event_idx);
```
cagatay-y, line 186: Should this subtraction be done? If `avail_event` exists, it is part of the slice.

cagatay-y: I think we should include a check similar to the one above to see if this `len` is a multiple of (`size_of::<UsedElem> / size_of::<le16>`).

virtio-spec/src/virtq/alloc.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    pub fn try_new_in<A: Allocator>(
+        queue_size: u16,
+        has_event_idx: bool,
+        alloc: A,
+    ) -> Result<Box<Self, A>, AllocError> {
+        let layout = Self::layout(queue_size, has_event_idx);
+
+        let mem = alloc.allocate_zeroed(layout)?;
+        let mem = NonNull::slice_from_raw_parts(mem.cast(), layout.size());
```
cagatay-y, line 63: Is `Used` padded at the end? If I understand correctly, this line will provide a slice with the padding included and I think the code in mod.rs is written with the assumption that the slice does not include any padding.

virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        let ptr = ptr::slice_from_raw_parts(ptr, len) as *mut Used;
+        Some(NonNull::new(ptr).unwrap())
+    }
+
+    pub fn ring_ptr(this: NonNull<Self>) -> NonNull<[UsedElem]> {
+        let ptr = unsafe { addr_of_mut!((*this.as_ptr()).ring_and_avail_event) };
+        let len = ptr.len() * mem::size_of::<le16>() / mem::size_of::<UsedElem>();
```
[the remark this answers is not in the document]
cagatay-y: Also for `avail_event_ptr`

virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        #[doc(alias = "VIRTQ_DESC_F_USED")]
+        const USED = 1 << 15;
+    }
+}
+
+/// The Virtqueue Available Ring
+#[doc(alias = "virtq_avail")]
+#[derive(Debug)]
+#[repr(C)]
+pub struct Avail {
```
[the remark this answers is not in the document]
mkroening: While it might be nicer, this crate will always use the naming from the spec: `virtq_avail` -> `virtq::Avail`.

It can, of course, always be imported with a different name, but for this PR, I'd like to stick to the spec in the kernel as well.

virtio-spec/src/virtq/mod.rs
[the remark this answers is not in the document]
mkroening: I don't think so, since `size_of_val_raw` is unstable.

virtio-spec/src/virtq/mod.rs
mkroening: > Should this subtraction be done? If `avail_event` exists, it is part of the slice.

Yes, this is subtracting the padding.

> I think we should include a check similar to the one above to see if this `len` is a multiple of (`size_of::<UsedElem> / size_of::<le16>`).

I don't understand. The slice contains `avail_event` after all.

virtio-spec/src/virtq/alloc.rs
mkroening: This is padded, since `layout` is padded. I think `Used::from_ptr` still works fine?

virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        let len = len / mem::size_of::<le16>() - 2;
+        let ptr = ptr::slice_from_raw_parts_mut(ptr, len) as *mut Self;
+        Some(NonNull::new(ptr).unwrap())
+    }
+
+    pub fn ring_ptr(this: NonNull<Self>, has_event_idx: bool) -> NonNull<[le16]> {
+        let ptr = unsafe { addr_of_mut!((*this.as_ptr()).ring_and_used_event) };
+        let len = ptr.len().saturating_sub(usize::from(has_event_idx));
```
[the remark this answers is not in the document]
mkroening: Done 👍

virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+
+impl Used {
+    pub fn layout(queue_size: u16, has_event_idx: bool) -> Layout {
+        let mut layout = Layout::array::<le16>(2)
+            .unwrap()
+            .extend(Layout::array::<UsedElem>(queue_size.into()).unwrap())
+            .unwrap()
+            .0;
+
+        if has_event_idx {
```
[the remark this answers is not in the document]
mkroening: Done 👍

virtio-spec/src/virtq/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+            layout = layout
+                .extend(Layout::new::<le16>())
+                .unwrap()
+                .0
+                .pad_to_align();
```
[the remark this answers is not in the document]
mkroening: Done 👍

virtio-spec/src/virtq/mod.rs
cagatay-y: I meant something like
```rust
if (len - usize::from(has_event_idx)) % (mem::size_of::<UsedElem>() / mem::size_of::<le16>()) != 0 {
    return None;
}
```
to make sure the part of the slice that is supposed to be `[UsedElem]` actually has a size that is a multiple of `mem::size_of::<UsedElem>()`.

virtio-spec/src/virtq/alloc.rs
cagatay-y: You are right

cagatay-y (approved):

virtio-spec/src/virtq/mod.rs
mkroening: I have replaced the previous check with

```rust
if len % mem::size_of::<UsedElem>() != usize::from(!has_event_idx) * mem::size_of::<le32>() {
    return None;
}
```

PR #1314, opened 2024-07-09
fix(virtqueue/packed): improve flag handling

cagatay-y:


mkroening (approved):
Looks good to me, thanks! :)

PR #1318, opened 2024-07-15
add lseek support for in-memory files

stlankes:
In addition, the synchronous `lseek` interface is converted asynchronous interface. Furthermore, if the feature `mmap` is available, the heap is reduced because it will expected that `mmap` will be used to map additional pages into the user address space.


mkroening (approved):
Looks good to me! 👍

PR #1328, opened 2024-07-22
feat(newlib): add support for riscv64

johnyob:
This PR simply defines `kernel_heap_end()` as `KERNEL_VIRTUAL_MEMORY_END` for the riscv64 architecture. 
This function is required for building the hermit kernel for riscv64 with the `newlib` feature flag enabled.


mkroening (approved):
Thanks! :)

PR #1329, opened 2024-07-22
fix(virtq/split): return id with the correct index for the chain head

### LINKED ISSUE #1327: The buff_id is incorrect or the reference to the TransferToken was misplaced.
If I test hermit-os/hermit-rs#602 with a TAP device, I received an internal virtio error.

I created the tap device as followed

```shell
 sudo ip tuntap add tap10 mode tap
 sudo ip addr add 10.0.5.1/24 broadcast 10.0.5.255 dev tap10
 sudo ip link set dev tap10 up
``` 

Afterwards I revised the test script from hermit-os/hermit-rs#602 as followed:

```shell
sleep 10 && cargo run --manifest-path ./Cargo.toml --bin client-bw --release --target x86_64-unknown-linux-gnu -- --nonblocking --address 10.0.5.3 --bytes 6000 --rounds 1000 & sudo /usr/libexec/qemu-kvm -display none -serial stdio -kernel hermit-loader-x86_64 -cpu host -device isa-debug-exit,iobase=0xf4,iosize=0x04 -initrd target/x86_64-unknown-hermit/release/server-bw -smp 1 -m 1024M -netdev tap,id=net0,ifname=tap10,script=no,downscript=no,vhost=on -device virtio-net-pci,netdev=net0,disable-legacy=on -append '-- --address 10.0.5.3 --bytes 6000 --rounds 1000'
``` 

I received following error message:

```shell
[0][PANIC] panicked at src/drivers/virtio/virtqueue/split.rs:174:18:
The buff_id is incorrect or the reference to the TransferToken was misplaced.
```

cagatay-y:
The incorrect index was used to iterate over the descriptor chain. Fixes #1327.

### DIFF
--- a/src/drivers/virtio/virtqueue/split.rs
+++ b/src/drivers/virtio/virtqueue/split.rs
@@ -159,12 +159,11 @@ impl DescrRing {
 		// need to move [Self::used_ring_ref] lines into a separate scope.
 		loop {
 			let used_elem;
-			let cur_ring_index;
 			{
 				if self.read_idx == self.used_ring().idx.to_ne() {
 					break;
 				} else {
-					cur_ring_index = self.read_idx as usize % self.token_ring.len();
+					let cur_ring_index = self.read_idx as usize % self.token_ring.len();
 					used_elem = self.used_ring().ring()[cur_ring_index];
 				}
 			}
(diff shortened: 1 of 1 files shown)


mkroening (approved):
Thanks for the swift fix!

PR #1344, opened 2024-07-31
refactor(virtq): eliminate RefCell and Rc usage

cagatay-y:
Since we no longer store a reference to the queue inside the transfer structures, we don't need to wrap the queues inside `Rc`s and the queue fields inside `RefCell`s.


mkroening (approved):
Wonderful, seeing this come together! Thanks!

PR #1345, opened 2024-07-31
refactor(virtio-net): remove unused is_multi field

cagatay-y:


mkroening (approved):
I think this was meant for multiqueue support. But since have never actually supported that, we might as well start with a clean slate. 👍

PR #1378, opened 2024-09-03
remove spinlock in interrupt handlers

stlankes:


src/drivers/mod.rs
```diff
@@ -112,6 +112,9 @@ pub mod error {
 pub(crate) trait Driver {
 	/// Returns the interrupt number of the device
 	fn get_interrupt_number(&self) -> InterruptLine;
+
+	/// Returns the device driver name
+	fn get_name(&self) -> &'static str;
```
mkroening, line 117: This method is never called currently, right?

mkroening (changes requested):
This looks good, but I am confused about one thing.

src/drivers/mod.rs
mkroening: But it might be used in the future to get rid of the additional map for names entirely.

mkroening (approved):
Looks good to me

PR #1390, opened 2024-09-17
refactor(virtq): remove intermediate virtqueue channels

cagatay-y:
Rather than pushing used buffers from the queues to channels, return them directly to the drivers. This removes the dependency on `async-channel`.


mkroening (approved):
Thanks!

PR #1391, opened 2024-09-19
Enhanced uhyve memory detection

jounathaen:
The current memory detection relies on some moreless undocumented assumptions about how uhyve prepares the memory. However, we do get all the information in the boot-info to determine the memory correctly.

This is one step in enabling "ASLR-like" arbitrary memory locations for hermit (https://github.com/hermit-os/uhyve/pull/725)


mkroening:
Note that https://github.com/hermit-os/uhyve/pull/758/commits/1a83fa5b3efae091f594b6a79b77f71acaf84633 would make `detect_from_uhyve` inactive.

mkroening (changes requested):
`detect_from_uhyve` is inactive since https://github.com/hermit-os/uhyve/pull/758.

jounathaen:
Despite this being not relevant for new Uhyve versions, I think this code still contains some improvements to the memory detection code. Is it worth rebasing this?

PR #1394, opened 2024-09-20
feat(env): parse env=KEY=VALUE for env vars

mkroening:
This PR

1. unifies FDT bootargs parsing across all platforms,
2. adds `env=KEY=VALUE` kernel arguments for environment variable initialization, and
3. disables the Uhyve `CMDSIZE` and `CMDVAL` hypercalls to use the kernel-maintained environment on Uhyve (breaking).

## Questions

1. Should we use the proposed `env=KEY=VALUE` format? This differs from our current format for other kernel arguments, such as `-freq FREQ` and `-ip IP`. I think we should migrate those to `freq=FREQ` and `ip=IP` eventually.
2. Do we need to support `=` in environment variable names?
3. Should we immediately disable the Uhyve hypercalls (easy)? This would break older versions of Uhyve, but I don't think this is a big problem. Alternatively, we could merge the Uhyve hypercall environment with the kernel one (complicated), or we only use the kernel environment when Uhyve returns an empty one (medium).

Extracted 0. commit: https://github.com/hermit-os/kernel/pull/1396
Corresponding Uhyve PR: https://github.com/hermit-os/uhyve/pull/758

### DIFF
--- a/src/arch/aarch64/kernel/mod.rs
+++ b/src/arch/aarch64/kernel/mod.rs
@@ -91,15 +91,7 @@ pub fn get_processor_count() -> u32 {
 }
 
 pub fn args() -> Option<&'static str> {
-	let dtb = unsafe {
-		hermit_dtb::Dtb::from_raw(ptr::with_exposed_provenance(
-			boot_info().hardware_info.device_tree.unwrap().get() as usize,
-		))
-		.expect(".dtb file has invalid header")
-	};
-
-	dtb.get_property("/chosen", "bootargs")
-		.map(|property| str::from_utf8(property).unwrap())
+	None
 }
 
 /// Earliest initialization function called by the Boot Processor.
--- a/src/env.rs
+++ b/src/env.rs
@@ -2,9 +2,10 @@
 
 use alloc::string::{String, ToString};
 use alloc::vec::Vec;
-use core::str;
+use core::{ptr, str};
 
 use ahash::RandomState;
+use fdt::Fdt;
 use hashbrown::hash_map::Iter;
 use hashbrown::HashMap;
 use hermit_entry::boot_info::PlatformInfo;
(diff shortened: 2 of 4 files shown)


src/env.rs
```diff
@@ -36,6 +37,17 @@ pub fn is_uhyve() -> bool {
...
 }
 
+pub fn fdt() -> Option<Fdt<'static>> {
+	kernel::boot_info().hardware_info.device_tree.map(|fdt| {
+		let ptr = ptr::with_exposed_provenance(fdt.get().try_into().unwrap());
+		unsafe { Fdt::from_ptr(ptr).unwrap() }
+	})
+}
+
+pub fn fdt_args() -> Option<&'static str> {
```
jounathaen, line 47: I like brief doc comments on functions, but this is a personal preference.

Shouldn't these functions be `pub (crate)`?

src/arch/aarch64/kernel/mod.rs
```diff
@@ -91,15 +91,7 @@ pub fn get_processor_count() -> u32 {
...
-	let dtb = unsafe {
-		hermit_dtb::Dtb::from_raw(ptr::with_exposed_provenance(
-			boot_info().hardware_info.device_tree.unwrap().get() as usize,
-		))
-		.expect(".dtb file has invalid header")
-	};
-
-	dtb.get_property("/chosen", "bootargs")
-		.map(|property| str::from_utf8(property).unwrap())
+	None
```
jounathaen, line 94: Couldn't we throw out these functions while we are on it?

jounathaen:
LGTM

>     Should we use the proposed env=KEY=VALUE format? This differs from our current format for other kernel arguments, such as -freq FREQ and -ip IP. I think we should migrate those to freq=FREQ and ip=IP eventually.

Sounds reasonable.

>     Do we need to support = in environment variables? That would require some tuning of shell_words, I think.

I think this is a must. It is exotic, but would lead to some strange bugs if there is just some environment that uses it.

>     Should we immediately disable the Uhyve hypercalls (easy)? This would break older versions of Uhyve, but I don't think this is a big problem. Alternatively, we could merge the Uhyve hypercall environment with the kernel one (complicated), or we only use the kernel environment when Uhyve returns an empty one (medium).

Just throw them out. On my V2 of the uhyve interface, these are also not present anymore. I doubt that uhyve is used "in production" right now.

src/env.rs
mkroening: > Shouldn't these functions be `pub (crate)`?

I prefer not setting the export scope when the whole module is private.

src/arch/aarch64/kernel/mod.rs
mkroening: We can do away with these architecture-specific bootargs once we have migrated all platforms to FDT bootargs and do the bootinfo version jump. We still have the bootargs in the bootinfo for x86-64 QEMU and Linux boot.

mkroening:
> > Do we need to support = in environment variables? That would require some tuning of shell_words, I think.
> 
> I think this is a must. It is exotic, but would lead to some strange bugs if there is just some environment that uses it.

I have implemented support for `=` in environment variable values, but not in names. This is fine according to [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html).

> Just throw them out.

👍

PR #1397, opened 2024-09-21
ci: run uhyve only with release profile kernels

mkroening:


mkroening:
This is obsolete, since https://github.com/hermit-os/uhyve/issues/790 is solved. 🥳

PR #1398, opened 2024-09-21
add option to increase queue limit for incoming connections

stlankes:
Revise object interface to be a complete asynchronous interface. In addition, add support of listen to increase queue limit for incoming connections.

smoltcp-rs/smoltcp#852 describes if N sockets are listening on the same port, an incoming connections go to a random socket. Consequently, this PR checks all listening sockets to accept a connection.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -397,12 +397,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dyn-clone"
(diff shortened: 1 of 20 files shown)


mkroening (approved):
This works on Linux! 🥳 
It's not fast, but it works.

Thanks a lot! :)

PR #1402, opened 2024-09-25
virtio-pci: store only the most preferred capability of a type

cagatay-y:
According to the section 4.1.4 of the VIRTIO specification v1.2, "[t]he order of the capabilities in the capability list specifies the order of preference suggested by the device." Accordingly, store only the most preferred capability of a type in the `UniCapsCollection` structure when only one is necessary.

This also allows the corresponding fields to be changed from vectors to singles (when appropriate) and the existence of the mandatory capabilities to be checked early on.


mkroening (approved):
Thanks!

PR #1442, opened 2024-10-31
Use `memory_address::{VirtAddr, PhysAddr}` in the kernel

jounathaen:
This depends on https://github.com/hermit-os/rust-aarch64/pull/9


jounathaen:
Would probably be an interesting case for continuous benchmarking...

mkroening (commented):
Memory addresses should ideally be formatted as `{:p}` ([`std::fmt::Pointer`](https://doc.rust-lang.org/stable/std/fmt/trait.Pointer.html)).

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -264,7 +265,7 @@ trait VirtqPrivate {
 				.chain(recv_desc_iter)
 				.map(|(mem_descr, len, incomplete_flags)| {
 					Self::Descriptor::incomplete_desc(
-						paging::virt_to_phys(VirtAddr::from(mem_descr.addr() as u64))
```
cagatay-y, line 267: We still need the `BufferElem::addr` call here. It provides a pointer to the first byte of the struct that is contained by the `BufferElem` enum, which is what the VIRTIO device was interested in. Without it, the device gets a pointer to the wrapper enum and corrupts the variant information. That was what caused the pop function to fail: It expected a `Sized` variant but the corrupted byte told it that it was a `Vector`. It was a confusing choice of name in my part, we may want to change into something like `ptr_to_inner`.

cagatay-y: With this change, the failing test case passes on my machine.

jounathaen: Thanks! That was it! :heart:

mkroening (approved):
Looks good to me! :)

PR #1467, opened 2024-11-25
feat: add stable exit code message in non-Uhyve VMs

mkroening:


jounathaen:
Hmm, now the exit code print can't be suppressed anymore, which [contradicts the Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) a little (and wouldn't it be cool, if the execution of an application in Hermit won't be distinguishable from a host execution at some day :slightly_smiling_face: ). But it is in line with the statistic printing. :thinking: 

Maybe both can be switched with an argument/boot flag instead?

src/syscalls/interfaces/mod.rs
```diff
@@ -49,6 +49,9 @@ pub trait SyscallInterface: Send + Sync {
 	}
 
 	fn shutdown(&self, error_code: i32) -> ! {
+		// This is a stable message used for detecting exit codes for different hypervisors.
+		println!("exit status {error_code}");
```
jounathaen, line 53: One could debate, whether `exit_status={error_code}` might be easier to parse. But I'm fine with it as is.

jounathaen (approved):

stlankes (approved):

PR #1393, opened 2024-09-20
Use hypercall definitions from uhyve-interface

jounathaen:
This depends on https://github.com/hermit-os/uhyve/pull/757 (and a publication of uhyve interface V0.1.2)

This is the final step of putting the hypercall definitions in a central crate.
This PR should just deduplicate code without changing functionality.

PR #1468, opened 2024-11-25
Uhyve: Speed up serial output

jounathaen:
- Uhyve: Use the `SerialBufferWrite` hypercall instead of the `Uart` hypercall: This reduces the number of serial output related hypercalls by ~3x.

Attention:
- Depends on https://github.com/hermit-os/kernel/pull/1393
- Depends on a new Uhyve release that has support for this hypercall (v0.4.0)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -626,6 +626,7 @@ dependencies = [
  "free-list",
  "fuse-abi",
  "hashbrown",
+ "heapless",
  "hermit-dtb",
  "hermit-entry",
  "hermit-macro",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -114,6 +114,7 @@ fdt = { version = "0.1", features = ["pretty-printing"] }
 free-list = "0.3"
 fuse-abi = { version = "0.2", features = ["linux", "zerocopy"], optional = true }
 hashbrown = { version = "0.15", default-features = false }
+heapless = "0.8"
 hermit-entry = { version = "0.10", features = ["kernel"] }
 hermit-sync = "0.1"
 lock_api = "0.4"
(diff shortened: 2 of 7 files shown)


mkroening (approved):
I don't feel good about including `heapless` since we have come so far without it, but I can't think of anything better. 👍 

Regarding the general approach, it would be a lot nicer to have a proper `io::BufWriter`, but `core::io` is too far away still. (The most up-to-date approach I could find is [wcampbell0x2a/no-std-io2](https://github.com/wcampbell0x2a/no-std-io2), and even that has issues.

So leaving `io::BufWriter` aside, it might still be worthwhile to have this buffering strategy take place outside `arch` such that all serial ports can take advantage of it. For those that can't, the overhead will probably not be too bad, I think. Avoiding the `\n` check here by extracting this into external `flush` calls would also be nice sometime in the future when we have nice `core::io` abstractions.

jounathaen:
I agree with all your points.  Especially, the unification of some IO functions across the architectures is desirable. But I'd still propose to merge this now and enhance later on.

mkroening:
> But I'd still propose to merge this now and enhance later on.

Yeah, sure, but this is blocked on https://github.com/hermit-os/kernel/pull/1393.

mkroening (changes requested):
Could you rebase this, since https://github.com/hermit-os/kernel/pull/1393 is merged? :)

mkroening (changes requested):
Would it make sense to move this to the arch-independent part `Console` ([src/console.rs#L15-L17](https://github.com/hermit-os/kernel/blob/b2eef95abb4012cde30a8e1c47e5c67b3ad053e9/src/console.rs#L15-L17))?

That would drastically reduce the code duplication, make the code simpler, and avoid `clippy::large_enum_variant`. The additional overhead of doing an extra copy for byte-wise serial devices is probably negligible, I'd expect.

mkroening (changes requested):
The output of the second commit is wrong. I'll look for the new line issue tomorrow.

mkroening (changes requested):

mkroening (approved):

PR #1482, opened 2024-11-30
refactor(virtio-net): avoid placefolder values in driver construction

cagatay-y:


src/drivers/net/virtio/mod.rs
```diff
@@ -471,20 +528,35 @@ impl VirtioNetDriver {
...
+	) -> Result<VirtioNetDriver, VirtioNetError> {
+		let ctrl_vq;
+		let recv_vqs;
+		let send_vqs;
+		let mtu;
+		let mut checksums;
+
+		// We do the whole initialization process inside an inner scope so that if any of the steps fail,
+		// the scope returns an `Err` that we can catch to set the driver status as failed.
+		{
```
mkroening, line 546: This does not work. Returning from a scope returns the whole function immediately without inspecting the result. We'd have to use a closure or proper function, since try blocks are not there yet.

mkroening (changes requested):

Issue #1483, opened 2024-12-01
Uhyve: Implement `chdir`


n0toose:
Related to https://github.com/hermit-os/uhyve/issues/815.

jounathaen:
@n0toose is this still relevant? If so, could you add a brief description the issue is?

jounathaen:
I don't think this is an issue. Hermit should access all Uhyve files with a full path. If it isn't, then we should create a specific issue for that.

PR #1477, opened 2024-11-27
feat(uhyve): change mountpoint env var

n0toose:
Related to https://github.com/hermit-os/uhyve/pull/783 and its fierce debate regarding the naming of the `--mount` variable.

PR #1485, opened 2024-12-01
feat(hypercall): introduce chdir and uhyve mountpoint restrictions

n0toose:
Under heavy construction.

Related to: https://github.com/hermit-os/uhyve/pull/820
Fixes: https://github.com/hermit-os/kernel/issues/1483
Succeeds: https://github.com/hermit-os/kernel/pull/1477

### DIFF
--- a/src/fs/uhyve.rs
+++ b/src/fs/uhyve.rs
@@ -321,7 +321,7 @@ impl VfsNode for UhyveDirectory {
 pub(crate) fn init() {
 	info!("Try to initialize uhyve filesystem");
 	if is_uhyve() {
-		let mount_point = hermit_var_or!("UHYVE_MOUNT", "/root").to_string();
+		let mount_point = hermit_var_or!("UHYVE_MOUNTPOINT", "/root").to_string();
 		info!("Mounting uhyve filesystem at {}", mount_point);
 		fs::FILESYSTEM
 			.get()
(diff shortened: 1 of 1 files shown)


jounathaen:
IMHO, we should merge https://github.com/hermit-os/kernel/pull/1393 first and put the definitions of the hypercalls directly in uhyve-interface.

mkroening:
Why would we need a `chdir` hypercall? The current working directory is a property of the currently running process, not of the operating system.

I think Hermit should manage the current working directory and only expose canonicalized paths to Uhyve (or virtiofsd via QEMU).

jounathaen:
Well, it is a property, but not an immutable one. [POSIX defines calls to change it during a processes runtime](https://linux.die.net/man/3/chdir). Also, [Rust std has support for it](https://doc.rust-lang.org/std/env/fn.set_current_dir.html).
Sure, it is a rare use-case, but whilst @n0toose is working on file-systems anyway, why not increase the compatibility.
The only question is, if this should be a hypercall, or something that Hermit keeps track of in the VM and resolves it before the hypercall. I guess I'm leaning towards the second approach.

n0toose:
Would a `get_mount_point()` like in the FUSE driver be a better / less controversial approach for now instead of a new `chdir` hypercall? It would serve the same purpose.

https://github.com/hermit-os/kernel/blob/fcb871c4ecf1490a5a9f02efa13224c668436bbd/src/fs/fuse.rs#L1156

mkroening:
> Well, it is a property, but not an immutable one. [POSIX defines calls to change it during a processes runtime](https://linux.die.net/man/3/chdir). Also, [Rust std has support for it](https://doc.rust-lang.org/std/env/fn.set_current_dir.html). Sure, it is a rare use-case, but whilst @n0toose is working on file-systems anyway, why not increase the compatibility. The only question is, if this should be a hypercall, or something that Hermit keeps track of in the VM and resolves it before the hypercall. I guess I'm leaning towards the second approach.

That's precisely what I was trying to say. :)

PR #1488, opened 2024-12-02
fix(virtio-net): prepare checksum correctly

cagatay-y:
Introduces a nightly feature (`iter_array_chunks`). I didn't avoid it as upstreamability was not a focus.


cagatay-y:
> Introduces a nightly feature (`iter_array_chunks`). I didn't avoid it as upstreamability was not a focus.

I rewrote the relevant part to not use the unstable feature.

cagatay-y:
I should note that the adapted implementation was not tested on grobi as it is offline but shouldn't be substantially different from the one that was previously tested with it.

src/drivers/net/virtio/mod.rs
```diff
@@ -488,6 +462,90 @@ impl VirtioNetDriver<Init> {
...
 		self.inner.recv_vqs.enable_notifs();
 	}
+
+	/// If necessary, sets the TCP or UDP checksum field to the checksum of the
+	/// pseudo-header and returns the IP header length and the checksum offset.
+	/// Otherwise, returns None.
+	fn should_request_checksum<T: AsRef<[u8]> + AsMut<[u8]>>(
+		&self,
+		frame: T,
+	) -> Option<(u16, u16)> {
```
mkroening, line 474: Can we use early returns here, please?

mkroening (changes requested):
Thanks! :)

Can you rebase this and undo/revert 437bb484304b9361db0b52e2c6df03a377111dd4?

mkroening (approved):
Thanks! 🥳

PR #1501, opened 2024-12-06
fix(x86_64): don't set `CR0Flags::TASK_SWITCHED`

sarahspberrypi:
As previously mentioned, the manual setting of this flag seems to be unnecessary, so I propose to remove it from the code.

### DIFF
--- a/src/arch/x86_64/kernel/processor.rs
+++ b/src/arch/x86_64/kernel/processor.rs
@@ -800,9 +800,6 @@ pub fn configure() {
 	cr0.insert(Cr0::CR0_MONITOR_COPROCESSOR | Cr0::CR0_NUMERIC_ERROR);
 	cr0.remove(Cr0::CR0_EMULATE_COPROCESSOR);
 
-	// if set, the first FPU access will trigger interrupt 7.
-	cr0.insert(Cr0::CR0_TASK_SWITCHED);
-
 	// Prevent writes to read-only pages in Ring 0.
 	cr0.insert(Cr0::CR0_WRITE_PROTECT);
 


mkroening (changes requested):
I have done some digging on why this was added:

https://github.com/hermit-os/kernel/blob/77bd5d1dda86df97dbbf1db5b8697b053d10b606/librs/src/arch/x86_64/processor.rs#L489-L490

https://github.com/hermit-os/kernel/blob/77bd5d1dda86df97dbbf1db5b8697b053d10b606/librs/src/arch/x86_64/irq.rs#L211-L218

If the processor executes a floating-point instruction while the TS flag in control register CR0 was set and the EM flag is clear, a device-not-available exception is generated.
I understand this was added intentionally to ensure the behavior for the initial task is the same as for subsequent tasks that were switched to and have set the TS flag implicitly.

I have not done enough digging to say whether it would be fine to remove this, but maybe you don't have a problem with the TS flag but with running interrupts in general.

stlankes:
We should accept the PR. I have the same opinion like @sarahspberrypi

stlankes (approved):

stlankes:
@sarahspberrypi I talked to @mkroening We would like to set TASK_SWITCHED flag. Do you really need to remove the flag?

sarahspberrypi:
> @sarahspberrypi I talked to @mkroening We would like to set TASK_SWITCHED flag. Do you really need to remove the flag?

The flag is set later during `switch_to_task()` call anyways (see [here](https://github.com/hermit-os/kernel/blob/461455cbd5a08e0c61030953c5fbae98b69ade8d/src/arch/x86_64/kernel/switch.rs#L171) ), so I don't really see why we need to have it pre-enabled? 
With SEV enabled, this apparently triggers an immediate hardware interrupt that cannot be masked. This interrupt can only be handled by hermit later when the IDT is loaded. But I'll talk with @mkroening in the new year about this.

PR #1503, opened 2024-12-08
fix(uhyve): fix improper unlink string handling

n0toose:
Taken from `traverse_open`.

### DIFF
--- a/src/fs/uhyve.rs
+++ b/src/fs/uhyve.rs
@@ -286,13 +286,16 @@ impl VfsNode for UhyveDirectory {
 
 	fn traverse_unlink(&self, components: &mut Vec<&str>) -> io::Result<()> {
 		let path: String = if components.is_empty() {
-			"/".to_string()
+			"/\0".to_string()
 		} else {
-			components
+			let mut path: String = components
 				.iter()
 				.rev()
 				.map(|v| "/".to_owned() + v)
-				.collect()
+				.collect();
+			path.push('\0');
+			path.remove(0);
+			path
 		};
 
 		let mut sysunlink = SysUnlink::new(VirtAddr::from_ptr(path.as_ptr()));


n0toose:
I caught this as I was working on benchmarks for Uhyve. Using a modified test of Uhyve, this seems to be working. There was an error when converting a char pointer to an str because it was not in UTF-8 - which is a check that I built into Uhyve to prevent weird behavior like this. This means that `fs::remove_file` has been broken on Uhyve for a while now - but I did not catch that, as I tested the call on my host system without a `UhyveVm` under the assumption that it previously did.

Anyways, this is not the case anymore. CC: @jounathaen 

image

n0toose:
I caught this as I was working on benchmarks for Uhyve. Using a modified test of Uhyve, this seems to be working. There was an error when converting a char pointer to an str because it was not in UTF-8 - which is a check that I built into Uhyve to prevent weird behavior like this. This means that `fs::remove_file` has been broken on Uhyve for a while now - but I did not catch that, as I tested the call on my host system without a `UhyveVm` under the assumption that it previously did.

Anyways, this is not the case anymore. CC: @jounathaen 

```
[0][INFO] Try to initialize uhyve filesystem
[0][INFO] Mounting uhyve filesystem at /root
[0][INFO] Jumping into application
Hello from create_file!
Successfully removed file!
Number of interrupts
[0][FPU]: 1
```

n0toose:
Here's the code to reproduce, @mkroening:

```rs
use std::{fs::File, io::prelude::*, fs::remove_file};

#[cfg(target_os = "hermit")]
use hermit as _;

fn main() {
	println!("Hello from create_file!");

	let mut file = File::create("/root/foo.txt").unwrap();
	file.write_all(b"Hello, world!").unwrap();
        // this is what I added to the test myself
	remove_file("/root/foo.txt").unwrap();
	println!("Successfully removed file!");
}
```

Seems like this "bug" was caused by the following commit: https://github.com/hermit-os/kernel/commit/ba9f92cc9186c7b4790e90b1aa89ca1c8302e4e5

... which took some extra steps to ensure that null terminators were handled properly in `traverse_open` but were perhaps forgotten about in `traverse_unlink`, which was unused at the time. The `unlink` call at the time accepted any path on the host OS and was / is not 'mount-aware' the same way `open` is.

mkroening (approved):
Thanks! And sorry for the wait! :)

PR #1508, opened 2024-12-10
uhyve: don't panic on initial unmap if page is not mapped

jounathaen:
I'm reworking the initial page mapping of uhyve to resolve https://github.com/hermit-os/uhyve/issues/426, but the kernel currently expects the first GiB to be mapped and panics otherwise. 
In a transition phase, it makes sense to at least not panic if the page is not mapped.

### DIFF
--- a/src/arch/x86_64/mm/paging.rs
+++ b/src/arch/x86_64/mm/paging.rs
@@ -387,8 +387,11 @@ pub fn init_page_tables() {
 
 		let mut page_table = unsafe { recursive_page_table() };
 		for page in page_range {
-			let (_frame, flush) = page_table.unmap(page).unwrap();
-			flush.ignore();
+			match page_table.unmap(page) {
+				Ok((_frame, flush)) => flush.ignore(),
+				Err(UnmapError::PageNotMapped) => {} // If it wasn't mapped, that's not an issue
+				Err(e) => panic!("Couldn't unmap page {page:?}: {e:?}"),
+			}
 		}
 
 		tlb::flush_all();


src/arch/x86_64/mm/paging.rs
```diff
@@ -387,8 +387,11 @@ pub fn init_page_tables() {
 
 		let mut page_table = unsafe { recursive_page_table() };
 		for page in page_range {
-			let (_frame, flush) = page_table.unmap(page).unwrap();
-			flush.ignore();
+			match page_table.unmap(page) {
+				Ok((_frame, flush)) => flush.ignore(),
+				Err(UnmapError::PageNotMapped) => {} // If it wasn't mapped, that's not an issue
```
mkroening, line 392: This is similar to what we do in [`paging::unmap`](https://github.com/hermit-os/kernel/blob/v0.8.0/src/arch/x86_64/mm/paging.rs#L255-L259). Would it make sense to log this too?

jounathaen: I don't think so unless you like ~510 lines of `info` log output on a normal boot.

mkroening: We would not have to use `info`. But if you don't think a `debug` or `trace` is useful, that's fine for me.

jounathaen: Honestly, I don't think it is worth it.

mkroening: Should we add a `TODO` into the comment or will this be tracked in an issue?

jounathaen: I think it is sufficient not to panic for a while, and as soon as the Uhyve version that doesn't map the memory anymore (https://github.com/hermit-os/uhyve/pull/725/commits/700425d2003d4baec74aaeb42e79cf7fb786d3c9 btw) we completely remove this part. We can track it in https://github.com/hermit-os/uhyve/issues/426 or create a new issue.

mkroening (approved):

PR #1512, opened 2024-12-13
implement fstat for fuse file handles via getattr

m-mueller678:
Fixes hermit-os/kernel#1516


mkroening (approved):
Looks good to me, thanks! :)

PR #1521, opened 2024-12-17
avoid unnecessary dereference in futex_wake

m-mueller678:
In the `futex_wake` syscall, the address of the passed pointer is used to index a hash map. This pull request removes an unnecessary roundtrip through `&AtomicU32` and thereby make it safe to call `futex_wake` with a dangling pointer. This change enables using `futex_wake` in `parking_lot_core` without additional locking.


mkroening (approved):
Thanks!

PR #1538, opened 2025-01-03
refactor(stdio): migrate to `uhyve-interface`

mkroening:
This was probably forgotten in https://github.com/hermit-os/kernel/pull/1393.


mkroening:
CC: @jounathaen

PR #1558, opened 2025-01-16
remove unsafe code regions

stlankes:
By using slices instead of raw pointers, the number of unsafe code regions can be reduced.


mkroening (approved):
Looks good to me! 👍

PR #1560, opened 2025-01-17
feat: add version note

mkroening:


src/note.rs
```diff
@@ -0,0 +1,48 @@
...
+	() => {
+		#[used]
+		#[unsafe(link_section = ".note.ABI-tag")]
+		static ABI_TAG: $crate::note::_AbiTag = $crate::note::_AbiTag::new();
+	};
+}
+
+#[repr(C)]
+#[doc(hidden)]
+pub struct _AbiTag {
```
jounathaen, line 18: It's fine as is, but wouldn't something like `_hermit_version` or `_hermit_version_tag` be a little easier to comprehend?

src/note.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl _AbiTag {
+	pub const fn new() -> Self {
+		Self {
+			header: Nhdr32 {
+				n_namesz: 7,
+				n_descsz: 1,
+				n_type: 1,
+			},
+			name: *b"HERMIT\0\0",
```
jounathaen, line 32: if `n_namesz` is `7`, why the double zero termination?

src/note.rs
```diff
@@ -0,0 +1,48 @@
...
+	data: [u32; 4],
+}
+
+impl _AbiTag {
+	pub const fn new() -> Self {
+		Self {
+			header: Nhdr32 {
+				n_namesz: 7,
+				n_descsz: 1,
+				n_type: 1,
```
jounathaen, line 30: I'm not too deep into elf headers and even if this field does matter on a note header, but for the hermit-entry version field is set to `0x5a00`.

mkroening:
I moved the code and the discussion over to https://github.com/hermit-os/hermit-entry/pull/45. This code is complete, but will only compile once there is a new hermit-entry release.

PR #1571, opened 2025-01-27
feat(processor): read TSC frequency from FDT

mkroening:
See https://github.com/hermit-os/uhyve/pull/876.

### DIFF
--- a/src/arch/x86_64/kernel/processor.rs
+++ b/src/arch/x86_64/kernel/processor.rs
@@ -5,7 +5,7 @@ use core::arch::x86_64::{
 	__rdtscp, _fxrstor, _fxsave, _mm_lfence, _rdseed64_step, _rdtsc, _xrstor, _xsave,
 };
 use core::hint::spin_loop;
-use core::num::NonZeroU32;
+use core::num::{NonZero, NonZeroU32};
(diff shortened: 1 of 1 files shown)


src/arch/x86_64/kernel/processor.rs
```diff
@@ -388,6 +390,23 @@ impl CpuFrequency {
...
+				.as_usize()?;
+			let khz = u32::try_from(khz).ok()?;
+			let mhz = u16::try_from(khz / 1000).ok()?;
+			NonZero::new(mhz)
+		}
+
+		let mhz = mhz_from_fdt().ok_or(())?;
+		self.set_detected_cpu_frequency(mhz.get(), CpuFrequencySources::Fdt)?;
+
+		Ok(())
```
jounathaen, line 407: The other functions in this file are written like
```suggestion
		return self.set_detected_cpu_frequency(mhz.get(), CpuFrequencySources::Fdt);
```

jounathaen (approved):

src/arch/x86_64/kernel/processor.rs
mkroening: Yes, but I consider returning a result instead of handling it and repackaging it to be bad style. :D

I just did not want to trick myself into rewriting this today.

PR #1585, opened 2025-02-06
exclude CITATION.cff

stlankes:


mkroening (commented):
The `exclude` field does not have an effect, since we don't publish to crates.io, but use the GitHub source tarball.

And even if it would matter, I think including it does not hurt, similar to also including licenses.

jounathaen (commented):
I'm with Martin in this case. Why not publishing this information as well?

stlankes:
It was just a request. For me wasn't clear if the file required in source tarball.

PR #1606, opened 2025-02-19
Use uninitialized buffers for `read` and `recvfrom`

thaliaarchi:
Currently, when using an uninitialized buffer for `read` or `recvfrom`, as would be typical in C or required for `Read::read_buf`, it is casted from `*mut u8` at the FFI boundary in `sys_read`/`sys_readv`/`sys_recvfrom` to a `&[u8]`. I think this is unsound. Instead, use `&[MaybeUninit<u8>]` internally.

I use `&[u8]` instead of `core::io::BorrowedCursor<'_>`, because there are currently no cases where the initialized portion needs to be separately tracked.

Since `smoltcp::socket::tcp::Socket::recv_slice` takes a `&[u8]`, inline it and refactor. As a bonus, this avoids a copy and clear in the error case and it makes it more clear that packets are dropped when given an insufficiently large buffer.

This PR enables implementing `std::io::Read::read_buf` for `std::fs::File` and `std::io::Stdin` on Hermit. That effort is tracked in https://github.com/rust-lang/rust/issues/136756.

### DIFF
--- a/src/fd/eventfd.rs
+++ b/src/fd/eventfd.rs
@@ -1,7 +1,7 @@
 use alloc::boxed::Box;
 use alloc::collections::vec_deque::VecDeque;
 use core::future::{self, Future};
-use core::mem;
+use core::mem::{self, MaybeUninit};
 use core::task::{Poll, Waker, ready};
 
 use async_lock::Mutex;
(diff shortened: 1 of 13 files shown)


thaliaarchi:
The CI failure seems likely to be unrelated:
```rust
Laplace iterations
Error: QEMU timeout
```

mkroening (approved):
Thanks a lot! :)

thaliaarchi:
How should I coordinate with the Hermit side when I submit the PR to Rust for `Read::read_buf`? Does this need to be released in Hermit for a while?

mkroening:
The kernel is built separately from the application and then linked together without LTO at the moment, so I think the chances of something bad happening due to undefined behavior are really slim in this case.

So I'd say you can go ahead with that PR right away. Feel free to Tag @stlankes and me in everything Hermit-related to keep us in the loop, of course. :)

PR #1611, opened 2025-02-23
fix: VIRTIO on UEFI

mkroening:
This PR adds support for the remaining paging functions on UEFI and sets up VIRTIO on UEFI.


mkroening:
CC: @sarahspberrypi

PR #1629, opened 2025-03-06
fix(riscv64): improved interrupt_handling with virtio_mmio devices

mrm-develop:
The function external_eoi() was removed to prevent premature end-of-interrupt (EOI) signaling before the interrupt was fully processed. 
In the previous implementation, external_eoi() was called too early, which could lead to lost or duplicated interrupts.
Additionally, the new implementation reduces redundant lock acquisitions (by external_eoi() and external_handler()) and improves efficiency by consolidating the interrupt handling into a single function. 
This ensures correct processing order and improves system stability, particularly for Virtio-MMIO-based network communication.


mkroening (approved):
Thanks!

PR #1639, opened 2025-03-13
feat(aarch64-interrupts): use local cpuid for interrupts

fstracke:
Instead of defaulting to core 0 for interrupt handling use the local cpu to dispatch interrupts. This allows for the handling of PPIs.


mkroening (approved):
Thanks! :)

Issue #200, opened 2021-05-28
Re-enable `mutable-noalias`


mkroening:
Requires:

* [x] #199

See:

* https://github.com/hermitcore/rusty-hermit/issues/128
* https://github.com/hermitcore/rusty-hermit/pull/135

PR #1652, opened 2025-03-20
perf: enable `mutable-noalias`

mkroening:
Closes https://github.com/hermit-os/kernel/issues/200.


mkroening:
Let's enable this again and hope for the best. We should be using volatile operations wherever needed now.

PR #1654, opened 2025-03-21
prepare kernel to publish source code as crate for upcoming releases

stlankes:


Cargo.toml
```diff
@@ -98,7 +98,7 @@ uninlined_format_args = "warn"
 unreadable_literal = "warn"
 
 [dependencies]
-hermit-macro = { path = "hermit-macro" }
+hermit-macro = { version = "0.1.0" }
```
mkroening, line 101: Let's enforce the specific version and allow using the path if possible.

```suggestion
hermit-macro = { version = "=0.1.0", path = "hermit-macro" }
```

mkroening (changes requested):

mkroening (approved):

PR #1655, opened 2025-03-23
feat(time): add `SystemTime::UNIX_EPOCH` and `SystemTime::duration_since`

stlankes:


jounathaen:
Looks good. As Martin perfers to have every commit with a successful CI, I'd suggest to squash commits 1-4 or reorder them.

PR #1665, opened 2025-04-03
fix(build): error message for missing llvm_tools

mrm-develop:


mkroening:
Interesting! 🤔

How did you trigger this? As far as I understand, this should truly be unreachable. Did you modify or remove the `rust-toolchain.toml`?

mrm-develop:
Nope, I used a fresh copy of the original repository on Ubuntu. There were some installation problems with the development software. Then I had the error. After manually installing llvm-tools it started working.

xtask/src/binutil.rs
```diff
@@ -33,7 +33,14 @@ impl LlvmTools {
...
 		}
-		unreachable!()
+		eprintln!(
+			"Error: Could not find the required LLVM tool '{}'. Searched in: {:?}",
+			example_exe, rustlib
+		);
+		Err(io::Error::new(
+			io::ErrorKind::NotFound,
+			"Required LLVM tool not found",
+		))
```
mkroening, line 43: Can we align the error with [build.rs#L77-L81](https://github.com/hermit-os/kernel/blob/1f503cd8b8937cac564d495580bbab55d223065c/build.rs#L77-L81) and remove the output? It should get printed via the error message.

```suggestion
		Err(io::Error::new(
			io::ErrorKind::NotFound,
			"Could not find llvm-tools component\n\
			\n\
			Maybe the rustup component `llvm-tools` is missing? Install it through: `rustup component add llvm-tools`"
		))
```

mkroening (changes requested):
I cannot reproduce this:

```bash
docker run --rm -it rust:latest
apt-get update && apt-get install -y --no-install-recommends qemu-system-x86
git clone --recurse-submodules https://github.com/hermit-os/hermit-rs.git
cd hermit-rs/kernel
curl -LO "https://github.com/hermit-os/loader/releases/download/v0.5.4/hermit-loader-x86_64"
cargo xtask ci rs --arch x86_64 --package hello_world qemu
```

But maybe it is reachable if the installation gets corrupted, I guess.

mkroening (approved):

PR #1670, opened 2025-04-07
perf(device_alloc): don't re-map device memory

mkroening:
This increases bandwidth from 400 Mbit/s to 550 Mbit/s for me.

### DIFF
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -2,11 +2,9 @@ use core::alloc::{AllocError, Allocator, Layout};
 use core::ptr::{self, NonNull};
 
 use align_address::Align;
+use memory_addresses::PhysAddr;
 
-use crate::arch;
-#[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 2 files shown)


src/mm/device_alloc.rs
```diff
@@ -37,8 +29,6 @@ unsafe impl Allocator for DeviceAlloc {
 		let addr = ptr.as_ptr().expose_provenance().into();
 
 		if let Some(phys_addr) = arch::mm::paging::virtual_to_physical(addr) {
```
Gelbpunkt, line 31: This is now redundant, virtual and physical address are the same, we can just directly deallocate the address returned by `expose_provenance`.

Gelbpunkt (changes requested):
Very nice! From my quick testing this roughly doubles the throughput I see in benchmarks, not very surprising given we allocate a lot

src/mm/device_alloc.rs
mkroening: Yes! Thanks! :)

Gelbpunkt (approved):

PR #1685, opened 2025-04-14
Improve aarch64 support

stlankes:
This PR is a collection of commits to improve the aarch64 support. It provides following features:

- The arm64 MMU supports a contiguous bit which is a hint that the TTE is one of a set of contiguous entries which can be cached in a single TLB entry. To support this feature, the function `map` checks if contiguous region has to be mapped into the address space. In addition, the function `map_heap` tries to allocate large memory blocks to increase the propobility of contiguos memory regions.
- Workaround for M1 processors: add additional TLB flushes
- Improve kernel message 
- According to other kernels and tutorials, a memory barrier should be added before and after modifying DAIF
- Add busy waiting loop to initialize processor one after the other
- Align CPU_ONLINE to a 8 byte boundary, which is required to support relocation.


stlankes:
@mkroening @fstracke Why do test SMP support on aarch64. It has never worked

PR #1693, opened 2025-04-22
perf(virtio-net): Don't copy packets when num_buffers is 1

Gelbpunkt:
Currently, we would copy every single received packet from `DeviceAlloc` to a new `Vec` that is allocated with `GlobalAlloc`. If we use `DeviceAlloc` for the payload of the `RxToken`, we can avoid copying all packets for the special (and common) case of `num_buffers = 1`.

While at it, I refactored the code to make it more readable and avoid allocating the intermediate `Vec` that would be collected into a flattened `Vec<u8>` anyways.

My benchmarking shows that this increases throughput over TCP by roughly ~20%.

### DIFF
--- a/src/drivers/net/gem.rs
+++ b/src/drivers/net/gem.rs
@@ -340,7 +340,7 @@ impl NetworkDriver for GEMDriver {
 				};
 				trace!("BUFFER: {buffer:x?}");
 				self.rx_buffer_consumed(index as usize);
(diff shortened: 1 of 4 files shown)


Gelbpunkt:
Oops, the other net drivers will need to be adjusted as well. I'll do that in follow-up commits

mkroening (approved):
Thanks! :)

PR #1694, opened 2025-04-25
switch to crate arm_pl031 to read the realtime clock

stlankes:
If the application is running on top of Qemu, the ARM PrimeCell Real Time Clock(PL031) can be used to determine the boot time. By this patch, the crate arm_pl031 is used as interface to the real time clock.

uhyve doesn't support the real-time clock. To detemine the unix timestamp, the boot time from bootinfo is used

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arm_pl031"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "13696b1c2b59992f4223e0ae5bb173c81c63039367ca90eee845346ad2a13421"
+dependencies = [
+ "chrono",
+]
+
+[[package]]
 name = "async-lock"
 version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


src/arch/aarch64/kernel/processor.rs
```diff
@@ -221,27 +207,14 @@ pub fn detect_frequency() {
 fn __set_oneshot_timer(wakeup_time: Option<u64>) {
 	if let Some(wt) = wakeup_time {
 		// wt is the absolute wakeup time in microseconds based on processor::get_timer_ticks.
-		let deadline = u128::from(wt) * u128::from(CPU_FREQUENCY.get()) / 1_000_000;
-		let deadline = u64::try_from(deadline).unwrap();
+		let deadline = (wt * u64::from(get_frequency())) / 1000;
```
jounathaen, line 210: I assume this is fine, but I'd suggest double-checking that a `u64` is sufficient here (was `u128` before).

jounathaen:
In general: LGTM :+1:

src/arch/aarch64/kernel/processor.rs
stlankes: move this part to #1703

mkroening (approved):
Looks good to me. 👍 

First, I had concerns regarding the license: "GPL-3.0-or-later OR Apache-2.0 OR MulanPSL-2.0"

This is why I added cargo-deny:
- https://github.com/hermit-os/kernel/pull/1715

This seems fine, though, so please rebase and we can merge. :)

PR #1701, opened 2025-04-30
ci(gem-net): test GEM NIC on RISC-V with QEMU

cagatay-y:


cagatay-y:
Depends on  #686.

mkroening (approved):
Thanks! :)

Issue #638, opened 2023-01-03
Supply correct image filename


mkroening:
The first argument to the application (the path to the application itself) is filled with a placeholder (`{name}`) at the moment.

https://github.com/hermitcore/libhermit-rs/blob/1d72195fe6ef198a8d48cc7ed6e190293a8f7ad1/src/syscalls/interfaces/mod.rs#L22

shehreenr:
Hello! I am a computer science student from UT Austin. In my virtualization course, our final project requires us to contribute to an open-source project, and this project looked interesting! Can my group and I be assigned to this issue?

mkroening:
Sure! Though I am not sure, what we can supply there instead. I don't know of a good name to get the binary name. Maybe changing `{name}` to `hermit-app` would already be an improvement? 🤔

mkroening:
Maybe this could be a weak static symbol that can be overridden by applications. 🤔

shehreenr:
Thank you for your quick response! Due to time constraints, I decided to go with the simple implementation of changing the name to "hermit-app." I have submitted a PR (https://github.com/hermit-os/kernel/actions/runs/14824394983/job/41615620850?pr=1706) but it appears to fail the "Format" check. Could you explain what the error is? I am having a hard time interpreting it. Thank you!

mkroening:
I don't think there is a good way to resolve this. The fix introduced in https://github.com/hermit-os/kernel/pull/1535 should be fine.

PR #1706, opened 2025-05-04
Change default image filename from `{name}` to `hermit-app`

shehreenr:
Related to https://github.com/hermit-os/kernel/issues/638.


mkroening (changes requested):
The CI complains about the formatting.

The issue is that you removed the trailing newline from the file. `cargo fmt` should fix the formatting. :)

mkroening:
This was done in https://github.com/hermit-os/kernel/pull/1535.

PR #1712, opened 2025-05-07
perf(drivers): remove `virt_to_phys`

mkroening:
This is made possible by https://github.com/hermit-os/kernel/pull/1670.

I can't measure an improvement reliably, but this should be a strict performance improvement and also be visible on the flamegraph.


Gelbpunkt (approved):

mkroening:
I'll wait for https://github.com/hermit-os/kernel/pull/1701 before merging to make sure, this does not break gem-net.

cagatay-y (approved):

PR #1714, opened 2025-05-07
refactor(drivers/net): Unify MTUs and fix buffer sizes for virtio-net

Gelbpunkt:
See "5.1.6.3.1 Driver Requirements: Setting Up Receive Buffers" in the VirtIO specification. We only need to allocate at least the MTU if GUEST_TSO4, GUEST_TSO6 or GUEST_UFO were negotiated, otherwise we SHOULD allocate at least 1514 (1526 minus 12 bytes header) for the payload. Currently we were allocating 65550 (MTU) and always receiving 1514 in that case.

### DIFF
--- a/src/drivers/net/virtio/mod.rs
+++ b/src/drivers/net/virtio/mod.rs
@@ -25,7 +25,7 @@ use volatile::access::ReadOnly;
 use self::constants::MAX_NUM_VQ;
 use self::error::VirtioNetError;
 use crate::config::VIRTIO_MAX_QUEUE_SIZE;
-use crate::drivers::net::NetworkDriver;
+use crate::drivers::net::{NetworkDriver, mtu};
 #[cfg(not(feature = "pci"))]
 use crate::drivers::virtio::transport::mmio::{ComCfg, IsrStatus, NotifCfg};
 #[cfg(feature = "pci")]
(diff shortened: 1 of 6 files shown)


src/drivers/net/virtio/mod.rs
```diff
@@ -65,8 +65,13 @@ impl RxQueues {
 		//
 		let packet_size = if dev_cfg.features.contains(virtio::net::F::MRG_RXBUF) {
 			1514
-		} else {
+		} else if dev_cfg.features.contains(virtio::net::F::GUEST_TSO4)
+			|| dev_cfg.features.contains(virtio::net::F::GUEST_TSO6)
+			|| dev_cfg.features.contains(virtio::net::F::GUEST_UFO)
+		{
 			dev_cfg.raw.as_ptr().mtu().read().to_ne().into()
```
mkroening, line 72: If MTU is 65550, should we not add the size of `virtio_net_hdr` (12 Bytes) here?

src/drivers/net/virtio/mod.rs
```diff
@@ -65,8 +65,13 @@ impl RxQueues {
...
+		} else {
+			1514
```
cagatay-y, line 74: I think this branch will cause an error when we have MTU negotiated. 

> If the driver negotiates VIRTIO_NET_F_MTU, it MUST supply enough receive buffers to receive at least one receive packet of size mtu (plus low level ethernet header length) with gso_type NONE or ECN.

I don't know what MTU the device reports if guest segmentation offload is negotiated, but I think the safe behaviour is to use a constant 65550 (not the mtu) as the buffer size when it is negotiated, and if not, to use the MTU.  

([VIRTIO Specification v1.2, § 5.1.4.2](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-2250002))

cagatay-y (changes requested):

cagatay-y:
By the way, the way we handle MTUs in general is wrong. At the point the RxQueues are created, the feature negotiation is not complete and we don't know if the device supports MTU reporting, but we try to read that field here. I have to rework the driver initialisation to avoid that.

Gelbpunkt:
> By the way, the way we handle MTUs in general is wrong. At the point the RxQueues are created, the feature negotiation is not complete and we don't know if the device supports MTU reporting, but we try to read that field here. I have to rework the driver initialisation to avoid that.

I hope I got it right now. I refactored the driver initialization to only initialize the virtqueues after the feature negotiation is complete. It probably needs some more polishing, but it looks right now. I realized that if we actually use the negotiated features, I can see that `MRG_RXBUF` is negotiated for me, which explains what I initially observed.

mkroening:
Is the first committee specification draft of VIRTIO 1.3 helpful? I think it answers my first comment with yes:

[Virtual I/O Device (VIRTIO) Version 1.3—5.1.6.3 Setting Up Receive Buffers](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-2510003)

mkroening (approved):

PR #1726, opened 2025-05-20
ci: add continuous benchmarking

CarlWachter:
This will execute the benchmarks in the hermit-rs repo according to the `bench.json` file. Results will be published to Github Pages.

CI will fail until https://github.com/hermit-os/hermit-rs/pull/708 gets merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   merge_group:
-
+  
 env:
   GH_TOKEN: ${{ github.token }}
   RUSTFLAGS: -Dwarnings
@@ -240,3 +240,56 @@ jobs:
         if: matrix.arch == 'x86_64'
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} --package thread_test --smp 4 qemu ${{ matrix.flags }}
         if: matrix.arch == 'x86_64'
+
+  benchmark:
+    name: Bench
+    runs-on: [self-hosted]
+    steps:
+    - name: Checkout hermit-rs
+      uses: actions/checkout@v4
+      with:
+        repository: hermit-os/hermit-rs
+        submodules: true
+    - name: Remove hermit-kernel submodule
+      run: git rm -r kernel
+      working-directory: .
+    - name: Checkout hermit-kernel
+      uses: actions/checkout@v4
+      with:
+        path: kernel
+    - name: Install QEMU
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-system qemu-system-x86 qemu-kvm
+    - uses: mkroening/rust-toolchain-toml@main
+    - name: Download loader  
+      run: |
+        sudo apt install gh -y
+        gh release download --repo hermit-os/loader --pattern hermit-loader-x86_64
+    - name: Install uhyve
+      run: cargo +stable install --locked uhyve
+    - name: Check KVM availability
+      shell: bash
+      run: |
+         lscpu
+         kvm-ok
+    - name: Add wasm target
+      run: |
+         rustup component add llvm-tools
+         rustup target add wasm32-wasip1
+         rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
+    - name: Run benchmarks
+      uses: hermit-os/hermit-bench@kernel
+      with: 
+        benchmark-file: /kernel/.github/benchmarks/bench.json
+        build-command: "HERMIT_LOG_LEVEL_FILTER=Off cargo build -Zbuild-std=std,panic_abort --target x86_64-unknown-hermit --release"
+    - name: Store benchmark result
+      # Only save the result if the commit is on main
+      # This is to avoid saving results for PRs
+      if: (github.event_name == 'push' && github.ref == 'refs/heads/main')
+      uses: hermit-os/github-action-benchmark@hermit-rs
+      with:
+        tool: 'hermit-bench'
+        output-file-path: results.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        auto-push: true
--- a/.github/benchmarks/bench.json
+++ b/.github/benchmarks/bench.json
@@ -1,0 +1,127 @@
+[
+    {
+       "name": "1 core",
+       "command": "sudo qemu-system-x86_64 -display none -smp 1 -m 128M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/startup_benchmark",
+       "external_time": true,
+       "iterations": 25,
+       "group": "General",
+       "plot_group": "Startup Time"
+    },
+    {
+        "name": "2 cores",
+        "command": "sudo qemu-system-x86_64 -display none -smp 2 -m 128M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/startup_benchmark",
+        "external_time": true,
+        "iterations": 25,
+        "group": "General",
+        "plot_group": "Startup Time"
+     },
+    {
+        "name": "4 cores",
+        "command": "sudo qemu-system-x86_64 -display none -smp 4 -m 128M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/startup_benchmark",
+        "external_time": true,
+        "iterations": 25,
+        "group": "General",
+        "plot_group": "Startup Time"
+    },
+    {
+        "name": "alloc_benchmarks",
+        "path": "target/x86_64-unknown-hermit/release/alloc_benchmarks",
+        "group": "File Size",
+        "plot_group": "Benchmark File Size"
+    },
+    {
+        "name": "micro_benchmarks",
+        "path": "target/x86_64-unknown-hermit/release/micro_benchmarks",
+        "group": "File Size",
+        "plot_group": "Benchmark File Size"
+    },
+    {
+        "name": "mutex test",
+        "path": "target/x86_64-unknown-hermit/release/mutex",
+        "group": "File Size",
+        "plot_group": "Benchmark File Size"
+    },
+    {
+        "name": "Micro Benchmarks",
+        "command": "sudo qemu-system-x86_64 -display none -smp 4 -m 512M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/micro_benchmarks",
+        "iterations": 20,
+        "group": "Micro"
+    },
+    {
+        "name": "Allocation Benchmarks",
+        "command": "sudo qemu-system-x86_64 -display none -smp 1 -m 5G -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/alloc_benchmarks",
+        "iterations": 30,
+        "group": "Allocations"
+    },
+    {
+        "name": "Mutex Stress Test",
+        "command": "sudo qemu-system-x86_64 -display none -smp 2 -m 512M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/mutex",
+        "iterations": 40,
+        "group": "Mutex"
+    },
+    {
+        "name": "Pi Multithreaded Benchmark",
+        "command": "sudo qemu-system-x86_64 -display none -smp 4 -m 512M -serial stdio -enable-kvm -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -kernel hermit-loader-x86_64 -initrd target/x86_64-unknown-hermit/release/multithreaded_benchmark",
+        "iterations": 15,
+        "group": "General",
+        "plot_group": "Multithreaded Pi Efficiency"
+    },
+    {
+        "name": "Netbench TCP BW - Client",
+        "command": "parallel ::: 'cargo run --manifest-path ./Cargo.toml --bin tcp-server-bw --release --target x86_64-unknown-linux-gnu -- --address 10.0.5.3 --bytes 1048576 --rounds 1000' 'sleep 10 && sudo qemu-system-x86_64 -display none -serial stdio -kernel hermit-loader-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -enable-kvm -initrd target/x86_64-unknown-hermit/release/tcp-client-bw -smp 2 -m 1024M -netdev user,id=u1,hostfwd=tcp::9975-:9975,hostfwd=udp::9975-:9975,net=192.168.76.0/24,dhcpstart=192.168.76.9 -device virtio-net-pci,netdev=u1,disable-legacy=on,packed=on,mq=on -append \"-- --nonblocking --address 127.0.0.1 --bytes 1048576 --rounds 1000\"'",
+        "iterations": 3,
+        "group": "Networking",
+        "plot_group": "Netbench Bandwidth - 1000x1MB"
+    },
+    {
+        "name": "Netbench TCP BW - Server",
+        "command": "parallel ::: 'sleep 10 && cargo run --manifest-path ./Cargo.toml --bin tcp-client-bw --release --target x86_64-unknown-linux-gnu -- --nonblocking --address 127.0.0.1 --bytes 1048576 --rounds 1000' 'sudo qemu-system-x86_64 -display none -serial stdio -kernel hermit-loader-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -enable-kvm -initrd target/x86_64-unknown-hermit/release/tcp-server-bw -smp 2 -m 1024M -netdev user,id=u1,hostfwd=tcp::7878-:7878,hostfwd=udp::9975-:9975,net=192.168.76.0/24,dhcpstart=192.168.76.9 -device virtio-net-pci,netdev=u1,disable-legacy=on,packed=on,mq=on -append \"-- --address 10.0.5.3 --bytes 1048576 --rounds 1000\"'",
+        "iterations": 3,
+        "group": "Networking",
+        "plot_group": "Netbench Bandwidth - 1000x1MB"
+    },
+    {
+        "name": "Netbench TCP Latency - Client",
+        "command": "parallel ::: 'cargo run --manifest-path ./Cargo.toml --bin tcp-server-latency --release --target x86_64-unknown-linux-gnu -- --address 10.0.5.3 --bytes 1048576 --rounds 250' 'sleep 10 && sudo qemu-system-x86_64 -display none -serial stdio -kernel hermit-loader-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -enable-kvm -initrd target/x86_64-unknown-hermit/release/tcp-client-latency -smp 2 -m 1024M -netdev user,id=u1,hostfwd=tcp::9975-:9975,hostfwd=udp::9975-:9975,net=192.168.76.0/24,dhcpstart=192.168.76.9 -device virtio-net-pci,netdev=u1,disable-legacy=on,packed=on,mq=on -append \"-- --nonblocking --address 127.0.0.1 --bytes 1048576 --rounds 250\"'",
+        "iterations": 3,
+        "group": "Networking",
+        "plot_group": "Netbench Latency - 250x1MB"
+    },
+    {
+        "name": "Netbench TCP Latency - Server",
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -240,3 +243,54 @@ jobs:
         if: matrix.arch == 'x86_64'
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} --package thread_test --smp 4 qemu ${{ matrix.flags }}
         if: matrix.arch == 'x86_64'
+
+  benchmark:
+    name: Bench
+    runs-on: ubuntu-latest
```
jounathaen, line 249:
```suggestion
    runs-on: [self-hosted]
```

jounathaen:
Ci is unfortunately still failing, despite https://github.com/hermit-os/hermit-rs/pull/708 being merged and also used.

CarlWachter:
@jounathaen it's passing now and results look good

jounathaen:
The benchmarks take 27 minutes now, which is 10 mins longer than the tests. Can we change the benchmark setup, so that the CI duration is not dominated by the benchmarks anymore (parallelize it on two runners or reduce the benchmark size)?

jounathaen:
> I think the benchmarks taking more time is fine, but I would suggest running the benchmark suite after merging and on PRs only on demand.

But in that case, we wouldn't know if a PR affects performance before merging it.

jounathaen:
As @mkroening pointed out, we already have gh-pages installed for this repository (https://hermit-os.github.io/kernel/hermit/). We would need to find a way of publishing the results on a different URL. Martin suggested a separate repository for gh-pages, where this action pushes to. Maybe a combination of both gh-pages into subdirectories is also possible.

.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ name: CI
 on:
   pull_request:
   merge_group:
-
+
```
jounathaen: Tiny nitpick: Whitespace added

.github/workflows/publish_docs.yml
```diff
@@ -38,5 +38,6 @@ jobs:
         with:
           target_branch: gh-pages
           build_dir: target/x86_64-unknown-none/doc
+          keep_history: true
```
mkroening:
```suggestion
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+            uses: hermit-os/github-action-benchmark@hermit-rs
+            with:
+              tool: 'hermit-bench'
+              output-file-path: ${{ steps.run-bench.outputs.result-file }}
+              github-token: ${{ secrets.GITHUB_TOKEN }}
+              benchmark-data-dir-path: benchmarks
+              comment-always: true
+              benchmark-matrix-name: ${{ matrix.benchmark-matrix-name }}
+              auto-push: ${{ steps.publish-benchmarks.outputs.publish_benchmarks == 'true' }} 
+
```
mkroening: Trailing whitespace:

```suggestion

```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
+name: Benchmark
+
+on:
+  pull_request_target:
```
mkroening, line 4:
```suggestion
  pull_request:
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+          - name: Remove hermit-kernel submodule
+            run: git rm -r kernel
+          - name: Checkout hermit-kernel
+            uses: actions/checkout@v4
+            with:
+              path: kernel
+          - name: Install dependencies
+            run: |
+              sudo apt-get update
+              sudo apt-get install -y qemu-system qemu-system-x86 qemu-kvm parallel gh
```
mkroening: Why install QEMU thrice? This should also not install recommended packages to avoid pulling in the QEMU GUI dependencies.

```suggestion
              sudo apt-get install -y --no-install-recommends qemu-system-x86 parallel gh
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+            run: cargo +stable install --locked uhyve
+          - name: Check KVM availability
+            run: |
+              lscpu
+              kvm-ok
+          - name: Add wasm target
+            run: |
+              rustup component add llvm-tools
+              rustup target add wasm32-wasip1
+              rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
```
mkroening: This does more than just adding the wasm target. We should align the toolchain installation to the CI, by unifying this with line 51 and moving this after the package installation:

```suggestion
      - uses: dtolnay/rust-toolchain@stable
      - uses: dtolnay/rust-toolchain@nightly
        with:
          components: rust-src
      - uses: mkroening/rust-toolchain-toml@main
      - run: rustup component add llvm-tools
        working-directory: .
      - run: rustup target add wasm32-wasip1
        working-directory: .
      - uses: mkroening/rust-toolchain-toml@main
        with:
          toolchain-file: "kernel/rust-toolchain.toml"
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+          - name: Run benchmarks
+            uses: hermit-os/hermit-bench@kernel
+            id: run-bench
+            with: 
+              benchmark-file: ${{ matrix.benchmark-file }}
+              benchmark-build: ${{ matrix.benchmark-build }}
+              build-command: "HERMIT_LOG_LEVEL_FILTER=Off cargo build -Zbuild-std=std,panic_abort --target x86_64-unknown-hermit --release"
```
mkroening: Just out of curiosity: Would having the default log level slow us down so much? I'd say that we should benchmark the default, which should not affect performance once started at all and if startup performance were bad, we should fix that instead of disabling output.

It would be great to also see that output, of course, for debugging and reproducing.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+          - name: Store benchmark results
+            uses: hermit-os/github-action-benchmark@hermit-rs
+            with:
+              tool: 'hermit-bench'
+              output-file-path: ${{ steps.run-bench.outputs.result-file }}
+              github-token: ${{ secrets.GITHUB_TOKEN }}
+              benchmark-data-dir-path: benchmarks
+              comment-always: true
+              benchmark-matrix-name: ${{ matrix.benchmark-matrix-name }}
+              auto-push: ${{ steps.publish-benchmarks.outputs.publish_benchmarks == 'true' }}
```
mkroening: Could we not just use the corresponding variables here directly, instead of mixing context and environment variables in an extra step?

```suggestion
              auto-push: ${{ github.event_name == "push" && github.ref == "refs/heads/main" }} 
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+          - name: Determine if benchmarks should be published
+            id: publish-benchmarks
+            run: |
+              if [[ "${{ github.event_name }}" == "push" && "${GITHUB_REF##*/}" == "main" ]]; then
+                echo "publish_benchmarks=true" >> $GITHUB_OUTPUT
+              else
+                echo "publish_benchmarks=false" >> $GITHUB_OUTPUT
+              fi
+          - name: Store benchmark results
+            uses: hermit-os/github-action-benchmark@hermit-rs
```
mkroening: Is there a reason we are using `hermit-rs` instead of `main` for the default branch for that repo?

.github/benchmarks/netbench.json
```diff
@@ -0,0 +1,108 @@
...
+    },
+    {
+        "name": "Netbench UDP Latency - Server",
+        "path": "target/x86_64-unknown-hermit/release/udp-server-latency",
+        "group": "File Size",
+        "plot_group": "Netbench Benchmark File Size"
+    },
+    {
+        "name": "Netbench TCP BW - Client",
+        "command": "parallel ::: 'cargo run --manifest-path ./Cargo.toml --bin tcp-server-bw --release --target x86_64-unknown-linux-gnu -- --address 10.0.5.3 --bytes 1048576 --rounds 1000' 'sleep 10 && sudo qemu-system-x86_64 -display none -serial stdio -kernel hermit-loader-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -enable-kvm -initrd target/x86_64-unknown-hermit/release/tcp-client-bw -smp 2 -m 1024M -netdev user,id=u1,hostfwd=tcp::9975-:9975,hostfwd=udp::9975-:9975,net=192.168.76.0/24,dhcpstart=192.168.76.9 -device virtio-net-pci,netdev=u1,disable-legacy=on,packed=on,mq=on -append \"-- --nonblocking --address 127.0.0.1 --bytes 1048576 --rounds 1000\"'",
```
mkroening: Is there any way of making these giant lines more readable? In YAML, I'd imagine something like

```yaml
command: |
  parallel \
    'cargo run ...' \
    && 'sleep 10 \
      && qemu -cpu ... \
        -initrd ...'
```

mkroening: The QEMU CPU args are not ideal.

For benchmarking, they should probably be `-cpu host,migratable=no,+invtsc`.

mkroening: I don't think we should be using nonblocking sockets anywhere. The application does nothing else meanwhile and just tries again.

`tcp-server-bw` even crashes, which is probably why it is only used on the client side, but the client does nothing useful either:

```
thread 'main' panicked at benches/netbench/src/rust-tcp-bw/server.rs:24:37:
called `Result::unwrap()` on an `Err` value: Kind(WouldBlock)
```

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,88 @@
...
+            run: |
+              lscpu
+              kvm-ok
+          - name: Add wasm target
+            run: |
+              rustup component add llvm-tools
+              rustup target add wasm32-wasip1
+              rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
+          - name: Run benchmarks
+            uses: hermit-os/hermit-bench@kernel
```
mkroening: Is there a reason we are using the kernel branch instead of main?

mkroening: Is there documentation about the inputs to this action (`with`)? I did not find it in the actions' README.

.github/benchmarks/netbench.json
mkroening: Is there a reason for using user networking instead of taps? I get more than 15% more bandwidth on my machine.

.github/benchmarks/netbench.json
```diff
@@ -0,0 +1,108 @@
...
+        "iterations": 3,
+        "group": "Networking",
+        "plot_group": "Netbench Latency - 250x1MB"
+    },
+    {
+        "name": "Netbench TCP Latency - Server",
+        "command": "parallel ::: 'sleep 10 && cargo run --manifest-path ./Cargo.toml --bin tcp-client-latency --release --target x86_64-unknown-linux-gnu -- --nonblocking --address 127.0.0.1 --bytes 1048576 --rounds 250' 'sudo qemu-system-x86_64 -display none -serial stdio -kernel hermit-loader-x86_64 -cpu qemu64,apic,fsgsbase,rdtscp,xsave,xsaveopt,fxsr,rdrand -enable-kvm -initrd target/x86_64-unknown-hermit/release/tcp-server-latency -smp 2 -m 1024M -netdev user,id=u1,hostfwd=tcp::7878-:7878,hostfwd=udp::9975-:9975,net=192.168.76.0/24,dhcpstart=192.168.76.9 -device virtio-net-pci,netdev=u1,disable-legacy=on,packed=on,mq=on -append \"-- --address 10.0.5.3 --bytes 1048576 --rounds 250\"'",
```
mkroening: The forwarded ports for user networking are quite inconsistent right now. The application uses port 7878 by default and is configurable via CLI, but parts of the UDP and latency benches have 9975 hardcoded, which should be fixed.

The QEMU invocations should only forward the correct ports that are necessary.

mkroening (changes requested):
Thanks for the PR! :)

I have a few requests. :D

.github/workflows/publish_docs.yml
CarlWachter: If that line is removed, the benchmark data will be deleted by the publish_docs push every time as it overwrites the whole repo

.github/workflows/benchmark.yml
CarlWachter: `pull_request` would not grant the action the necessary permissions to write comments in the PR if pulling from a fork

.github/workflows/benchmark.yml
CarlWachter: Originally it was so that the output format wouldn't brea, but i'll try it without it since it should no longer be an issue

.github/workflows/benchmark.yml
CarlWachter: I used to split the branches like that into hermit-rs/hermit-playground but since that is outdated now i've changed it to main for the hermit kernel support

.github/workflows/benchmark.yml
CarlWachter: I've added it to the actions README

.github/workflows/publish_docs.yml
mkroening: Ah, sorry. I missed https://github.com/hermit-os/kernel/pull/1726#issuecomment-2929839557. I would personally prefer not mixing the git history for docs and benches. Would committing to a branch on the hermit-bench repo not be a better approach?

CarlWachter: Yeah you're right, I'll do that

.github/workflows/benchmark.yml
mkroening: Sorry, I also missed https://github.com/hermit-os/kernel/pull/1726#issuecomment-2930066282. We should move to top-level right under `on` then, to restrict the default write access for the whole workflow, right? 

```yaml
permissions:
  pull-requests: write
```

CarlWachter: Ah i did not know you could define that for the whole workflow rather than just one job. Moved that down 👍

.github/benchmarks/netbench.json
CarlWachter: I've changed the forwarded ports to be minimal now. As for the hardcoding that is in reference to the port that is being connected to, rather than the one that is binded to (which is determined by the CLI arg). Since this is explicitly a benchmark, not an example, i don't see any issue with that being hardcoded.

.github/workflows/benchmark.yml
mkroening: Thanks! :)

What kind of code in the action is specific to the benchmarked repo?

CarlWachter: For `github-action-benchmark` it was only the name of the kernel folder, to get the corresponding commit, but this is no longer relevant, i.e. we could very easily add hermit-c benchmarks

.github/benchmarks/netbench.json
mkroening: Ah, I see. So you mean that the remote port we connect to is inferred from CLI but the local port that we send the request from is hardcoded, right? In that case, the port should better be chosen by the operating system by specifying 0, or does Hermit not support that?

.github/workflows/benchmark.yml
mkroening: Great! :)

.github/benchmarks/netbench.json
CarlWachter: Yes i've tried that and had no success getting it to bind to port 0, always failing when i attempted that. It works when i run the code like that on a non-hermit system so i assume hermit does not support that.

.github/benchmarks/netbench.json
CarlWachter: No particular reason that i can think of, I'd be happy to switch to taps if you tell me the args you used to achieve that performance bump

.github/workflows/benchmark.yml
CarlWachter: Changed that...

n0toose:
> But in that case, we wouldn't know if a PR affects performance before merging it.

Hi, just chiming in: The (relatively newer) `workflow_dispatch` event could help with triggering workflows on demand, should a reviewer think that a PR could affect performance: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

.github/benchmarks/netbench.json
CarlWachter: Agreed and changed

CarlWachter: Short of switching to a different format, i couldn't think of anything. Given that these should only need to be set very rarely I think this is something we can live with

CarlWachter:
I've addressed most of the issues brought up now, but once again ran into issues with permissions. Namely pushing to the `hermit-bench` Repo is not permissible with the default `GITHUB_TOKEN`. 

@jounathaen  @mkroening  How would you feel about a Organisation owned PAT? It would allow the action to push to `hermit-bench` and be able to function on the standard `pull_request` trigger.

.github/benchmarks/netbench.json
mkroening: It should work as described in the [loader's README](https://github.com/hermit-os/loader?tab=readme-ov-file#network-support). :)

.github/benchmarks/netbench.json
mkroening: Oh, I see. I have opened an issue for supporting ephemeral ports (https://github.com/hermit-os/kernel/issues/1753). Could you add a TODO note that the hardcoded value is not relevant and should be changed to zero once possible? I am also wondering if changing 7878 to 9975 as well would be more or less confusing. 🤔

CarlWachter: Shouldn't that issue be on `hermit-rs`, since that is where the benchmarks with the hardcoded ports are?

mkroening: Well, the issue is that the kernel does not support that. Sure, user space should be changed once the root issue is resolved.

CarlWachter:
I've cut netbench out for now as i have not been able to test with tap devices and there is a kernel side incompatibility of the current netbench implementation with the kernel anyhow (https://github.com/hermit-os/kernel/pull/1780). 

Otherwise this PR is ready and I would recommend we merge this today without netbench.

I once again tested everything on my fork to ensure: https://github.com/CarlWachter/kernel/pull/9

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,80 @@
...
+  merge_group:
+  push:
+    branches:
+      - main
+
+permissions:
+          pull-requests: write
+
+env:
+  GH_TOKEN: ${{ github.token }}
```
mkroening: Nit: since these are more permissive than the default, we could move them to job scope.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,80 @@
+name: Benchmark
+
+on:
+  pull_request_target:
+  merge_group:
+  push:
+    branches:
+      - main
```
mkroening: Having both merge_group and push to main is redundant, right?

We should discuss which one to keep.
If the benchmarks working is already guaranteed by CI, it would be fine to defer to main to free up resources in the merge queue.
If benchmarks might fail independently of the CI, we have two options: consider benchmarks as required checks that are tests semantically, or make the CI tests test what could break in benchmarks.

I'd prefer making the benchmark not required for merging, since they run on self-hosted runners that might fail or require maintenance and should not hold up upstream.

.github/benchmarks/general.yaml
```diff
@@ -0,0 +1,71 @@
+- name: 1 core
+  bin: startup_benchmark
```
mkroening: I am not sure how this looks on the web page (https://carlwachter.github.io/hermit-bench/ does not seem to have GH pages activated), but I was wondering about the name.
So this is the name of a plot line inside of a plot with the plot group name?
If so, moving group and plot_group up might make this more obvious.

.github/benchmarks/general.yaml
```diff
@@ -0,0 +1,71 @@
...
+  manifest_path: Cargo.toml
```
mkroening: This is relative to hermit-rs, right? Would it make sense to make this relative to the kernel repo (../Cargo.toml)?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,80 @@
...
+        runs-on: [self-hosted]
+        environment: ${{ github.event_name == 'push' && 'benchmark' || '' }}
+
+        strategy:
+          matrix:
+            include:
+              - benchmark-file: /kernel/.github/benchmarks/general.yaml
+                benchmark-matrix-name: General
+              - benchmark-file: /kernel/.github/benchmarks/misc.yaml
+                benchmark-matrix-name: Misc
```
mkroening: I am confused why absolute paths work here. These should be relative paths, right?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,80 @@
...
+          - name: Remove hermit-kernel submodule
+            run: git rm -r kernel
+          - name: Checkout hermit-kernel
+            uses: actions/checkout@v4
+            with:
+              path: kernel
+          - name: Install dependencies
+            run: |
+              sudo apt-get update
+              sudo apt-get install -y --no-install-recommends qemu-system-x86 cpu-checker parallel gh
```
mkroening:
```suggestion
              sudo apt-get install -y --no-install-recommends qemu-system-x86 cpu-checker gh
```

.github/benchmarks/general.yaml
CarlWachter: https://carlwachter.github.io/hermit-bench/benchmarks/ is the url at the moment (ill put it in the root of the repo in my next commit).

The bin option is there so it can build it and is only used to label the corresponding plots in the Build and File Size tabs

.github/workflows/benchmark.yml
CarlWachter: I agree, since they are checked during the pull request anyways. In the unlikely case that a benchmark fails only on after the merge only one data point is lost. Changing this

.github/benchmarks/general.yaml
CarlWachter: Since the benchmarks are in hermit-rs i think it makes sense to keep it relative to there. Especially since the build process is inherently relative to hermit-rs (since we are building hermit-rs with the current kernel, not the other way around) it makes sense to keep benchmark execution relative to it too. 

Or at least that seems more intuitive to me.

.github/workflows/benchmark.yml
CarlWachter: jep that was a stupid formatting thing from hermit-bench, i've updated this to be properly relative

.github/benchmarks/general.yaml
mkroening: Just from looking at this file, it is not clear that this is relative to the parent directory (hermit-rs).
When I looked at this during the review, I thought it weird that we do something with the kernel's Cargo.toml here.

Changing this would make it immediately obvious that we refer to a Cargo.toml in the parent dir of this repo.
Following that thought, we could even think about making this `../../../Cargo.toml`, which would be relative to the file not relative to the current working directory.

In CI, we use

```yaml
    defaults:
      run:
        working-directory: kernel
```

to make all paths kernel-repo-relative. (Not sure if that works with nodejs-based actions, though).

mkroening (approved):
> All squashed and ready to go. Any last objections?

I have pushed a commit reformatting and removing some things, that should be unnecessary (I don't think the benchmarks require extra toolchains or targets).

If you could check that this still works, you can squash that into your commit and I will merge. 🥳

CarlWachter:
> > All squashed and ready to go. Any last objections?
> 
> I have pushed a commit reformatting and removing some things, that should be unnecessary (I don't think the benchmarks require extra toolchains or targets).
> 
> If you could check that this still works, you can squash that into your commit and I will merge. 🥳

just ran it through and it still works and is a little faster

PR #1727, opened 2025-05-21
Added `O_CLOEXEC` flag to OpenOptions

jounathaen:
The main goal of this is to avoid crashes if this flag is set. The flag itself doesn't have any meaning in an unikernel, as we don't have `execve`  or similar process control functionality. 
However, programs [using our newlib](https://github.com/hermit-os/newlib/blob/9ef4bd93cf9cf2cee8d5a05c87b117591dcb4a35/newlib/libc/posix/opendir.c#L85) set this flag and crashe at https://github.com/hermit-os/kernel/blob/d7203edd9df40e42097f3af3568659361f3ddead/src/syscalls/mod.rs#L334


jounathaen:
Update: Added doc comment

mkroening (approved):

PR #1732, opened 2025-05-27
perf(virtqueue): Use smallvec for UsedDeviceWritableBuffer

Gelbpunkt:
Currently, we were allocating a two-item `VecDeque` on the heap with talc for every call to `{PackedVq, SplitVq}::try_recv`, which is horribly inefficient. The reason I presume this was done before was to have cheap `pop_front` operations, but since they usually only contain two items, moving around all the items of a smallvec is super cheap anyways and we can skip the conversion and allocation.

We may want to replace this with a smallvec equivalent of `VecDeque` in the future, but there is no such struct yet and we currently don't really need it either.

Results in a ~15% improvement in TCP bandwidth for me.

### DIFF
--- a/src/drivers/virtio/virtqueue/mod.rs
+++ b/src/drivers/virtio/virtqueue/mod.rs
@@ -13,7 +13,6 @@ pub mod packed;
 pub mod split;
 
 use alloc::boxed::Box;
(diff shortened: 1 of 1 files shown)


mkroening (approved):
Thanks! :)

PR #1733, opened 2025-06-02
perf(virtqueue): Use smallvec for control descriptors

Gelbpunkt:
Haven't measured the differences yet, and I wouldn't exactly bet that 4 is the correct number here, though I think it is correct (since it is TX + RX combined which are 2 each).


mkroening (commented):
You worry about spilling to the heap here, right? Will this always be at most 4, or will it be more sometimes? We could use `heapless::Vec` to rule out spilling.

PR #1734, opened 2025-06-02
perf(virtqueue/packed): Don't box transfer tokens

Gelbpunkt:
This gets rid of the last talc allocation calls from `receive_packet` in netbench, everything else is now either on the stack or allocated with `DeviceAlloc`.

I haven't been able to measure a reliable improvement, but it should strictly be an improvement.


mkroening:
> everything else is now either on the stack or allocated with `DeviceAlloc`.

Just to clarify: That's only on the default path, though, right? For example, if we had an indirect table at `TransferToken::ctrl_desc`, that would still be talc-allocated.

mkroening (approved):
Thanks! :)

Issue #1589, opened 2025-02-07
`getdents64` should be stateful


mkroening:
Currently, Hermit's [`getdents64`] is not stateful. This means that when reading a directory, the caller needs to provide a buffer that is large enough to hold all directory members. A buffer that can only hold a few members returns `EINVAL`.

Instead, [`getdents64`] should be stateful. This means that when reading a directory, it is fine if the caller provides a buffer that can only hold a few members of the directory. [`getdents64`] is then called again and again to return the next members until it returns 0 (end of directory). Only if the buffer is too small to hold the next member, `EINVAL` is returned.

This issue shows itself in newlib, which [allocates a buffer of size 512](https://github.com/bminor/newlib/blob/newlib-4.5.0/newlib/libc/posix/opendir.c#L57-L58) and then calls [`getdents64`] [until it returns 0 (end of directory)](https://github.com/bminor/newlib/blob/newlib-4.5.0/newlib/libc/posix/readdir.c#L58-L63). Currently, this loops endlessly, since we always start from the beginning of the directory and never return 0.

The issue has been worked around in rust-std by [growing the buffer until it is big enough to hold all directory members](https://github.com/rust-lang/rust/blob/1.84.1/library/std/src/sys/pal/hermit/fs.rs#L535-L539). This is not ideal for performance, since we eagerly copy all the directory members instead of copying lazily while iterating over `ReadDir`. This is also not ideal for memory consumption, since the buffer we allocate can grow arbitrarily large instead of staying constant.

The state should be exposed by and resettable via [`lseek`] (which is used in newlib's [`seekdir`] and [`rewinddir`]).

[`getdents64`]: https://www.man7.org/linux/man-pages/man2/getdents64.2.html
[`lseek`]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/lseek.html
[`seekdir`]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/seekdir.html
[`rewinddir`]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/rewinddir.html

jounathaen:
Just thinking out loud: An alternative approach might be to keep it stateless in the kernel and pass the starting position for reading to the systemcall as well. The newlib would then keep track of the state. (We never claimed that our syscall interface is POSIX compatible, after all)

PR #1738, opened 2025-06-03
made `readdir` stateful

jounathaen:
This fixes https://github.com/hermit-os/kernel/issues/1589 by saving the read state in the backends of the `sys_getdents64` syscall.

Currently, implemented only for RAMFS and FUSE - mostly because I couldn't find an example on how to use the ROMFS.

In general, I'm not a huge fan of this, as we basically reimplement the old POSIX interface instead of designing a better interface ourselves, but this avoids complications and breaking changes in the standard libraries (Rust and Newlib).

Notable Quirks/Issues:
- We don't have Inode numbers for RAMFS. All entries have Inode nr. `1`. This is unlikely to be a problem because I can't imagine an application utilizing Inodes being seriously executed in an unikernel.
- As the RAMFS is currently based on paths, we can't tell the application which file type a direntry has. This is also unlikely to cause a problem, as Linux has introduced this field "recently" (21 years ago) and states that "All applications must properly handle a return of DT_UNKNOWN." (https://linux.die.net/man/2/getdents64) 
- This would be a great use-case for a unit test, but it is quite tricky to get file system related things working in a unit test. This is because the file system relies on async code, but the async executor also tries to execute other maintenance tasks which are not present in a unit-test environment.


src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
+			.finish()
+	}
+}
 
 #[hermit_macro::system]
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn sys_getdents64(
 	fd: FileDescriptor,
 	dirp: *mut Dirent64,
+	// rename: bytes/size/length
```
mkroening: I'd argue for keeping the name consistent with [Linux](https://www.man7.org/linux/man-pages/man2/getdents64.2.html):

```c
ssize_t getdents64(int fd, void dirp[.count], size_t count);
```

src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
 	pub d_reclen: u16,
 	/// File type
-	pub d_type: u8,
+	pub d_type: FileType,
 	/// Filename (null-terminated)
 	pub d_name: PhantomData<c_char>,
 }
+impl core::fmt::Debug for Dirent64 {
+	fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+		let cstr = unsafe { CStr::from_ptr((&raw const self.d_name).cast::<c_char>()) };
```
mkroening: This is not sound, since the API allows creating the struct without a filename. This would then read out of bounds.

src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
+			4 => Ok(Self::DtDir),
+			6 => Ok(Self::DtBlk),
+			8 => Ok(Self::DtReg),
+			10 => Ok(Self::DtLnk),
+			12 => Ok(Self::DtSock),
+			14 => Ok(Self::DtWht),
+			_ => Err("inv. Discriminant"),
+		}
+	}
+}
```
mkroening: We already have these constants in [`fs::FileType`](https://github.com/hermit-os/kernel/blob/v0.11.0/src/fs/mod.rs#L302-L313), which is also what we use in [`sys/dirent.h`](https://github.com/hermit-os/newlib/blob/9ef4bd93cf9cf2cee8d5a05c87b117591dcb4a35/newlib/libc/sys/hermit/include/sys/dirent.h#L9-L17).

src/syscalls/mod.rs
```diff
@@ -324,6 +323,7 @@ pub unsafe extern "C" fn sys_opendir(name: *const c_char) -> FileDescriptor {
 	if let Ok(name) = unsafe { CStr::from_ptr(name) }.to_str() {
 		crate::fs::opendir(name).unwrap_or_else(|e| -num::ToPrimitive::to_i32(&e).unwrap())
 	} else {
+		error!("sys opendir invalid name");
```
mkroening: I am not sure if kernel logging is useful here, since this would be an application error, not a kernel error and the application should log and report errors.

Regardless, this is probably a separate commit or even PR, right?

src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
+	}
+}
+
+#[repr(C)]
+#[derive(Clone, Copy)]
 pub struct Dirent64 {
 	/// 64-bit inode number
 	pub d_ino: u64,
-	/// 64-bit offset to next structure
+	/// Field without meaning. Kept for BW compatibility.
```
mkroening, line 554: > d_off is a filesystem-specific value with no specific meaning to user space, though on older filesystems it used to be the distance from the start of the directory to the start of the next linux_dirent

([getdents(2)](https://www.man7.org/linux/man-pages/man2/getdents64.2.html))

I am fine with making this field meaningless, I just wanted to be conscious about this. Could there be a benefit of using absolute offsets in user space instead of iterating incrementally using `d_reclen`?

src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
+			10 => Ok(Self::DtLnk),
+			12 => Ok(Self::DtSock),
+			14 => Ok(Self::DtWht),
+			_ => Err("inv. Discriminant"),
+		}
+	}
+}
+
+#[repr(C)]
+#[derive(Clone, Copy)]
```
mkroening: This was `Clone` and `Clone` before, and we don't have to change that in this PR, but is it not bad that this copies all the metadata including name length but not the name itself?

src/syscalls/mod.rs
```diff
@@ -549,71 +549,88 @@ pub extern "C" fn sys_lseek(fd: FileDescriptor, offset: isize, whence: i32) -> i
...
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn sys_getdents64(
 	fd: FileDescriptor,
 	dirp: *mut Dirent64,
+	// rename: bytes/size/length
 	count: usize,
 ) -> i64 {
+	debug!("getdents for fd {fd:?} - count: {count}");
 	if dirp.is_null() || count == 0 {
+		error!("dirp is null");
```
mkroening: I am not sure if this makes sense as kernel error log. It is an application error, not a kernel error.

src/fs/mod.rs
```diff
@@ -129,8 +129,9 @@ impl DirectoryReader {
 
 #[async_trait]
 impl ObjectInterface for DirectoryReader {
-	async fn readdir(&self) -> io::Result<Vec<DirectoryEntry>> {
-		Ok(self.0.clone())
+	async fn getdents(&self, _buf: &mut [core::mem::MaybeUninit<u8>]) -> io::Result<usize> {
+		let _ = self.0.clone(); // Dummy instruction to avoid warning for the moment
+		unimplemented!("")
```
mkroening:
```suggestion
		let _ = &self.0; // Dummy statement to avoid warning for the moment
		unimplemented!()
```

mkroening: @stlankes, I think we can also remove `DirectoryReader` completely and panic on `sys_opendir`, which is deprecated, right?

src/fs/mem.rs
```diff
@@ -387,19 +391,71 @@ impl MemDirectoryInterface {
...
+					d_type: FileType::DtUnknown, // TODO: Proper filetype
+					d_name: PhantomData {},
+				});
+				let nameptr = core::ptr::from_mut(&mut (*(target_dirent)).d_name).cast::<u8>();
+				core::ptr::copy_nonoverlapping(
+					name.as_bytes().as_ptr().cast::<u8>(),
+					nameptr,
+					namelen,
+				);
+				nameptr.add(namelen).write_bytes(0, 1); // zero termination
```
mkroening:
```suggestion
				nameptr.add(namelen).write(0); // zero termination
```

src/fs/mem.rs
```diff
@@ -387,19 +391,71 @@ impl MemDirectoryInterface {
...
 		}
+		Ok(ret)
+	}
 
-		Ok(entries)
+	/// lseek for a directory entry is the equivalent for seekdir on linux. But on Hermit this is
+	/// logically the same operation, so we can just use the same fn in the backend.
+	/// Any other offset than 0 is not supported. (Mostly because it doesn't make any sense, as
+	/// userspace applications have no way of knowing valid offsets)
+	async fn lseek(&self, offset: isize, _whence: SeekWhence) -> io::Result<isize> {
```
mkroening: User space could know offsets through seeking without moving the head, which returns the current position. Not sure if it makes sense to expose that now, though. I don't really understand newlib's implementation of telldir, seekdir and rewinddir yet.

Can we check that `whence` is `SeekWhence::Set`?

src/fs/fuse.rs
```diff
@@ -840,44 +848,63 @@ impl ObjectInterface for FuseDirectoryHandle {
...
+					d_type: (dirent.type_ as u8).try_into().unwrap(),
+					d_name: PhantomData {},
+				});
+				let nameptr = core::ptr::from_mut(&mut (*(target_dirent)).d_name).cast::<u8>();
+				core::ptr::copy_nonoverlapping(
+					dirent.name.as_ptr().cast::<u8>(),
+					nameptr,
+					dirent.namelen as usize,
+				);
+				nameptr.add(dirent.namelen as usize).write_bytes(0, 1); // zero termination
```
mkroening:
```suggestion
				nameptr.add(dirent.namelen as usize).write(0); // zero termination
```

mkroening (changes requested):
Cool! I have a few comments. :D

src/syscalls/mod.rs
jounathaen: I think it is fine to declare it meaningless. I mean, the whole purpose of this PR is to make the offset an internally stored information, right?

I think there are better designs for the whole interface, but if we go for a different approach, we probably would also not base that one on `Dirent64`.

src/syscalls/mod.rs
jounathaen: True. I changed it into an `unsafe` fn in case it will be needed for debugging purposes later. But we can also just throw it out.

src/syscalls/mod.rs
jounathaen: Ack. Created doc comment instead.

src/fs/mem.rs
jounathaen: Good catch!

src/syscalls/mod.rs
mkroening: We could do it with a `debug` function similar to this:

https://github.com/hermit-os/kernel/blob/b15388dcb32a0f257f195122912381762ca630f1/src/arch/x86_64/mm/paging.rs#L398-L425

jounathaen: Done!
Still debatable if we want to keep it or not.

mkroening (approved):
Thanks! :)

PR #1739, opened 2025-06-04
perf(virtqueue): Use swap_remove when feasible

Gelbpunkt:
`swap_remove` does not need to copy data around, while `remove` emits `memmove` calls that are costly. We do care about preserving ordering, but if there are less than 3 items in the smallvec, `swap_remove` behaves identically. I confirmed that this removes any traces of `memmove` in `receive_packet`.


mkroening (approved):
Thanks! :)

PR #1741, opened 2025-06-06
perf(virtio-net): Reuse allocations for virtqueue packet buffers

Gelbpunkt:
This lets us get rid of the remaining DeviceAlloc allocations in receive_packet.


mkroening:
Since the performance impact of this is not measurable, let's close this for now. :)

PR #1756, opened 2025-06-12
Minor enhancement of strace return value formatting

jounathaen:
Probably very opinionated, but the strace output of a printf syscall looks like:

```
sys_write(fd = 1, buf = 0x1d35a90, len = 13) = hello world!
13
sys_malloc(size = 8, align = 1) = 0x400000009f80
```
At first sight, I got pretty confused what the `13` means.

This PR changes the output to, hopefully increasing the readability a little:
```
sys_write(fd = 1, buf = 0x1d35a90, len = 13) hello world!
-> 13
sys_malloc(size = 8, align = 1) -> 0x400000009f80
```

### DIFF
--- a/hermit-macro/src/system.rs
+++ b/hermit-macro/src/system.rs
@@ -116,7 +116,7 @@ fn emit_func(mut func: ItemFn, sig: &ParsedSig) -> Result<ItemFn> {
 		.map(|ident| format!("{ident} = {{:?}}"))
 		.collect::<Vec<_>>()
 		.join(", ");
-	let strace_format = format!("{}({input_format}) = ", sig.ident);
+	let strace_format = format!("{}({input_format} ", sig.ident);
 
 	let block = func.block;
 	func.block = parse_quote! {{
@@ -127,7 +127,7 @@ fn emit_func(mut func: ItemFn, sig: &ParsedSig) -> Result<ItemFn> {
 			print!(#strace_format, #(#input_idents),*);
 			let ret = #block;
 			#[cfg(feature = "strace")]
-			println!("{ret:?}");
+			println!(") = {ret:?}");
 			ret
 		}
 	}};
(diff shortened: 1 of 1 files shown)


hermit-macro/src/system.rs
```diff
@@ -116,7 +116,7 @@ fn emit_func(mut func: ItemFn, sig: &ParsedSig) -> Result<ItemFn> {
 		.map(|ident| format!("{ident} = {{:?}}"))
 		.collect::<Vec<_>>()
 		.join(", ");
-	let strace_format = format!("{}({input_format}) = ", sig.ident);
+	let strace_format = format!("{}({input_format}) ", sig.ident);
```
mkroening:
```suggestion
	let strace_format = format!("{}({input_format} ", sig.ident);
```

hermit-macro/src/system.rs
```diff
@@ -127,7 +127,7 @@ fn emit_func(mut func: ItemFn, sig: &ParsedSig) -> Result<ItemFn> {
 			print!(#strace_format, #(#input_idents),*);
 			let ret = #block;
 			#[cfg(feature = "strace")]
-			println!("{ret:?}");
+			println!("-> {ret:?}");
```
mkroening:
```suggestion
			println!(") = {ret:?}");
```

mkroening (changes requested):
I like the change, but I'd like to keep close to Linux's strace format.

We also need to update the macro's unit tests.

Edit: I compared this using

```c
#include <unistd.h>

int main(int argc, char *argv[]) {
    write(STDERR_FILENO, "123\n", 4);
    write(STDERR_FILENO, "456\n", 4);
}
```

which gives

```console
...
write(2, "123\n", 4123
)                    = 4
write(2, "456\n", 4456
)                    = 4
...
```

jounathaen:
Makes sense - updated!

mkroening (approved):

PR #1761, opened 2025-06-13
Implement fstat for GenericStdout/GenericStdIn

jounathaen:
Currently, `sys_fstat` for fd 0, 1, and 2 fails with `EINVAL`. This PR fixes this.

### DIFF
--- a/src/fd/stdio.rs
+++ b/src/fd/stdio.rs
@@ -152,6 +152,14 @@ impl ObjectInterface for UhyveStdin {
 	async fn isatty(&self) -> io::Result<bool> {
 		Ok(true)
 	}
+
+	async fn fstat(&self) -> io::Result<FileAttr> {
+		let attr = FileAttr {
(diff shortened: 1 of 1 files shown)


mkroening (commented):
Cool!

First off, this PR makes a test case (https://github.com/hermit-os/hermit-c/pull/34) pass, which is good.
In Linux, the mode of the FD depends on the environment, though. Locally (interactively), I also get S_IFCHR for stdin, but in CI, we get S_IFIFO for all three FDs.

So I am fine with merging this as is, but I just wanted to hear your thoughts on whether we should default to S_IFCHR or S_IFIFO (which might probably make Python behave worse interactively, but maybe better when run non-interactively? also S_IFIFO but might be more correct maybe since we are talking about a serial port here?) or whether we should make this dependent on the environment somehow (maybe a kernel arg?, I can't think of anything else we can inspect here).

PR #1762, opened 2025-06-13
Fixed interactive console output beeing stalled until newline

jounathaen:
The console currently doesn't echo any characters entered until a newline, which makes interactive prompts a little hard to use.

Side note:
image

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -21,13 +21,15 @@ impl Console {
 		}
 	}
 
+	/// Writes a buffer to the console.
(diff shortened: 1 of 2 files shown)


mkroening (changes requested):
Cool! :)

We can use `Console::flush` inside `Console::write` then, right?

I'd personally also prefer sticking with `write` instead of `write_buffered` for alignment with [`std::io::Write`](https://doc.rust-lang.org/stable/std/io/trait.Write.html), that we eventually migrate to.

PR #1776, opened 2025-06-17
refactor(arch/x86_64): Dynamically allocate XSAVE area

Gelbpunkt:
The real size of the XSAVE area depends on the enabled features in XRC0. To avoid having to hardcode the order of the save regions, we can instead rely on the raw-cpuid crate's APIs that expose the required size of the area.

However, we cannot *fully* rely on this and just zero out the memory, since some fields are required to be initialized with non-zero values. Therefore we need to keep the legacy header definition and manually set those.

Additionally, this commit introduces support for XSAVEC and XSAVEOPT, which are both faster and more efficient than XSAVE.


src/arch/x86_64/kernel/processor.rs
```diff
@@ -133,135 +140,58 @@ pub struct XSaveLegacyRegion {
...
-	pub buffer_base: u64,
-	pub buffer_size: u32,
-	pub filters: u32,
-	pub saved_event_record: [u64; 4],
-	pub event_counter: [u32; 16],
-}
+impl FPUState {
+	pub fn new() -> Self {
+		let xsave_size = if supports_xsave()
+			&& let Some(extended_state_info) = CpuId::new().get_extended_state_info()
```
mkroening: Would it make sense to unwrap instead? Can this ever be none on a CPU that supports XSAVE?

src/arch/x86_64/kernel/processor.rs
```diff
@@ -133,135 +140,58 @@ pub struct XSaveLegacyRegion {
...
+			extended_state_info.xsave_area_size_enabled_features() as usize
+		} else {
+			size_of::<XSaveLegacyRegion>()
+		};
 
-#[repr(C)]
-pub struct XSaveBndregs {
-	pub bound_registers: [u8; 4 * 16],
-}
+		info!("XSAVE area size: {xsave_size}");
```
mkroening: Is this generally useful or should we downgrade this to debug?

```suggestion
		debug!("XSAVE area size: {xsave_size}");
```

src/arch/x86_64/kernel/processor.rs
```diff
@@ -133,135 +140,58 @@ pub struct XSaveLegacyRegion {
...
-#[repr(C)]
-pub struct XSaveBndcsr {
-	pub bndcfgu_register: u64,
-	pub bndstatus_register: u64,
-}
+		// Allocate a 64-byte aligned Vec
+		let mut xsave_area = unsafe {
+			//let n_units = xsave_size.div_ceil(size_of::<AlignToSixtyFour>());
+			let n_units = xsave_size;
+			let maybe_uninit_aligned = Box::<[AlignToSixtyFour]>::new_zeroed_slice(n_units);
```
mkroening: Does this not allocate way too much memory? ~64 times as much?

src/lib.rs
```diff
@@ -16,6 +16,7 @@
 #![feature(maybe_uninit_slice)]
 #![feature(maybe_uninit_write_slice)]
 #![feature(never_type)]
+#![feature(new_zeroed_alloc)]
```
mkroening: What do we need this for? `Box::new_zeroed_slice_in` should only require `allocator_api`.

src/arch/x86_64/kernel/processor.rs
Gelbpunkt: Yes, we should just unwrap here.

src/arch/x86_64/kernel/processor.rs
Gelbpunkt: Debugging leftover, the line above should be commented out

src/lib.rs
Gelbpunkt: Also a leftover from when I was doing manual allocations

Gelbpunkt: Hmm, nevermind.

```rs
error[E0658]: use of unstable library feature `new_zeroed_alloc`
   --> src/arch/x86_64/kernel/processor.rs:174:31
    |
174 |             let maybe_uninit_aligned = Box::<[AlignToSixtyFour]>::new_zeroed_slice(n_units);
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #129396 <https://github.com/rust-lang/rust/issues/129396> for more information
    = help: add `#![feature(new_zeroed_alloc)]` to the crate attributes to enable
    = note: this compiler was built on 2025-06-05; consider upgrading it if it is out of date
```

See https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.new_zeroed_slice

mkroening: Ah, that makes sense. I was looking at `new_zeroed_slice_in`.

Something like `vec![AlignToSixtyFour([0; 64]); n_units].into_boxed_slice()` should optimize the same, though.

mkroening (approved):
Thanks! :)

PR #1778, opened 2025-06-18
timespec: use i64 for tv_nsec as in POSIX

jounathaen:
Reference: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/time.h.html

> The <time.h> header shall declare the timespec structure, which shall include at least the following members:
> ```
> time_t  tv_sec    Whole seconds.
> long    tv_nsec   Nanoseconds [0, 999999999].
> ```

Without this fix, we observed funny time measurements using newlib.

### DIFF
--- a/src/fs/fuse.rs
+++ b/src/fs/fuse.rs
@@ -431,15 +431,15 @@ impl From<fuse_attr> for FileAttr {
 			st_blocks: attr.blocks.try_into().unwrap(),
 			st_atim: timespec {
 				tv_sec: attr.atime as time_t,
-				tv_nsec: attr.atimensec as i32,
+				tv_nsec: i64::from(attr.atimensec),
 			},
 			st_mtim: timespec {
 				tv_sec: attr.mtime as time_t,
-				tv_nsec: attr.mtimensec as i32,
+				tv_nsec: i64::from(attr.mtimensec),
 			},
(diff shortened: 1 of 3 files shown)


mkroening:
After this is merged and released, we need to propagate this to hermit-abi and the libc crate as well.

mkroening (approved):
Thanks! :)

Fortunately, this change is backwards compatible (since we only support little-endian machines at the moment) and can be merged immediately. :D

jounathaen:
I'm not familiar with the structure of the libc crate, but it looks to me as if we already use `c_long` there, which is `i64`.

mkroening:
> it looks to me as if we already use `c_long` there

[We don't](https://github.com/rust-lang/libc/blob/0.2.174/src/hermit.rs#L113). But we should wait for a kernel release before opening a PR there.

mkroening:
I diffed the resulting ELF file from 8751d7b38b73841c6114e826de6370c9c894901f with its parent using `elf_diff` and this change really does change inlining considerably. The inlining changes probably trigger a race condition on GitHub runners most of the time. Since I can't reproduce the hang whatsoever, let's merge this after https://github.com/hermit-os/kernel/pull/1678, which should work around triggering the underlying issue.

PR #1780, opened 2025-06-19
fix(virtio): allow exact buffer size in send_packet

CarlWachter:
The assert in the `send_packet` function for the length of the buffer, only allowed buffers which were larger than the message. I think we should allow exactly sized buffers.


mkroening (approved):
@cagatay-y, this was introduced in 1749ab4e547662c49eacdcd1490b63706de35978. I think this PR is good but would like your approval too. :)

Issue #1775, opened 2025-06-17
Panics on virtiofs host errors


mkroening:
When we perform a file system operation that turns out invalid on the host (for example `lseek`ing to before the beginning of a file), virtiofsd [replies with an error](https://gitlab.com/virtio-fs/virtiofsd/-/blob/v1.13.2/src/server.rs#L1551-1571) containing the host errno.

In this case, `RspHeader` does not contain the `OutStruct`, but only the `fuse_out_header` containing the error number. We currently fail at receiving these structs because [`UsedDeviceWritableBuffer::pop_front_downcast`](https://github.com/hermit-os/kernel/blob/v0.11.0/src/drivers/virtio/virtqueue/mod.rs#L426-L428) returns `None` since `fuse_out_header` itself is often smaller than the whole `RspHeader`, which often expects a non-ZST `OutStruct`. We then panic due to unwrapping `None` in [`VirtioFsDriver::send_command`](https://github.com/hermit-os/kernel/blob/v0.11.0/src/drivers/fs/virtio_fs.rs#L192).

Instead, we should have a way of succeeding in receiving a `fuse_out_header` without its `OutStruct` to then properly handle the error, which we [already try to do](https://github.com/hermit-os/kernel/blob/v0.11.0/src/fs/fuse.rs#L712-L714) but never get there.

cagatay-y:
We had a similar situation in #1337. If that approach is satisfactory, I can implement a similar fix for lseek (and any other commands that have a similar issue). If _the out struct not existing if the error code is not 0_ is the case for all commands, we can change the struct definition to make the field optional.

PR #1785, opened 2025-06-23
fix(fuse): check for the existence of op_header

cagatay-y:
If the request fails, the host does not write an op_header and code that assumes its exitence fails. To prevent this, only assume the existence of the out_header and check its error field before continuing with the rest of the response. Fixes https://github.com/hermit-os/kernel/issues/1775.

I am marking the PR as draft as it contains commits from the open PR #1730.

### DIFF
--- a/src/fs/fuse.rs
+++ b/src/fs/fuse.rs
@@ -13,7 +13,6 @@ use core::{future, mem};
 use async_lock::Mutex;
 use async_trait::async_trait;
 use fuse_abi::linux::*;
-use zerocopy::FromBytes;
 
 use crate::alloc::string::ToString;
 #[cfg(not(feature = "pci"))]
@@ -404,16 +403,14 @@ pub(crate) mod ops {
 		const OP_CODE: fuse_opcode = fuse_opcode::FUSE_LOOKUP;
 		type InStruct = ();
 		type InPayload = CString;
-		type OutStruct = ();
-		// Lookups return [fuse_entry_out] only when there actually is a result. For this reason,
-		// it is not part of the header (since all other headers are always there).
-		type OutPayload = [u8];
+		type OutStruct = fuse_entry_out;
+		type OutPayload = ();
 	}
 
 	impl Lookup {
 		pub(crate) fn create(name: CString) -> (Cmd<Self>, u32) {
 			let cmd = Cmd::with_cstring(FUSE_ROOT_ID, (), name);
-			(cmd, size_of::<fuse_entry_out>().try_into().unwrap())
+			(cmd, 0)
 		}
 	}
 }
@@ -578,12 +575,7 @@ fn lookup(name: CString) -> Option<u64> {
 		.lock()
 		.send_command(cmd, rsp_payload_len)
 		.ok()?;
-	if rsp.headers.out_header.error == 0 {
-		let entry_out = fuse_entry_out::ref_from_bytes(rsp.payload.as_ref().unwrap()).unwrap();
-		Some(entry_out.nodeid)
-	} else {
-		None
-	}
+	Some(rsp.headers.op_header.nodeid)
 }
 
 fn readlink(nid: u64) -> io::Result<String> {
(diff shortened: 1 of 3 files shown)


src/drivers/fs/virtio_fs.rs
```diff
@@ -195,9 +203,38 @@ impl FuseInterface for VirtioFsDriver {
...
+				let payload = transfer_result.used_recv_buff.pop_front_vec();
+				Ok(Rsp { headers, payload })
+			} else {
+				// "However, if the reply is an error reply (i.e., error is set), then no further payload data should be sent,
+				// independent of the request." (fuse man page)
+
+				Err(FuseError::IOError(
+					io::Error::try_from_primitive(-headers.out_header.error)
+						.unwrap_or(io::Error::EIO),
+				))
```
mkroening: I think we should ignore the error value for now because it's based on the host system and always return EIO.

Not sure what a good solution would be eventually. Convert from host errnos? How do we know the host? Why is this not a problem for Linux? Just because most of the time both the host and the guest are Linux?

src/fs/fuse.rs
```diff
@@ -534,9 +535,15 @@ where
...
-	op_header: O::OutStruct,
+// The generic H parameter allows us to handle RspHeaders with their op_header
+// potenitally uninitialized. After checking for the error code in the out_header,
+// the object can be transmuted to one with an initialized op_header.
+pub(crate) struct RspHeader<O: ops::Op, H = <O as ops::Op>::OutStruct> {
+	pub out_header: fuse_out_header,
+	op_header: H,
+	// We use PhantomData<fn(O)> instead of PhantomData<O> to circumvent compiler
+	// errors about O not being Send.
+	_phantom: PhantomData<fn(O)>,
```
mkroening:
```suggestion
	_phantom: PhantomData<O::OutStruct>,
```

src/drivers/fs/virtio_fs.rs
```diff
@@ -195,9 +203,38 @@ impl FuseInterface for VirtioFsDriver {
...
-		let payload = transfer_result.used_recv_buff.pop_front_vec();
-		Ok(Rsp { headers, payload })
+		let (dyn_headers, written_header_len) =
+			transfer_result.used_recv_buff.pop_front_raw().unwrap();
+		let headers = dyn_headers
+			.downcast::<MaybeUninit<RspHeader<O, MaybeUninit<O::OutStruct>>>>()
+			.unwrap();
+		if written_header_len < size_of::<fuse_out_header>() {
+			Err(VirtqError::IncompleteWrite.into())
+		} else {
```
mkroening: Can we do early returns instead of this huge expression with increasing indentation (here and further down)?

```suggestion
		if written_header_len < size_of::<fuse_out_header>() {
			return Err(VirtqError::IncompleteWrite.into());
		}
```

mkroening (changes requested):
Thanks! A few comments. :)

src/drivers/fs/virtio_fs.rs
cagatay-y: Some of the existing error handling code in the FUSE driver<sup>[1],[2],[3],[4]</sup> do this conversion, so I think we should at least return the raw error code to make the conversion possible where appropriate.

Existing FUSE documentation is mostly Linux specific so I don't know if this is specified somewhere for other platforms but I would expect other platforms to follow the Linux error codes as they would be implementing a Linux API. Of course, it may not make sense to design the code around that expectation. It would also be very annoying if the kernel misreported the cause of an error.

[1]: https://github.com/hermit-os/kernel/blob/f1f7d93f45b3789caa832aee871dfc740c085944/src/fs/fuse.rs#L1037
[2]: https://github.com/hermit-os/kernel/blob/f1f7d93f45b3789caa832aee871dfc740c085944/src/fs/fuse.rs#L1064
[3]: https://github.com/hermit-os/kernel/blob/f1f7d93f45b3789caa832aee871dfc740c085944/src/fs/fuse.rs#L1106
[4]: https://github.com/hermit-os/kernel/blob/f1f7d93f45b3789caa832aee871dfc740c085944/src/fs/fuse.rs#L1190

cagatay-y: Turns out, it was me who introduced the conversion from the host error code in the cases 1, 2 and 3 with the commit 71bc6297a7b0c404d41add24c4eddbe691b409de. Only example 4 is from [a commit](https://github.com/hermit-os/kernel/blame/fbb5789508286aa5a019f8b9e5371c3ecbf108d2/src/fs/fuse.rs#L1429) not authored by me.

mkroening (approved):
Thanks! :)

PR #1796, opened 2025-06-27
feat(logging): print time and target

mkroening:
This PR makes our logger log the time (unconditionally) as well as the log target (via a feature).


jounathaen (approved):
Funny, I think I hacked something very similar a few weeks ago, but I can't find it any more :grinning:

PR #1807, opened 2025-06-30
fix(aarch64/interrupts): clippy::unwrap_or_default

mkroening:
So this Clippy suggestion is only shown when using build std???

This shows the suggestion:

```bash
cargo clippy --target=aarch64-unknown-none-softfloat -Zbuild-std=core,alloc
```

This does not:

```bash
cargo clippy --target=aarch64-unknown-none-softfloat
```

How peculiar!


mkroening:
I opened an upstream issue for this at https://github.com/rust-lang/rust-clippy/issues/15183.

PR #1809, opened 2025-07-01
fix(arch/aarch64): Restore and save FPU state

Gelbpunkt:
Currently, we don't save or restore any FPU state on aarch64, while we do so on x86_64 and RISC-V.

ARMv8-A mandates NEON support, which adds 32 128-bit vector registers v0-v32. Of these, only the lower 64 bits of v8-v15 (=d8-d15) are callee-saved and therefore need to be taken care of by the kernel.

For implementing this, I chose a similar approach to our x86_64 code: We disable floating point when switching to a new task by setting FPEN in CPACR_EL1 to 00 so that any use of v0-v32 triggers a trap. We then handle the trap to enable floating point again and switch the FPU owner lazily.

There are multiple extensions to ARMv8-A that add more things we need to save, but for now let's at least implement the baseline properly.


src/arch/aarch64/kernel/processor.rs
```diff
@@ -7,6 +8,128 @@ use hermit_sync::{Lazy, OnceCell, without_interrupts};
...
+				"ldr d10, [{fpu_state}, {off_d10}]",
+				"ldr d11, [{fpu_state}, {off_d11}]",
+				"ldr d12, [{fpu_state}, {off_d12}]",
+				"ldr d13, [{fpu_state}, {off_d13}]",
+				"ldr d14, [{fpu_state}, {off_d14}]",
+				"ldr d15, [{fpu_state}, {off_d15}]",
+				"msr fpcr, {intermediate}",
+				"ldr {intermediate}, [{fpu_state}, {off_fpcr}]",
+				"msr fpsr, {intermediate}",
+				"ldr {intermediate}, [{fpu_state}, {off_fpsr}]",
```
mkroening: `mrs <Xt>, fpcr` stores fpcr in Xt.
`msr fcpr, <Xt>` stores Xt in fpcr.

You got this mixed up here.

src/arch/aarch64/kernel/processor.rs
```diff
@@ -7,6 +8,128 @@ use hermit_sync::{Lazy, OnceCell, without_interrupts};
...
+				"str d10, [{fpu_state}, {off_d10}]",
+				"str d11, [{fpu_state}, {off_d11}]",
+				"str d12, [{fpu_state}, {off_d12}]",
+				"str d13, [{fpu_state}, {off_d13}]",
+				"str d14, [{fpu_state}, {off_d14}]",
+				"str d15, [{fpu_state}, {off_d15}]",
+				"mrs {intermediate}, fpcr",
+				"str {intermediate}, [{fpu_state}, {off_fpcr}]",
+				"mrs {intermediate}, fpsr",
+				"str {intermediate}, [{fpu_state}, {off_fpsr}]",
```
mkroening, line 114: Here too.

mkroening (changes requested):
Thanks! :)

This is great, but I think the assembly code is not quite correct.

src/arch/aarch64/kernel/processor.rs
Gelbpunkt: I think this is correct then. Just the order of the `msr` and `ldr` in this method is wrong. During `restore`, we want to load the saved value of `fpcr` to the intermediate GPR and then use `msr fpcr, <Xt>` to then move the value from the intermediate GPR to `fpcr`.

The `save` is implemented correctly - we use `mrs` to store the fpcr value in a GPR and then store the GPR value in memory.

mkroening (approved):
Thanks! :)

PR #1812, opened 2025-07-03
bench: publish benchmark data in separate workflow

CarlWachter:
Addresses issue with the current benchmarking implementation, that led to it always benchmarking main. This is done by splitting it into two workflows, one of which runs the benchmarks and the other publishes to the website or PR. They pass data via artifacts.

Since the publish_benchmark workflow is triggered by workflow_run, it will not show any data in this PR. (Tested in https://github.com/CarlWachter/kernel/pull/20)


.github/workflows/publish_benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+          - benchmark-matrix-name: Misc
+
+    env:
+      GH_TOKEN: ${{ github.token }}
+
+    steps:
+    - name: Set up GitHub CLI
+      run: |
+        sudo apt-get update
+        sudo apt-get install gh -y
```
mkroening: We can skip this, since this runs on GitHub-hosted runners.

```suggestion
```

.github/workflows/publish_benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  publish_benchmarks:
+    runs-on: ubuntu-latest
+    environment: benchmark
+
+    permissions:
+      pull-requests: write
+      actions: read
+      contents: read
```
mkroening: Is `read` not the default?


```suggestion
```

.github/workflows/publish_benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+      uses: hermit-os/github-action-benchmark@main
+      with:
+        tool: "hermit-bench"
+        output-file-path: results.json
+        github-token: ${{ secrets.HERMIT_BENCH_GITHUB_TOKEN }}
+        benchmark-data-dir-path: .
+        gh-repository: github.com/hermit-os/hermit-bench
+        comment-always: true
+        benchmark-matrix-name: ${{ matrix.benchmark-matrix-name }}
+        auto-push: true
```
mkroening: This should not push unconditionally but only on pushes to main, right?

mkroening (changes requested):

CarlWachter:
I split it into two different condiftional steps for pushing and commenting on PRs. (tested on https://github.com/CarlWachter/kernel/pull/21)

mkroening (approved):

PR #1815, opened 2025-07-04
fix(mm): reintroduce explicit virtual to physical address translation for device memory

mkroening:
Before https://github.com/hermit-os/kernel/pull/1609, https://github.com/hermit-os/kernel/pull/1669, and https://github.com/hermit-os/kernel/pull/1670, we were mapping frames and _flushing_ on each device memory allocation, which was expensive.
Those PRs changed the initial memory mappings by ensuring an identity mapping of all physical memory.
That allowed for (basically) no-op device memory allocation while reducing TLB pressure.

Before https://github.com/hermit-os/kernel/pull/1712, we were walking the page table on every virtual to physical address translation for device communication, which was not costly, but still slower than necessary.

This PR reintroduces explicit virtual to physical address translation when handling device memory, but does so while avoiding the performance pitfalls of the past.
We now have the option to map the complete physical memory a second time at an offset.
This is currently done on `cfg!(careful)` to ensure all devices use the device allocator not only for memory management but also for address translation.
Eventually, this allows us to mark one of the mappings as private and the other as public, making flexible device communication as cheap as possible.

### DIFF
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -27,9 +27,18 @@ unsafe impl Allocator for DeviceAlloc {
 		assert!(layout.align() <= BasePageSize::SIZE as usize);
 		let size = layout.size().align_up(BasePageSize::SIZE as usize);
 
-		let virt_addr = ptr.as_ptr().expose_provenance();
-		let phys_addr = PhysAddr::from(virt_addr);
+		let phys_addr = self.phys_addr_from(ptr.as_ptr());
 
 		super::physicalmem::deallocate(phys_addr, size);
+	}
+}
+
+impl DeviceAlloc {
+	/// Returns the physical address of `ptr`.
+	///
+	/// The address is only correct if `ptr` has been allocated by this allocator.
+	pub fn phys_addr_from<T: ?Sized>(&self, ptr: *mut T) -> PhysAddr {
+		let addr = u64::try_from(ptr.expose_provenance()).unwrap();
+		PhysAddr::new(addr)
 	}
 }
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -18,7 +18,7 @@ unsafe impl Allocator for DeviceAlloc {
 
 		let phys_addr = super::physicalmem::allocate(size).unwrap();
 
-		let ptr = ptr::with_exposed_provenance_mut(phys_addr.as_usize());
+		let ptr = self.ptr_from(phys_addr);
 		let slice = ptr::slice_from_raw_parts_mut(ptr, size);
 		Ok(NonNull::new(slice).unwrap())
 	}
@@ -34,6 +34,12 @@ unsafe impl Allocator for DeviceAlloc {
 }
 
 impl DeviceAlloc {
+	/// Returns a pointer corresponding to `phys_addr`.
+	pub fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
+		let addr = phys_addr.as_usize();
+		ptr::with_exposed_provenance_mut(addr)
+	}
+
 	/// Returns the physical address of `ptr`.
 	///
 	/// The address is only correct if `ptr` has been allocated by this allocator.
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -36,6 +36,7 @@ unsafe impl Allocator for DeviceAlloc {
 
 impl DeviceAlloc {
 	/// Returns a pointer corresponding to `phys_addr`.
+	#[inline]
 	pub fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
 		let addr = phys_addr.as_usize() + self.phys_offset().as_usize();
 		ptr::with_exposed_provenance_mut(addr)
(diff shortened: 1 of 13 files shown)


src/mm/device_alloc.rs
```diff
@@ -27,9 +28,33 @@ unsafe impl Allocator for DeviceAlloc {
...
+	///
+	/// The address is only correct if `ptr` has been allocated by this allocator.
+	pub fn phys_addr<T: ?Sized>(&self, ptr: *mut T) -> usize {
+		ptr.expose_provenance() - self.phys_offset()
+	}
+
+	/// Returns the physical address offset.
+	///
+	/// This device allocator expects the complete physical memory to be mapped device-readable at this offset.
+	pub fn phys_offset(&self) -> usize {
```
zyuiop: Make the function constant? I think virtualmem::kernel_heap_end() can also be made const, so it'd probably reduce the overhead a bit.

mkroening: A function being `const` should not affect runtime performance. It only allows this to be run in `const` contexts, as far as I know.

This is a good candidate for `#[inline]`, though, which should do what you want. Thanks! 👍

zyuiop: Yes you're right, my hope was to evaluate this offset at compile time - all parameters involved are known at that time. But using a constant outside of the implementation block would help more to achieve that.

src/mm/device_alloc.rs
```diff
@@ -27,9 +28,33 @@ unsafe impl Allocator for DeviceAlloc {
...
-		let phys_addr = PhysAddr::from(virt_addr);
+		let phys_addr = PhysAddr::from(self.phys_addr(ptr.as_ptr()));
 
 		super::physicalmem::deallocate(phys_addr, size);
 	}
 }
+
+impl DeviceAlloc {
+	/// Returns a pointer corresponding to `phys_addr`.
+	pub fn ptr<T>(&self, phys_addr: usize) -> *mut T {
```
jounathaen: Would `PhysAddr` instead of `usize` be an option here?

mkroening: Yes. I am a bit hesitant because I find the address types a bit unwieldy and did not want to use `VirtAddr` here.

Many places (not all, though) want a physical address anyway so it makes sense to change both methods to work with `PhysAddr`. 👍

src/mm/device_alloc.rs
```diff
@@ -27,9 +28,33 @@ unsafe impl Allocator for DeviceAlloc {
...
+		ptr::with_exposed_provenance_mut(phys_addr + self.phys_offset())
+	}
+
+	/// Returns the physical address of `ptr`.
+	///
+	/// The address is only correct if `ptr` has been allocated by this allocator.
+	pub fn phys_addr<T: ?Sized>(&self, ptr: *mut T) -> usize {
```
sarahspberrypi: Changing the return type of this to `PhysAddr` might lead to less type conversions (e.g., in `kernel/src/drivers/net/gem.rs`) down the road. Is there an explicit reason for using a `usize` here? And if so, why not return a `u64` or a `u32`?

mkroening: Yeah, I'm already working on that (see https://github.com/hermit-os/kernel/pull/1815#discussion_r2189754410). 👍

src/mm/device_alloc.rs
```diff
@@ -27,9 +28,35 @@ unsafe impl Allocator for DeviceAlloc {
...
+	/// Returns a pointer corresponding to `phys_addr`.
+	pub fn ptr<T>(&self, phys_addr: PhysAddr) -> *mut T {
+		let addr = phys_addr.as_usize() + self.phys_offset().as_usize();
+		ptr::with_exposed_provenance_mut(addr)
+	}
+
+	/// Returns the physical address of `ptr`.
+	///
+	/// The address is only correct if `ptr` has been allocated by this allocator.
+	pub fn phys_addr<T: ?Sized>(&self, ptr: *mut T) -> PhysAddr {
```
jounathaen: I'd suggest renaming this fn to `phys_add_of`.
As it now is - e.g., `DeviceAlloc.phys_addr(foobar)` - it is not very clear that it is a translation and not an allocation of a phys_addr or something similar.

src/mm/device_alloc.rs
```diff
@@ -27,9 +28,35 @@ unsafe impl Allocator for DeviceAlloc {
...
-		let phys_addr = PhysAddr::from(virt_addr);
+		let phys_addr = self.phys_addr(ptr.as_ptr());
 
 		super::physicalmem::deallocate(phys_addr, size);
 	}
 }
+
+impl DeviceAlloc {
+	/// Returns a pointer corresponding to `phys_addr`.
+	pub fn ptr<T>(&self, phys_addr: PhysAddr) -> *mut T {
```
jounathaen: As argued below, maybe `ptr_from` or similar makes the code more readable.

src/mm/physicalmem.rs
```diff
@@ -43,6 +46,21 @@ pub unsafe fn init_frame_range(frame_range: PageRange) {
 		.map(|addr| PhysAddr::new(addr.try_into().unwrap()))
 		.for_each(paging::identity_map::<IdentityPageSize>);
 
+	if DeviceAlloc.phys_offset() != VirtAddr::zero() {
```
jounathaen, line 50:
```suggestion
	// Remap physical memory to device allocator's offset
	if DeviceAlloc.phys_offset() != VirtAddr::zero() {
```

jounathaen (approved):

src/mm/device_alloc.rs
mkroening: I changed it to `phys_addr_from`. 👍

sarahspberrypi (approved):

PR #1818, opened 2025-07-06
feat: log page tables

mkroening:


mkroening:
> Why does the CI fail? The PR itself look good for me.

The benchmark CI fails because printing the whole page table is slow, and we get a timeout. Printing on `info` was only enabled for testing in https://github.com/hermit-os/kernel/commit/b15388dcb32a0f257f195122912381762ca630f1, though. I'll remove that commit (so we only print the page tables on `debug`) and merge.

PR #1819, opened 2025-07-07
feat: add generic ioctl support

zyuiop:
This PR adds an ioctl handler to file descriptor objects, such as sockets, and a helper to register any ioctl handler at a particular file system path.

Some Linux-specific crates communicate with the kernel via the use of an ioctl file. Support of ioctls may provide an easy path when porting such libraries to work with hermit, without increasing too much the size of the ABI for too specific features.

### DIFF
--- a/src/fd/mod.rs
+++ b/src/fd/mod.rs
@@ -332,6 +332,15 @@ pub(crate) trait ObjectInterface: Sync + Send {
 		Err(Errno::Nosys)
 	}
 
+	/// Handles an ioctl
+	fn handle_ioctl(
+		&mut self,
+		_cmd: crate::fs::ioctl::IoCtlCall,
+		_argp: *mut core::ffi::c_void,
+	) -> io::Result<()> {
+		Err(Errno::Nosys)
+	}
+
 	/// `isatty` returns `true` for a terminal device
 	async fn isatty(&self) -> io::Result<bool> {
 		Ok(false)
--- a/src/fd/mod.rs
+++ b/src/fd/mod.rs
@@ -7,6 +7,7 @@ use core::pin::pin;
 use core::task::Poll::{Pending, Ready};
 use core::time::Duration;
 
+use bitfield_struct::bitfield;
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 #[cfg(feature = "net")]
 use smoltcp::wire::{IpEndpoint, IpListenEndpoint};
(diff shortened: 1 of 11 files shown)


src/fs/ioctl.rs
```diff
@@ -0,0 +1,96 @@
...
+use alloc::sync::Arc;
+use alloc::vec::Vec;
+use core::fmt::Debug;
+
+use crate::fd::{AccessPermission, ObjectInterface};
+use crate::fs::{NodeKind, VfsNode};
+use crate::io;
+
+#[derive(Copy, Clone)]
+pub struct IoCtlCall(pub u32);
```
mkroening: We _could_ use [`bitfield-struct`](https://docs.rs/bitfield-struct) for these bitfield types, which is already a transitive dependency.

We should link to the reference header ([this one?](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/ioctl.h?h=v6.15)) in the doc comments of this struct.

mkroening (changes requested):

zyuiop:
@mkroening it should be good now, I rebased and my new tests pass :)

mkroening (approved):
This looks good to me and I am personally fine with using `ioctl` for this.

@stlankes, @jounathaen, are you fine with us pivoting to `ioctl` for controlling drivers?

jounathaen:
I'm not a fan of "pivoting to ioctl for controlling drivers", but we can merge it for enhanced compatibility. I'd like to see a feature flag for this, but this can be added later once we have researched the impacts of this. (We should create an issue for this)

src/syscalls/mod.rs
```diff
@@ -624,30 +625,17 @@ pub unsafe extern "C" fn sys_writev(fd: FileDescriptor, iov: *const iovec, iovcn
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn sys_ioctl(
 	fd: FileDescriptor,
-	cmd: i32,
+	cmd: c_ulong,
```
zyuiop: Check if we can remove this change to preserve the ABI

zyuiop:
I have rebased the PR, in case there is still some interest for this!

jounathaen (approved):

src/fd/mod.rs
```diff
@@ -332,6 +332,15 @@ pub(crate) trait ObjectInterface: Sync + Send {
...
 
+	/// Handles an ioctl
+	fn handle_ioctl(
+		&mut self,
+		_cmd: crate::fs::ioctl::IoCtlCall,
+		_argp: *mut core::ffi::c_void,
+	) -> io::Result<()> {
+		Err(Errno::Nosys)
+	}
+
```
mkroening: Let's call these `ioctl` instead of `handle_ioctl`.

src/fs/ioctl.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::fd::Fd;
+use crate::fs::{NodeKind, VfsNode};
+use crate::io;
+
+/// Encoding for an IOCTL command, as done in the Linux Kernel.
+///
+/// See [relevant kernel header](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/ioctl.h?h=v6.15) for reference.
+///
+/// The goal of this interface is to easily support linux applications that communicate via IOCTL,
+/// so linux compatibility is an intended and explicit goal here.
```
mkroening: Nits:

```suggestion
/// Encoding for an `ioctl` command, as done in Linux.
///
/// For details, see [asm-generic/ioctl.h].
///
/// The goal of this interface is to easily support Linux applications that communicate via `ioctl`,
/// so Linux compatibility is an intended and explicit goal here.
///
/// [asm-generic/ioctl.h]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/asm-generic/ioctl.h?h=v7.2
```

src/fd/socket/mod.rs
```diff
@@ -20,3 +20,27 @@ pub const DEFAULT_BACKLOG: i32 = 128;
...
+#[macro_export]
+macro_rules! socket_handle_ioctl {
+	($this: ident, $cmd: ident, $argp: ident) => {{
+		use crate::errno::Errno;
+		use crate::fd::{IoCtlCall, StatusFlags};
+
+		const FIONBIO: IoCtlCall = IoCtlCall::from_bits(0x8008_667eu32);
+
+		if $cmd == FIONBIO {
+			let value: [u8; 4] = [$argp[0], $argp[1], $argp[2], $argp[3]];
```
zyuiop, line 33: Technically, the previous definition for FIONBIO requests a 8 bytes value, but just in case let's just consume four

PR #1822, opened 2025-07-07
x86-64: enable PCIe enhanced configuration support

zyuiop:


stlankes (changes requested):
Looks good for me. Maybe some debug messages would help to see that PCIe is detected. How do test your PR?

zyuiop:
Hi, I'll add some debug messages. For now, I tested on VMs in which standard PCI device discovery was not available, so I confirmed this works.

mkroening:
> For now, I tested on VMs in which standard PCI device discovery was not available, so I confirmed this works.

I think Stefan is interested in some example QEMU flags for running this. :D

zyuiop:
I don't know of any flag combination that disables support for legacy PCI support, besides enabling CPU hardware protection features.

PR #1826, opened 2025-07-08
feat: add rudimentary implementation for some fs system calls

zyuiop:


stlankes (approved):
You have just to remove the merge conflicts.

mkroening:
Before merging I'd like to think about whether it makes sense to implement cwd functionality implemented more natively. This is a great starting point for reference, for sure, though, and maybe it would still make sense to merge this first. 🤔

src/fs/fuse.rs
```diff
@@ -270,6 +270,76 @@ pub(crate) mod ops {
...
+			const FATTR_GID = 1 << 2;
+			const FATTR_SIZE = 1 << 3;
+			const FATTR_ATIME = 1 << 4;
+			const FATTR_MTIME = 1 << 5;
+			const FATTR_FH = 1 << 6;
+			const FATTR_ATIME_NOW = 1 << 7;
+			const FATTR_MTIME_NOW = 1 << 8;
+			const FATTR_LOCKOWNER = 1 << 9;
+			const FATTR_CTIME = 1 << 10;
+			const FATTR_KILL_SUIDGID = 1 << 11;
```
mkroening: Can we use the constants from `fuse_abi::linux` (already imported) on the right side here?

zyuiop:
Good point, I have opened two PRs on newlib :)

src/syscalls/mod.rs
```diff
@@ -345,8 +346,139 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
 	}
 }
 
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_getcwd(buf: *mut c_char, size: usize) -> i32 {
```
mkroening: I think we should use the POSIX definition here, returning a pointer and setting errno in case of failure.
That way, we can just forward the implementation in Newlib but can also use this function in rust-std.

```suggestion
pub unsafe extern "C" fn sys_getcwd(buf: *mut c_char, size: usize) -> *mut c_char {
```

zyuiop: I did a thing, tell me if the new `system(errno)` macro would have enabled a better implementation!

mkroening: This is fine, thanks! :)

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+pub unsafe extern "C" fn sys_faccessat(
+	dirfd: FileDescriptor,
+	name: *const c_char,
+	_mode: i32,
+	flags: i32,
+) -> i32 {
+	let access_option = AccessOption::from_bits_truncate(flags);
+	if access_option.bits() != flags {
+		return -i32::from(Errno::Inval);
+	}
```
mkroening: We can use `let else` here:

```suggestion
	let Some(access_option) = AccessOption::from_bits(flags) else {
		return -i32::from(Errno::Inval);
	};
```

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+	} else {
+		warn!("faccessat with directory relative to fd is not implemented!");
+		return -i32::from(Errno::Nosys);
+	};
+
+	if let Err(e) = stat {
+		return -i32::from(e);
+	}
+
+	let Ok(stat) = stat else { unreachable!() };
```
mkroening: We can `match` here too.

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+	}
+}
+
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_access(name: *const c_char, flags: i32) -> i32 {
+	let access_option = AccessOption::from_bits_truncate(flags);
+	if access_option.bits() != flags {
+		return -i32::from(Errno::Inval);
+	}
```
mkroening: We can use `let else` here too.

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+	let Ok(name) = unsafe { CStr::from_ptr(name) }.to_str() else {
+		return -i32::from(Errno::Inval);
+	};
+	let stat = crate::fs::read_lstat(name);
+
+	if let Err(e) = stat {
+		return -i32::from(e);
+	}
+
+	let Ok(stat) = stat else { unreachable!() };
```
mkroening: We can `match` here too.

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+	}
+}
+
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_fchmod(fd: FileDescriptor, mode: u32) -> i32 {
+	let access_permission = AccessPermission::from_bits_truncate(mode);
+	if access_permission.bits() != mode {
+		return -i32::from(Errno::Inval);
+	}
```
mkroening: We can use `let else` here too.

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+	if size == 0 {
+		return error(Errno::Inval);
+	}
+
+	let cwd = fs::get_cwd();
+	if let Err(e) = cwd {
+		return error(e);
+	}
+
+	let Ok(cwd) = cwd else { unreachable!() };
```
mkroening: We can use `match` here:

```suggestion
    let cwd = match cwd {
        Ok(cwd) => cwd,
        Err(err) => return -i32::from(err),
    };
```

src/syscalls/mod.rs
```diff
@@ -345,8 +347,152 @@ pub unsafe extern "C" fn sys_open(name: *const c_char, flags: i32, mode: u32) ->
...
+		return error(Errno::Noent);
+	};
+
+	if (cwd.count_bytes() + 1) > size {
+		return error(Errno::Range);
+	}
+
+	if buf.is_null() {
+		return cwd.into_raw();
+	}
```
mkroening: I don't think we should leak memory here, since this cannot be freed through `malloc`.

Let's do that in newlib if at all. Sorry for the back and forth. :D

mkroening (approved):
Last round of comments :D

src/syscalls/mod.rs
zyuiop: Let's not do it at all, since it's not mandatory and even discouraged in POSIX

PR #1827, opened 2025-07-08
feat: align frame range to (likely) page size when initializing physical memory

zyuiop:
This is an old commit on my local branch. 

I think this was linked to frequent issues with pages being too big when trying to de-allocate, since the physical free list contains 4KiB frames that may be part of larger 2MiB frames (and therefore not entirely free).

### DIFF
--- a/src/mm/physicalmem.rs
+++ b/src/mm/physicalmem.rs
@@ -7,7 +7,7 @@ use memory_addresses::{PhysAddr, VirtAddr};
 
 #[cfg(target_arch = "x86_64")]
 use crate::arch::mm::paging::PageTableEntryFlagsExt;
-use crate::arch::mm::paging::{self, HugePageSize, PageSize, PageTableEntryFlags};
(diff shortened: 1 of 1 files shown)


mkroening:
> I think this was linked to frequent issues with pages being too big when trying to de-allocate, since the physical free list contains 4KiB frames that may be part of larger 2MiB frames (and therefore not entirely free).

I don't really understand the issue yet. Why is it an issue to manage free physical memory on 4k-granularity?

The CI failure is due to allocating zero-sized layouts when setting up the heap. While this should be handled gracefully by the free list (tracked in https://github.com/mkroening/free-list/issues/3), we should also do something about that here if we want to continue with this PR.

zyuiop:
I think this is part of the larger discussion about the memory layout we inherit from UEFI.
I think we should close this.

For context, I think the issue is that we obtain memory mapped as large pages from firmware, but we mark memory as free in the free list in small pages. This is mostly fine, but if you get a page from the free-list and try to change its flags, it may fail, and you may not be able to safely split the parent page, especially if the free list starts at the end of a large-page that is also used for other stuff in the kernel.

PR #1830, opened 2025-07-12
publish_benchmark: merge results into one file

CarlWachter:
This fixes an issue with data being pushed to hermit-bench/gh-pages multiple times due to a race condition between the two publishing matrixes. This will make the plots look normal again.

This is also directly linked to changes in github-action-benchmark so until this is merged, the publish benchmarks pipeline will not work, as this change required a breaking change.


CarlWachter:
Also throwing the fix for PR comments from forks in here, as it affects the same file. That one just grabs the proper PR number via cli, since apparently the github.event.workflow_run.pull_requests[0].number is not passed if the PR is from a fork

mkroening (approved):

PR #1834, opened 2025-07-15
feat(drivers/net): Add loopback network driver

Gelbpunkt:
The idea is that in the immediate future, this driver will be the fallback network driver. Once we support multiple interfaces, it should always be the driver for the loopback interface.


src/drivers/net/loopback.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+impl NetworkDriver for LoopbackDriver {
+	fn get_mac_address(&self) -> [u8; 6] {
+		// This matches Linux' behavior
+		[0; 6]
+	}
+
+	fn get_mtu(&self) -> u16 {
+		// This matches Linux' behavior
```
mkroening: Did you not say Linux does `u16::MAX + 1`?

src/drivers/net/loopback.rs
```diff
@@ -0,0 +1,78 @@
...
+		u16::MAX
+	}
+
+	fn receive_packet(
+		&mut self,
+	) -> Option<(
+		crate::executor::device::RxToken,
+		crate::executor::device::TxToken,
```
mkroening:
```suggestion
		RxToken,
		TxToken,
```

src/drivers/net/mod.rs
```diff
@@ -56,3 +64,29 @@ pub(crate) fn mtu() -> u16 {
...
+	))] {
+		pub(crate) use crate::drivers::pci::get_network_driver;
+	} else {
+		use hermit_sync::InterruptTicketMutex;
+
+		pub(crate) fn get_network_driver() -> Option<&'static InterruptTicketMutex<loopback::LoopbackDriver>> {
+			Some(&loopback::LOOPBACK)
+		}
+	}
+}
```
mkroening, line 92: Can you make this refactoring (as well as the changes in `executor`) a separate commit that comes before the loopback driver addition?

mkroening:
> Can do, but that would depend on this PR getting merged and vice-versa, I think

Yes, please! :)

PR #1839, opened 2025-07-22
feat(drivers/net): Make virtio-net optional, take 2

Gelbpunkt:
Depends on #1834

### DIFF
--- a/src/drivers/net/mod.rs
+++ b/src/drivers/net/mod.rs
@@ -1,19 +1,19 @@
-#[cfg(all(target_arch = "riscv64", feature = "gem-net"))]
+#[cfg(all(target_arch = "riscv64", feature = "gem-net", not(feature = "pci")))]
 pub mod gem;
-#[cfg(all(
-	not(any(
-		all(target_arch = "riscv64", feature = "gem-net", not(feature = "pci")),
-		all(target_arch = "x86_64", feature = "rtl8139"),
-	)),
-	feature = "net"
-))]
+#[cfg(not(any(
+	all(target_arch = "riscv64", feature = "gem-net", not(feature = "pci")),
+	all(target_arch = "x86_64", feature = "rtl8139"),
+	feature = "virtio-net",
+)))]
 pub mod loopback;
 #[cfg(all(target_arch = "x86_64", feature = "rtl8139"))]
 pub mod rtl8139;
-#[cfg(not(all(target_arch = "x86_64", feature = "rtl8139")))]
+#[cfg(all(
+	not(all(target_arch = "riscv64", feature = "gem-net", not(feature = "pci"))),
+	not(all(target_arch = "x86_64", feature = "rtl8139")),
+	feature = "virtio-net",
+))]
 pub mod virtio;
-
-use core::str::FromStr;
 
 use smoltcp::phy::ChecksumCapabilities;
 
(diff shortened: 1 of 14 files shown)


src/drivers/net/mod.rs
```diff
@@ -44,42 +44,60 @@ pub(crate) trait NetworkDriver: Driver {
...
 	}
 }
 
+#[cfg(any(
+	all(target_arch = "riscv64", feature = "gem-net", not(feature = "pci")),
+	all(target_arch = "x86_64", feature = "rtl8139"),
+	feature = "virtio-net",
+))]
+pub(crate) use mtu::*;
+
```
mkroening: Can we do something like this instead?

```rust
#[cfg(..)]
pub(crate) fn mtu() -> u16 {
	const DEFAULT_IP_MTU: u16 = 1500;

	const DEFAULT_MTU: u16 = DEFAULT_IP_MTU + 14;

	if let Some(my_mtu) = hermit_var!("HERMIT_MTU") {
		u16::from_str(&my_mtu).unwrap()
	} else {
		DEFAULT_MTU
	}
}
```

src/drivers/pci.rs
```diff
@@ -336,6 +350,7 @@ pub(crate) fn print_information() {
 
 #[allow(clippy::large_enum_variant)]
 #[allow(clippy::enum_variant_names)]
+#[non_exhaustive]
```
mkroening, line 353: Is this related to the other changes or a separate change?

src/drivers/pci.rs
```diff
@@ -519,6 +537,7 @@ pub(crate) fn get_interrupt_handlers() -> HashMap<InterruptLine, InterruptHandle
 
 #[cfg(all(
 	not(all(target_arch = "x86_64", feature = "rtl8139")),
+	feature = "virtio-net",
 	any(feature = "tcp", feature = "udp")
 ))]
```
mkroening, line 542: Do we still need `any(feature = "tcp", feature = "udp")` here and at the other places?

src/drivers/pci.rs
```diff
@@ -29,18 +34,26 @@ use crate::console::IoDevice;
 use crate::drivers::console::{VirtioConsoleDriver, VirtioUART};
 #[cfg(feature = "fuse")]
 use crate::drivers::fs::virtio_fs::VirtioFsDriver;
-#[cfg(any(feature = "tcp", feature = "udp"))]
+#[cfg(all(
+	any(feature = "tcp", feature = "udp"),
+	any(
+		all(target_arch = "x86_64", feature = "rtl8139"),
```
mkroening, line 40: Why do we need both `feature = "rtl8139"` and `target_arch = "x86_64"`? Would the former not be enough?

mkroening (changes requested):
Thanks! :)

src/drivers/pci.rs
Gelbpunkt: This is related because it fixes a ton of the warnings when matching on the variants with an unreachable `_` arm. I figured this would be a better solution to that instead of allowing a clippy lint in every method that matches on the enum

src/drivers/pci.rs
Gelbpunkt: Yes, network support is currently still dependent on either of those being enabled. The next step after this PR would be to turn that into `feature = "net"` instead

src/drivers/pci.rs
Gelbpunkt: IIRC we don't emit a compile error if `rtl8139` is enabled on other architectures, so this currently makes sense. I think it's also nicer to be able to compile all feature combinations on other architectures, even if they have no effect.

mkroening: I'm very torn on this. With the kernel, we don't really have the issue of feature unification, since we should not appear multiple times in the dependency graph. That's why I would tend to favor a compilation error, since the RTL driver on other architectures is a configuration error that should be handled.
But since we don't emit a compiler error right now, we should not change that as part of this PR. 👍

mkroening (approved):

PR #1854, opened 2025-07-29
fix(aarch64): PL031 read is always little endian

Gelbpunkt:
Fixes the RTC reading on aarch64_be.


src/arch/aarch64/kernel/systemtime.rs
```diff
@@ -45,7 +45,7 @@ fn rtc_read(off: usize) -> u32 {
 		);
 	}
 
-	value
+	value.to_le()
```
mkroening: This should do the same but is clearer about the intention:

```suggestion
	u32::from_le(value)
```

PR #1857, opened 2025-07-30
add test related files to gitignore

cagatay-y:
Add files that are needed to run or are generated during tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
+/hermit-loader-*
+/opensbi-*-rv-bin
+/shared
 /target
+/vhostqemu.pid


.gitignore
```diff
@@ -1 +1,5 @@
 /target
+/opensbi-*-rv-bin
+/hermit-loader-*
+/shared
+/vhostqemu\.pid
```
mkroening: What function does `\.` have here?

cagatay-y: None. I thought '.' matched any character but it doesn't. I removed it now.

mkroening (approved):

PR #1862, opened 2025-08-09
feat: Add profiling profile

Gelbpunkt:
Profile for a release build with debuginfo. Very useful for rftrace


Cargo.toml
```diff
@@ -209,3 +209,7 @@ exclude = [
 
 [patch.crates-io]
 x86_64 = { git = "https://github.com/rust-osdev/x86_64.git" }
+
+[profile.release-debug]
```
mkroening: Should we call this `profiling` to be more descriptive?

```suggestion
[profile.profiling]
```

Cargo.toml
```diff
@@ -209,3 +209,7 @@ exclude = [
...
+inherits = "release"
+debug = true
```
mkroening: [The Cargo Book](https://doc.rust-lang.org/cargo/reference/profiles.html#debug) and [Nicholas Nethercote](https://nnethercote.github.io/perf-book/profiling.html#debug-info) suggest using `"line-tables-only"`, instead.

```suggestion
debug = "line-tables-only"
```

PR #1872, opened 2025-08-18
feat: print git version, features, build time

mkroening:
This should be very useful when looking at Hermit boot logs in the future.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,6 +83,10 @@ mod synch;
 pub mod syscalls;
 pub mod time;
 
+mod built_info {
+	include!(concat!(env!("OUT_DIR"), "/built.rs"));
+}
+
 hermit_entry::define_abi_tag!();
 
 #[cfg(target_os = "none")]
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -182,6 +186,24 @@ fn boot_processor_main() -> ! {
...
+			format_args!("")
+		} else {
+			format_args!(" (opt-level={})", built_info::OPT_LEVEL)
+		};
+
+		info!("Git version: {git_version}{dirty}{non_optimized}");
+	}
+	info!("Enabled features: {}", built_info::FEATURES_LOWERCASE_STR);
+	info!("Built on {}", built_info::BUILT_TIME_UTC);
+
```
sarahspberrypi, line 206: Why not also print `built_info::RUSTC` or `built_info::RUSTC_VERSION`? I think that may also be helpful, especially after a new std release

src/lib.rs
```diff
@@ -83,6 +83,10 @@ mod synch;
 pub mod syscalls;
 pub mod time;
 
+pub mod built_info {
```
jounathaen: Probably it doesn't matter, but is this mod being `pub` instead of `pub (crate)` relevant?

src/lib.rs
```diff
@@ -182,6 +186,24 @@ fn boot_processor_main() -> ! {
...
+			""
+		};
+
+		let non_optimized = if built_info::OPT_LEVEL == "3" {
+			format_args!("")
+		} else {
+			format_args!(" (opt-level={})", built_info::OPT_LEVEL)
+		};
+
+		info!("Git version: {git_version}{dirty}{non_optimized}");
```
jounathaen:
```suggestion
		let opt_level = if built_info::OPT_LEVEL == "3" {
			format_args!("")
		} else {
			format_args!(" (opt-level={})", built_info::OPT_LEVEL)
		};

		info!("Git version: {git_version}{dirty}{opt_level}");
```

Or similar. A little confusing at the moment.

src/lib.rs
mkroening: Made it private 👍

src/lib.rs
mkroening: Ah, yes. This got lost when changed this. :D

src/lib.rs
mkroening: This is not the userspace rustc, which would be interesting due to changing std bindings, but the kernel rustc, which is already included in the git commit.

Printing the std version or the Newlib version is interesting but would be more complex and can be done separately.

sarahspberrypi: Is there any way to print this rustc version then?

Issue #1870, opened 2025-08-13
Issue with fpu scheduler on macos/aarch64


m4b:
I tested two applications on with uhyve, and I think there's some issue with fpu_owner in the scheduler in kernel.

E.g., here is stdin app:

```
[    0.001090][0][DEBUG] Open /proc/version, OpenOption(O_RDWR | O_CREAT), AccessPermission(S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH)
[    0.001106][0][DEBUG] Open file /proc/version with OpenOption(O_RDWR | O_CREAT)
[    0.001804][0][DEBUG] Try to initialize fuse filesystem
[    0.001817][0][INFO] Try to initialize uhyve filesystem
[    0.001842][0][INFO] Mounting uhyve filesystem at /root
[    0.001879][0][DEBUG] Mounting /root
[    0.002006][0][DEBUG] Setting argv as: []
[    0.002015][0][DEBUG] Setting envv as: []
[    0.002053][0][DEBUG] Receive interrupt 1
[    0.002075][0][INFO] Jumping into application
[    0.002108][0][WARN] Unable to read entropy! Fallback to a naive implementation!
[    0.002130][0][DEBUG] Switching FPU owner from task 0 to 1
[    0.002261][0][DEBUG] Create EventFd 0, EventFlags(EFD_NONBLOCK | EFD_CLOEXEC)
[    0.002617][0][DEBUG] sys_socket: domain 3, type 18433, protocol 0
Error: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }
[    0.003833][0][DEBUG] Exit program with error code 1!
[    0.003844][0][INFO] Number of interrupts
[    0.003858][0][INFO] [0][Reschedule]: 2
```

I saw it crash in the scheduler, but I can't seem to repro, it was a double borrow on the mutable section, so I modified it to print:

```
diff --git a/src/scheduler/mod.rs b/src/scheduler/mod.rs
index 8944058a..38512999 100644
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -664,7 +664,11 @@ impl PerCoreScheduler {
                                self.current_task.borrow().id
                        );

-                       self.fpu_owner.borrow_mut().last_fpu_state.save();
+                       if let Some(mut owner) = self.fpu_owner.try_borrow_mut().ok() {
+                               owner.last_fpu_state.save();
+                       } else {
[...]

PR #1890, opened 2025-08-22
fix(kernel_stack): don't transmute args via memory

mkroening:
This PR avoids transmuting kernel function args via memory and instead reinterprets register contents. If we do something similar for AArch64, this should fix https://github.com/hermit-os/kernel/issues/1870.

Depends on https://github.com/hermit-os/kernel/pull/1889.


src/arch/aarch64/kernel/kernel_stack.rs
```diff
@@ -1,96 +1,83 @@
...
+		"ret",
+	)
+}
 
 macro_rules! kernel_function_impl {
-	($kernel_function:ident($($arg:ident: $A:ident),*) { $($operands:tt)* }) => {
+	($kernel_function:ident($($arg:ident: $A:ident),*; $($z:ident: usize),*)) => {
 		/// Executes `f` on the kernel stack.
 		#[allow(dead_code)]
 		pub unsafe fn $kernel_function<R, $($A),*>(f: unsafe extern "C" fn($($A),*) -> R, $($arg: $A),*) -> R {
```
Gelbpunkt: I think we wanted to change the order of parameters here to match `call_with_stack`/`call_with_kernel_stack` to avoid having to shuffle around parameters in registers

mkroening: Can you check whether the shuffle is gone now? :)

Gelbpunkt: This looks amazing now:

```rs
000000000002d9c8 <sys_write>:
   2d9c8: d503201f      nop
   2d9cc: 10ffdde6      adr     x6, 0x2d588 <hermit::syscalls::sys_write::_sys_write::h643f290bb925f294>
   2d9d0: aa1f03e3      mov     x3, xzr
   2d9d4: aa1f03e4      mov     x4, xzr
   2d9d8: aa1f03e5      mov     x5, xzr
   2d9dc: 14004b4e      b       0x40714 <hermit::hermit::arch::aarch64::kernel::kernel_stack::call_with_kernel_stack::h4114d760a9371663>
```

No more register shuffles :partying_face:

Gelbpunkt (approved):

src/arch/aarch64/kernel/kernel_stack.rs
mkroening: Thanks for posting the assembly! I should have removed zeroing the registers too now. :D

Gelbpunkt: Interesting, seems like that now makes the compiler inline `call_with_kernel_stack` on aarch64:

```asm
000000000002d854 <sys_write>:
   2d854: f81f0ffe      str     x30, [sp, #-0x10]!
   2d858: d503201f      nop
   2d85c: 10ffd028      adr     x8, 0x2d260 <hermit::syscalls::sys_write::_sys_write::h643f290bb925f294>
   2d860: d3607c00      lsl     x0, x0, #32
   2d864: d50041bf      msr     SPSel, #0x1
   2d868: d63f0100      blr     x8
   2d86c: d50040bf      msr     SPSel, #0x0
   2d870: f84107fe      ldr     x30, [sp], #0x10
   2d874: d65f03c0      ret
```
I won't complain, this looks really good :D

Gelbpunkt: Hmm, wait what, what is the `lsl` doing there

Gelbpunkt: Woops, I had accidentally dropped this commit from my kernel tree. One second

Gelbpunkt:
```asm
000000000002c878 <sys_write>:
   2c878: d503201f      nop
   2c87c: 10ffdf66      adr     x6, 0x2c468 <hermit::syscalls::sys_write::_sys_write::h643f290bb925f294>
   2c880: 14009334      b       0x51550 <hermit::hermit::arch::aarch64::kernel::kernel_stack::call_with_kernel_stack::ha8571ee6b396905c>
```

Yeah that looks more like the expected assembly now

PR #1893, opened 2025-08-25
fix(socket): accept non-zero proto values

mkroening:


sarahspberrypi (approved):
Looks great, thanks so much :)

PR #1894, opened 2025-08-25
feat(aarch64): Add support for stdin

Gelbpunkt:
Bothered me a bit when running the stdin example...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -177,6 +177,7 @@ memory_addresses = { version = "0.2.3", default-features = false, features = [
 [target.'cfg(target_arch = "aarch64")'.dependencies]
 aarch64 = { version = "0.0.14", default-features = false }
 arm-gic = { version = "0.6" }
+arm-pl011-uart = { version = "0.3", default-features = false }
 semihosting = { version = "0.1", optional = true }
 memory_addresses = { version = "0.2.3", default-features = false, features = [
 	"aarch64",
--- a/src/arch/aarch64/kernel/serial.rs
+++ b/src/arch/aarch64/kernel/serial.rs
@@ -1,14 +1,21 @@
-use core::arch::asm;
+use alloc::collections::vec_deque::VecDeque;
+use core::ptr::NonNull;
 
+use arm_pl011_uart::{DataBits, Interrupts, LineConfig, Parity, StopBits, Uart, UniqueMmioPointer};
 use embedded_io::{ErrorType, Read, ReadReady, Write};
+use hermit_sync::{InterruptTicketMutex, Lazy};
 
 use crate::errno::Errno;
 
-pub(crate) struct SerialDevice {
-	pub addr: u32,
+static UART_DEVICE: Lazy<InterruptTicketMutex<UartDevice>> =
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -176,6 +176,7 @@ memory_addresses = { version = "0.2.3", default-features = false, features = [
 
 [target.'cfg(target_arch = "aarch64")'.dependencies]
 aarch64 = { version = "0.0.14", default-features = false }
+arm-pl011-uart = { version = "0.3", default-features = false }
 arm-gic = { version = "0.6" }
```
mkroening, line 180: Let's keep this sorted:

```suggestion
arm-gic = { version = "0.6" }
arm-pl011-uart = { version = "0.3", default-features = false }
```

src/arch/aarch64/kernel/serial.rs
```diff
@@ -26,41 +59,34 @@ impl ErrorType for SerialDevice {
...
+		let mut guard = UART_DEVICE.lock();
+
+		if guard.buffer.is_empty() {
+			Ok(0)
+		} else {
+			let min = buf.len().min(guard.buffer.len());
+
+			for (dst, src) in buf[..min].iter_mut().zip(guard.buffer.drain(..min)) {
+				*dst = src;
+			}
```
mkroening, line 71: Why not use `copy_from_slice` here like on x86?

https://github.com/hermit-os/kernel/blob/8c012c672d5dd044d43c4b0e611b457028e14227/src/arch/x86_64/kernel/serial.rs#L55-L64

Cargo.toml
mkroening: Let's keep this sorted:

```suggestion
arm-gic = { version = "0.6" }
arm-pl011-uart = { version = "0.3", default-features = false }
```

mkroening (changes requested):
Thanks, this looks mostly fine to me. :)

src/arch/aarch64/kernel/serial.rs
Gelbpunkt: x86 allocates an intermediate Vec, this version doesn't require an allocation

mkroening: Oh, I see.
So we should change x86 in a separate PR, right?

I was now thinking about using `VecDeque::as_slices`, but we could do even better by just forwarding all methods to the `Read` implementation for `VecDeque`, right?

mkroening: Ah no, there is no `embedded_io::Read` impl for `VecDeque`. We could start by adapting the Std impl for a newtype in the kernel and in the long term add a proper impl to embedded_io.

mkroening: But this is probably out of scope for this PR. I'll have a go at this afterwards. :)

mkroening (approved):
Thanks! :)

I'll press the button once https://github.com/hermit-os/kernel/pull/1909 is merged. :)

src/arch/aarch64/kernel/serial.rs
mkroening: I have opened https://github.com/hermit-os/kernel/pull/1915.

PR #1901, opened 2025-08-26
feat(virtio-net): VIRTIO_NET_F_GUEST_CSUM support

cagatay-y:
Tested with valid transmission across the same host and from another machine and invalid transmission.

### DIFF
--- a/src/drivers/net/virtio/mod.rs
+++ b/src/drivers/net/virtio/mod.rs
@@ -249,6 +249,9 @@ pub(crate) struct VirtioNetDriver<T = Init> {
 
 	pub(super) num_vqs: u16,
 	pub(super) irq: InterruptLine,
+	/// Describes for what protocols and in which directions, if any, the checksum
+	/// should be calculated in software. It is the complement of what is offloaded
+	/// to the hardware.
 	pub(super) checksums: ChecksumCapabilities,
 }
 


src/drivers/net/virtio/mod.rs
```diff
@@ -306,6 +309,105 @@ impl smoltcp::phy::TxToken for TxToken<'_> {
...
+		// We cannot benefit from the same host optimization but we've promised smoltcp to only pass frames
+		// that are validated so we need to do the validation ourselves.
+		match ethernet_frame.ethertype() {
+			smoltcp::wire::EthernetProtocol::Ipv4 => {
+				let Ok(ip_packet) = Ipv4Packet::new_checked(ethernet_frame.payload()) else {
+					return false;
+				};
+
+				// DATA_VALID only validates the outermost packet checksum, which is IPv4 in this case. Thus,
+				// it does not save us from validating the layer above IP.
```
mkroening: Is Ethernet not the outermost layer that has been validated? Ethernet frames contain IP packets. The code looks good, but the comment seems wrong.

mkroening (changes requested):
This is looking great, thanks! :)

Could you also put some documentation on `VirtioNetDriver::checksums`? Either in this PR or in a separate one is fine. As far as I understand, that field signals which kinds of checksums should be computed by smoltcp (None means nothing is computed), right? That is also the same way the struct is used in smoltcp, right?

Could you stick to [Conventional Commits](https://www.conventionalcommits.org) (replace "feature" with "feat")?

PR #1912, opened 2025-09-02
fix(executor): avoid possible deadlock

mkroening:
Although it should usually be avoided, the current system permits blocking on async tasks from an interrupt handler. This restores this behavior after it got broken in b735657457c4daa50f59e0f8b5096ea2b4234afd.

I am not very sure about the logic, so please review very thoroughly. :)

Extracted from https://github.com/hermit-os/kernel/pull/1910.


src/executor/mod.rs
```diff
@@ -233,18 +231,15 @@ where
...
-
 				// switch to another task
 				task_notify.wait(wakeup_time);
 
 				// restore default values
-				if let Ok(device) = crate::executor::network::NIC.lock().as_nic_mut() {
-					device.set_polling_mode(true);
-				}
+				crate::executor::network::NIC
+					.lock()
```
cagatay-y: Is this lock not an issue?

mkroening: Yes, and that is also the part I am unsure about. I have pushed an alternative. 🤔

Gelbpunkt: I'm not so sure what to do about this one. I think the way it is now is definitely wrong, but I'm unsure what the proper way of working around this would be. We can probably just avoid re-enabling the polling mode (i.e. drop this entirely) because it currently isn't done *anywhere*, so it can't be worse than it already is?

stlankes: Why do you think that this code is wrong?

Gelbpunkt: Unwrapping the `try_lock` will panic if the lock is being held by a different core, which may be extremely unlikely, but could in theory happen

stlankes: Ah, your are talking from line 243. Yes, the unwrapping should be removed

mkroening: Like this, then? This might be bad since it opens up the possibility of a deadlock in the unlikely case of blocking in an interrupt handler, but that's better than panicking in regular usage, right?

PR #1916, opened 2025-09-06
feat(sockets): stub UNIX sockets

sarahspberrypi:


src/syscalls/socket/mod.rs
```diff
@@ -448,6 +456,14 @@ impl From<SocketAddrV6> for sockaddr_in6 {
 	}
 }
 
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
```
mkroening, line 460: Like the other sockaddr structs:

```suggestion
#[derive(Debug, Default, Copy, Clone)]
```

src/syscalls/socket/mod.rs
```diff
@@ -448,6 +456,14 @@ impl From<SocketAddrV6> for sockaddr_in6 {
...
+pub struct sockaddr_un {
+	pub sun_len: u8,
+	pub sun_family: sa_family_t,
+	pub sun_path: [c_char; 108],
```
mkroening: Let's stick to BSD as we do with the other sockaddr structs.

```suggestion
	pub sun_path: [c_char; 104],
```

mkroening (commented):
Thanks, this looks good in general.

Can we get rid of all the early checks and returns? The address family should get handled later on already with something like `EINVAL`.

mkroening (approved):
Thanks! :)

PR #1917, opened 2025-09-08
feat(syscalls): stub `fchdir`

sarahspberrypi:


src/syscalls/mod.rs
```diff
@@ -386,6 +386,12 @@ pub unsafe extern "C" fn sys_getcwd(buf: *mut c_char, size: usize) -> *const c_c
 	buf
 }
 
+#[hermit_macro::system(errno)]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_fchdir(_fd: FileDescriptor) -> i32 {
```
mkroening: We should be able to make this safe:

```suggestion
pub extern "C" fn sys_fchdir(_fd: FileDescriptor) -> i32 {
```

mkroening (approved):

PR #1930, opened 2025-09-09
feat(xtask): kill servers externally

mkroening:
This PR aims to fix the spurious

```console
Error: Resource temporarily unavailable (os error 11)
```

in

```bash
cargo xtask ci rs --arch x86_64 --profile release --package mioudp --features hermit/udp,hermit/dhcpv4,hermit/virtio-net qemu --accel --sudo --devices virtio-net-pci
```

by avoiding early server shutdowns before the message is actually transmitted.

Depends on https://github.com/hermit-os/kernel/pull/2299.


mkroening:
We have resolved this by sleeping before exiting the application. That results in a clean shutdown and the port can be reused.

PR #1932, opened 2025-09-10
network: change implementation of sockopt

zyuiop:
The goal is to permit non boolean options in the future (I initially thought KEEPALIVE was a non boolean, which is why it's implemented now), and also to add the keepalive option which is required by some other software.

Also, we offload the processing of the arguments to the actual Object implementation


zyuiop:
superseded by #2455

PR #1940, opened 2025-09-17
Memory Allocator: Switch from talc to galloc

jounathaen:
I found major performance issues with the current talc allocator. This one replaces it with [galloc](https://github.com/MaderNoob/galloc).

Just like https://github.com/hermit-os/kernel/pull/1935, this PR is for now only intended for testing purposes. Finding the right allocator requires deeper investigations and benchmarking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -535,6 +535,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "either"
+version = "1.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


m-mueller678:
I am happy that we are getting a faster allocator, but I think galloc is still not particularly good. The benchmark from hermit-os/hermit-rs/issues/829 reports around 1s for galloc, so still a 20x slow-down compared to linux. I think using a single-threaded allocator and wrapping it in a mutex will never get good performance in a multithreaded setting. Unfortunately, there do not seem to be any allocators with a better synchronization scheme that are also easily portable to hermit.

PR #1943, opened 2025-09-19
Support RTL8139 on all architectures

Gelbpunkt:
RISC-V cannot compile because it lacks a udelay implementation. I could hack one up, but RISC-V has no PCI support at the moment, so I wouldn't be able to test it.


mkroening (approved):
Thanks! :)

PR #1949, opened 2025-09-26
fix(virtio-net): free up send capacity before checking it

cagatay-y:
Before checking the send capacity to issue a send token, we need to poll to get the capacity back from completed sends.


mkroening:
@zyuiop, please review too :)

mkroening:
> Any idea, why the CI failed?

Might be spurious. Retrying.

PR #1950, opened 2025-09-28
feat(nvme): add NVMe driver support

valopok:
I added NVMe driver support by using an [adapted version](https://github.com/valopok/vroom) of the [vroom driver](https://github.com/bootreer/vroom).
Alongside this pull request, I will open another pull request in [hermit-rs](https://github.com/hermit-os/hermit-rs/pulls) to add an example testing the driver.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -791,6 +791,12 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
+version = "0.15.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


jounathaen:
Hi @valopok and thank you for the contribution.
This is an interesting idea. Yet, I'm a little unsure, if this fit's the goals of Hermit, as I personally also see a value in keeping the codebase as small as possible. So, could you maybe give us a little motivation for the addition of this driver?

I'd also be curious, if it could alternatively be implemented outside the kernel and which interfaces to the kernel are missing if it isn't possible at the moment.

PR #1951, opened 2025-09-28
fix(pci): ignore incompatible capabilities instead of crashing

cagatay-y:
It is possible for PCI devices to offer multiple capabilities of the same type that are offered through different kinds of BARs (e.g. IO or memory), so it is possible to encounter an incompatible capability but still function with an alternative.


mkroening (approved):
Thanks! :)

Issue #1969, opened 2025-10-10
Call Uhyve function initialization function only in Uhyve (style-related, suppress `info!` message)


n0toose:
When running Hermit under QEMU, both fuse (provided that the feature is enabled) and the uhyve host filesystem layer are enabled. In line with the minimality principle, we should probably only automatically enable the latter when necessary.

PR #1970, opened 2025-10-10
refactor(fs): move uhyve check outside of init

n0toose:
Fixes https://github.com/hermit-os/kernel/issues/1969


mkroening:
Since this does not change the behavior and only moves a condition around, I think this would be a `refactor`. The commit and PR title should be updated to reflect that no behavior is changed. Not logging something anymore is not that important compared to moving the condition, I think.

mkroening (approved):
Thanks! :)

Issue #1352, opened 2024-08-07
Skip APIC detection on Uhyve


jounathaen:
I'm not 100% sure about this one, but AFAIK, Uhyve does not provide an APIC or IO-APIC. But if the `acpi` feature is active, the kernel unsuccessfully probes the memory to find one. I think, we can skip the search for an APIC if we are on Uhyve and save a few CPU cycles.

stlankes:
I would assume that we need at least a local APIC to program the timer interrupt.

jounathaen:
Yes. Something like this is happening. But the probing can probably be skipped.

PR #1971, opened 2025-10-10
perf(x86_64): skip APIC probe on uhyve

n0toose:
- Skip memory probes, use default address directly on Uhyve.
- Checking whether the processor supports x2apic in init_x2apic goes against the component's own conventions.
- We need a local APIC, but IO-APIC is not supported.
- Show more relevant info/warning messages.
- Fix minor formatting issue that shows "0x0x" prefix.

Fixes https://github.com/hermit-os/kernel/issues/1352


mkroening:
All right, I have further separated the refactoring from the first commit. I think you missed adding a condition in `application_processor_init`. I wonder why this caused no problems, though. It should have, as far as I understand.

Are you okay with my changes? :) [`c2159c8..9d15052`](https://github.com/hermit-os/kernel/compare/c2159c8b89c05393b0542a45e013222d340839ad..9d15052680fb6adc5bbe7505632bca8b99837b82)

mkroening (approved):
Thanks! :)

PR #1976, opened 2025-10-13
Uhyve mounts in arbitrary paths

### LINKED ISSUE #1572: Rework current file-system logic
Currently, Hermit creates a RAMfs on `/`, and then (optionally) mounts virtio and uhyve file systems into subdirectories of `/` (e.g., `/root` for uhyve).
This approach has the following problems: 
- Multiple mounts would be necessary to mount e.g.,  `/etc` and `/var/log`
- it is not possible to mount a single host file in the root directory (e.g., `/foo.txt`)

Additionally, Uhyve now already provides a file sandbox (thanks to @n0toose) that filters desired files and keeps the host-fs clean.

### Solution draft:

- We should default to uhve if available for `/`, else we use the ramfs.
- We should mount a special filesystem for `/proc`, `/run`, `/dev` and `/sys`
- We should mount the ramfs for `/tmp` (and maybe `/usr/tmp`)
- We could provide a mechanism for specifying other ramfs mounts (overshadowing uhvye)

jounathaen:
This allows Uhyve to announce multiple "mounts" via FDT.

Uhyve places the folders to be mounted in the FDT as follows
```
    uhyve,mounts {
        mounts ="/dir1\0/dir2\0/dir3\0"
    };
```

This does not yet allow single file mounts, but should already help in a lot of scenarios.

Fixes #1572

### DIFF
--- a/src/fs/mod.rs
+++ b/src/fs/mod.rs
@@ -404,6 +404,19 @@ pub fn create_dir(path: &str, mode: AccessPermission) -> io::Result<()> {
 	})
 }
 
+/// Creates a directory and creates all missing parent directories as well.
+fn create_dir_recursive(path: &str, mode: AccessPermission) -> io::Result<()> {
+	trace!("create_dir_recursive: {path}");
+	create_dir(path, mode).or_else(|errno| {
+		if errno != Errno::Badf {
+			return Err(errno);
+		}
+		let (parent_path, _file_name) = path.rsplit_once('/').unwrap();
+		create_dir_recursive(parent_path, mode)?;
+		create_dir(path, mode)
+	})
+}
+
 /// Returns an vector with all the entries within a directory.
 pub fn readdir(name: &str) -> io::Result<Vec<DirectoryEntry>> {
 	debug!("Read directory {name}");
--- a/src/fs/uhyve.rs
+++ b/src/fs/uhyve.rs
@@ -1,3 +1,4 @@
+use alloc::borrow::ToOwned;
 use alloc::boxed::Box;
 use alloc::ffi::CString;
 use alloc::string::{String, ToString};
@@ -14,9 +15,11 @@ use uhyve_interface::parameters::{
 use uhyve_interface::{GuestPhysAddr, GuestVirtAddr, Hypercall};
 
 use crate::arch::mm::paging;
+use crate::env::fdt;
 use crate::errno::Errno;
 use crate::fs::{
 	self, AccessPermission, FileAttr, NodeKind, ObjectInterface, OpenOption, SeekWhence, VfsNode,
+	create_dir_recursive,
 };
 use crate::io;
 use crate::syscalls::interfaces::uhyve::uhyve_hypercall;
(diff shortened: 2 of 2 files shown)


src/fs/mod.rs
```diff
@@ -402,6 +402,19 @@ pub fn create_dir(path: &str, mode: AccessPermission) -> io::Result<()> {
...
 }
 
+/// Creates a directory and creates all missing parent directories as well.
+fn create_dir_recursive(path: &str, mode: AccessPermission) -> io::Result<()> {
+	trace!("create_dir_recursive: {path}");
+	create_dir(path, mode).or_else(|errno| {
+		if errno != Errno::Badf {
+			return Err(errno);
+		}
+		let parent_path = &path[0..path.rfind('/').unwrap()];
```
mkroening:
```suggestion
		let (parent, _file_name) = path.rsplit_once('/').unwrap();
```

src/fs/uhyve.rs
```diff
@@ -231,15 +233,49 @@ impl VfsNode for UhyveDirectory {
...
-			.mount(
-				&mount_point,
-				Box::new(UhyveDirectory::new(Some(mount_point.clone()))),
-			)
-			.expect("Mount failed. Duplicate mount_point?");
+		let mount_str = env::fdt().and_then(|fdt| {
+			fdt.find_node("/uhyve,mounts").and_then(|node| {
+				node.property("mounts")
+					.and_then(|property| str::from_utf8(property.value).ok())
+			})
```
mkroening:
```suggestion
			fdt.find_node("/uhyve,mounts")
				.and_then(|node| node.property("mounts"))
				.and_then(|property| property.as_str())
```

src/fs/uhyve.rs
```diff
@@ -231,15 +233,49 @@ impl VfsNode for UhyveDirectory {
...
+		});
+		if let Some(mount_str) = mount_str {
+			assert_ne!(mount_str.len(), 0, "Invalid /uhyve,mounts node in FDT");
+			for mount_point in mount_str.trim_end_matches('\0').split('\0') {
```
mkroening: This is already done in `as_str` in my previous suggestion.

```suggestion
			for mount_point in mount_str.split('\0') {
```

src/fs/uhyve.rs
```diff
@@ -231,15 +233,49 @@ impl VfsNode for UhyveDirectory {
...
+				info!("Mounting uhyve filesystem at {mount_point}");
+
+				if let Err(errno) = fs::FILESYSTEM.get().unwrap().mount(
+					mount_point,
+					Box::new(UhyveDirectory::new(Some(mount_point.to_owned()))),
+				) {
+					debug!(
+						"Mounting of {mount_point} failed with {errno:?}. Creating missing parent folders"
+					);
+					let parent_path = &mount_point[0..mount_point.rfind('/').unwrap()];
```
mkroening:
```suggestion
					let (parent, _file_name) = path.rsplit_once('/').unwrap();
```

src/fs/uhyve.rs
```diff
@@ -231,15 +233,49 @@ impl VfsNode for UhyveDirectory {
...
+			for mount_point in mount_str.trim_end_matches('\0').split('\0') {
+				info!("Mounting uhyve filesystem at {mount_point}");
+
+				if let Err(errno) = fs::FILESYSTEM.get().unwrap().mount(
+					mount_point,
+					Box::new(UhyveDirectory::new(Some(mount_point.to_owned()))),
+				) {
+					debug!(
+						"Mounting of {mount_point} failed with {errno:?}. Creating missing parent folders"
+					);
```
mkroening: Should we check the error and only attempt recovery on `EBADF` similar to `create_dir_recursive`?

src/fs/mod.rs
jounathaen: hmm, yes, I see. But is this really an improvement?

src/fs/uhyve.rs
jounathaen: This doesn't work, as the last element in `mount_str` also ends on `\0`, thus creating an empy string for `mount_point`.

src/fs/mod.rs
mkroening: I'd say so, yes. No nested expressions and sun slices anymore hiding the split. :D

src/fs/uhyve.rs
mkroening: That's why I mentioned the previous suggestion. `as_str` removes the trailing null.

src/fs/mod.rs
jounathaen: Well, doesn't hurt at least

src/fs/uhyve.rs
jounathaen: I wouldn't know how we should do this. This is run in the initialization phase, so to trigger the error, the respective path must have already been mounted differently. As Uhyve is responsible for the creation of the VM, and thus of all other ways of providing a file system, I'd say a conflict here is the fault of Uhyve and exiting with an assert is the proper way.

mkroening: I would have leaned towards not panicking, but if you prefer panicking, that's fine with me. 👍

jounathaen: Yeah, if something is wrong here, I'd say the remainder of the application run is flawed as well.

PR #1977, opened 2025-10-13
Feature: Delegate `/tmp` to uhyve

jounathaen:
This feature delegates all `/tmp` file accesses to uhyve, which can handle them since https://github.com/hermit-os/uhyve/pull/783
This might be desirable, if the tempfiles are larger than the assigned memory, and it is not desirable when high-performance is required.

This depends on https://github.com/hermit-os/kernel/pull/1976

CC @n0toose


n0toose (approved):
lgtm

mkroening (approved):
Looks good to me! :)

mkroening (changes requested):
Let's merge this once we have a use case that benefits.

The code itself is good.

mkroening:
Let me know if you want to merge this. Until then, I'll mark this as draft. :)

jounathaen:
I think we can close this until we have a real use-case.

PR #1985, opened 2025-10-16
fix(pci): handle error values for IRQ number

cagatay-y:


mkroening:
Yeah, for the first one, warn makes sense. We might even want to downgrade that once we can handle non-legacy interrupts. The other ones are clear errors, though, right? 🤔

PR #1986, opened 2025-10-16
fix(virtio): don't include optional fields in minimum size

cagatay-y:


src/drivers/virtio/transport/pci.rs
```diff
@@ -124,7 +124,8 @@ impl PciCap {
 		}
 
 		// Drivers MAY do this check. See Virtio specification v1.1. - 4.1.4.1
-		if self.len() < u64::try_from(mem::size_of::<CommonCfg>()).unwrap() {
+		// The CommonCfg declaration includes 2 16-bit fields that are optional, totaling 4 bytes.
+		if self.len() < u64::try_from(mem::size_of::<CommonCfg>() - 4).unwrap() {
```
mkroening: Maybe this would be more expressive?

```suggestion
		// `CommonCfg::queue_notify_data` and `CommonCfg::queue_reset` are optional.
		let min_size = mem::size_of::<CommonCfg>() - mem::size_of::<[le16; 2]>();
		if self.len() < u64::try_from(min_size).unwrap() {
```

cagatay-y: Makes sense. I think we can also make it a `const`.

mkroening: `const` does not matter much to me, since this will be constant folded anyway. Do whatever you prefer code-wise.

Issue #1987, opened 2025-10-16
Memory region clamping seems wrong


m-mueller678:
I found this code in `src/mm/physicalmem.rs`, `detect_from_fdt`:

https://github.com/hermit-os/kernel/blob/9a6f950540934fb65246b8eba55654eef72d4b4a/src/mm/physicalmem.rs#L102-L106

Should this not be comparing against `super::kernel_end_address().as_u64()` instead? This would only make a difference if the memory region starts somewhere in the kernel address range, which probably never happens with a sane loader? But it seems much more logical.
Also, I think it could then be simplified to:
```rust
VirtAddr::new(start_address).max(super::kernel_end_address())
```

On another note, any regions before the start of the kernel are ignored. I do not understand why. I think they should either be used or an explaining comment be added.

I am happy to prepare a PR if desired.

Also, as far as I can tell the memory reservation block from the FDT is not respected, but that seems like a more complex thing to fix.

mkroening:
Thanks for opening this issue! :)

This code has not been changed since its introduction in https://github.com/hermit-os/kernel/commit/a34f3771a54839ca6d22e8a7df8f709bbbbf7b48.

The bug you found will likely not be triggered in practice, since it should be impossible for the kernel to start outside a physical memory region. Still, a PR making this clearer would be appreciated. 👍 

> On another note, any regions before the start of the kernel are ignored. I do not understand why. I think they should either be used or an explaining comment be added.

That is a historical artifact and should be fixed eventually. Previously, Hermit had assumptions about memory before the kernel and would show issues when changing this. If you are interested in testing whether any real issues show up when enabling using memory in front of the kernel image, feel free! :D

> Also, as far as I can tell the memory reservation block from the FDT is not respected, but that seems like a more complex thing to fix.

What do you mean with it is not respected? Any memory region in the FDT that comes partly after the kernel should be handled correctly, right? Handling the memory in front of the kernel image would be good to fix, though, see above.

m-mueller678:
Also, for uefi, only the largest region is claimed. Is there a reason for this?

PR #1991, opened 2025-10-17
Fix physical memory initialization

m-mueller678:
This addresses #1987. Memory regions reserved by the FDT are no longer claimed as available physical memory. The same strategy is applied for both uefi and bios systems. I do not know why uefi was previously treated differently, as mentioned in the issue. There remain four problems:

- the region occupied by the FDT itself is not reserved, but should be. The fdt crate does not expose this range.
- the region below the kernel is not claimed. Doing so leads to a reboot loop. The FDT seems to end up here, so that may be why.
- From my understanding, adding more than 16 regions to the physical free list will cause it to attempt to allocate. This will fail because the allocator is not yet initialized. Cutting up available memory into the ranges between reserved regions may bring us closer to this limit.
- in `init_frame_range`, the region is rounded outwards, this seems suspicious.


mkroening:
Thanks for the PR! :)

On x86-64, we create the FDT ourselves in the loader. For ARM and RISC-V, we pass it through verbatim, as far as I know. Should the memory of the FDT be marked as reserved in the FDT itself? Is that the case for the provided FDTs on ARM and RISC-V? If yes, then we could also add that in the loader for x86, instead of having to special-case this in the kernel and having to maintain an FDT crate fork.

src/mm/physicalmem.rs
```diff
@@ -30,6 +33,8 @@ pub unsafe fn init_frame_range(frame_range: PageRange) {
 		}
 	}
 
+	// FIXME: rounding outwards adds physical memory that is not actually available.
+	// This seems wrong.
```
mkroening, line 37: This should be fine, since this is only for mapping the pages, not using the memory. It should be better visible in the new PR.

mkroening (commented):
Thank you so much for the PR and the initiative! I have had a look at the PR locally and refactored it. Please review https://github.com/hermit-os/kernel/pull/1996. :)

PR #1996, opened 2025-10-20
feat(physicalmem): respect memory reservations

mkroening:
This is https://github.com/hermit-os/kernel/pull/1991, but refactored.

Closes https://github.com/hermit-os/kernel/pull/1991.

### DIFF
--- a/src/mm/physicalmem.rs
+++ b/src/mm/physicalmem.rs
@@ -1,7 +1,7 @@
 use core::sync::atomic::{AtomicUsize, Ordering};
 
 use align_address::Align;
-use free_list::{FreeList, PageRange};
+use free_list::{FreeList, PageRange, PageRangeError};
 use hermit_sync::InterruptTicketMutex;
 use memory_addresses::{PhysAddr, VirtAddr};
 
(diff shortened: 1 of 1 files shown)


mkroening:
@m-mueller678, are you okay with these changes to your PR? I can't request a review from you via GitHub, but please do so anyway. :)

src/mm/physicalmem.rs
```diff
@@ -61,58 +57,102 @@ pub unsafe fn init_frame_range(frame_range: PageRange) {
...
 		}
+		TOTAL_MEMORY.fetch_add(range.len().get(), Ordering::Relaxed);
+		debug!("Claimed physical memory: {range:#x?}");
+	}
+
+	let reserve = |reservation: PageRange| {
+		debug!("Memory reservation: {reservation:#x?}");
+		if let Ok(reserved) = PHYSICAL_FREE_LIST
+			.lock()
+			.allocate_with(|range| reservation.and(range))
```
m-mueller678, line 106: I think this does not work if reservations overlap or lie outside memory. I do not know if that can happen, but I could not find anything in the device tree docs that prohibits it.

src/arch/x86_64/mm/paging.rs
```diff
@@ -367,12 +367,12 @@ pub unsafe fn log_page_tables() {
 
 	use self::mapped_page_range_display::OffsetPageTableExt;
 
-	if !log_enabled!(Level::Debug) {
+	if !log_enabled!(Level::Trace) {
```
m-mueller678, line 370: This is a great change, I agree this was too noisy, but was not sure if I should change it.

src/mm/physicalmem.rs
mkroening: Why would overlapping reservations be an issue here? We iterate over all reservations and remove each from the list of available memory independently.

Why would reservations outside the memory be an issue? That kind of memory would not be usable anyway.

I just noticed that there might be an issue when we have one large reservation overlap with multiple memory ranges, though, since we only remove the first match. Replacing this with a while loop should help, though ([`d6786a2..50453a6`](https://github.com/hermit-os/kernel/compare/d6786a266755ce71725b77ef85b4187b49c09d43..50453a6839e48924ff65251cdb171fa98e5f048e)).

m-mueller678: I think with the while loop it should work. Honestly, I find the `allocate_with` interface quite confusing, even more so when called in a loop. I think a comment would make sense. Something along the lines of:
> While there are free intervals overlapping the reserved region, we allocate the intersection

mkroening: Done 👍

PR #2002, opened 2025-10-21
feat(mm): migrate to PageAlloc, FrameAlloc

mkroening:
Similar to https://github.com/hermit-os/kernel/pull/1994, this makes the internals of the physical and virtual memory allocation private. Note that this will slow down parts of paging right now, since it is not possible to hold a global `FrameAlloc` lock anymore. This means the lock has to be reacquired again and again until we improve the paging code or the `FrameAlloc` code.

Depends on https://github.com/hermit-os/kernel/pull/2008.
Closes https://github.com/hermit-os/kernel/pull/1994.

### DIFF
--- a/src/mm/physicalmem.rs
+++ b/src/mm/physicalmem.rs
@@ -22,7 +22,9 @@ pub struct FrameAlloc;
 
 impl PageRangeAllocator for FrameAlloc {
 	unsafe fn init() {
-		init();
+		unsafe {
+			init();
+		}
 	}
 
 	fn allocate(layout: PageLayout) -> Result<PageRange, AllocError> {
@@ -99,7 +101,7 @@ pub unsafe fn map_frame_range(frame_range: PageRange) {
 	}
 }
 
-fn detect_from_fdt() -> Result<(), ()> {
+unsafe fn detect_from_fdt() -> Result<(), ()> {
 	let fdt = env::fdt().ok_or(())?;
 
 	let all_regions = fdt
@@ -198,7 +200,7 @@ impl PageRangeExt for PageRange {
 }
 
 #[cfg(any(target_arch = "aarch64", target_arch = "riscv64"))]
-fn detect_from_limits() -> Result<(), ()> {
+unsafe fn detect_from_limits() -> Result<(), ()> {
 	let limit = crate::arch::kernel::get_limit();
 	if limit == 0 {
 		return Err(());
@@ -220,7 +222,7 @@ fn detect_from_limits() -> Result<(), ()> {
 	Ok(())
 }
 
-fn init() {
+unsafe fn init() {
 	if env::is_uefi() && DeviceAlloc.phys_offset() != VirtAddr::zero() {
 		let start = DeviceAlloc.phys_offset();
 		let count = DeviceAlloc.phys_offset().as_u64() / HugePageSize::SIZE;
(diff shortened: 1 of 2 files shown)


src/arch/riscv64/mm/mod.rs
```diff
@@ -1,8 +1,13 @@
 pub mod paging;
 
 pub use self::paging::init_page_tables;
+use crate::mm::{FrameAlloc, PageAlloc, PageRangeAllocator};
 
 pub fn init() {
-	crate::mm::physicalmem::init();
-	crate::mm::virtualmem::init();
+	unsafe {
```
m-mueller678, line 6: could be merged into a single unsafe block

src/arch/aarch64/mm/mod.rs
```diff
@@ -1,11 +1,17 @@
 pub mod paging;
 
+use crate::mm::{FrameAlloc, PageAlloc, PageRangeAllocator};
+
 pub fn init() {
 	unsafe {
```
m-mueller678, line 6: could be merged into a single unsafe block

src/arch/x86_64/mm/mod.rs
```diff
@@ -52,22 +52,21 @@ pub fn create_new_root_page_table() -> usize {
...
-		PageRange::from_start_len(virtaddr.as_usize(), 2 * BasePageSize::SIZE as usize).unwrap();
-	unsafe {
-		KERNEL_FREE_LIST.lock().deallocate(range).unwrap();
-	}
 
 	physaddr.as_usize()
 }
 
 pub fn init() {
 	paging::init();
```
m-mueller678, line 59: Again, unsafe could be merged

src/mm/page_range_alloc.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub struct PageRangeBox<A: PageRangeAllocator>(PageRange, PhantomData<A>);
+
+impl<A: PageRangeAllocator> PageRangeBox<A> {
+	pub fn new(layout: PageLayout) -> Result<Self, AllocError> {
+		let range = A::allocate(layout)?;
+		Ok(Self(range, PhantomData))
+	}
+
+	pub unsafe fn from_raw(range: PageRange) -> Self {
```
m-mueller678, line 34: Feels like there should be an `into_raw` too for symmetry.

src/mm/page_range_alloc.rs
```diff
@@ -0,0 +1,52 @@
...
+
+	/// Deallocates the pages described by `range`.
+	///
+	/// # Safety
+	///
+	/// - `range` must described a range of pages _currently allocated_ via this allocator.
+	unsafe fn deallocate(range: PageRange);
+}
+
+pub struct PageRangeBox<A: PageRangeAllocator>(PageRange, PhantomData<A>);
```
m-mueller678, line 26: I very much like the idea of `PageRangeBox` :+1:

src/mm/physicalmem.rs
```diff
@@ -1,20 +1,60 @@
...
 use crate::env;
+use crate::mm::{PageRangeAllocator, PageRangeBox};
 use crate::mm::device_alloc::DeviceAlloc;
 
-pub static PHYSICAL_FREE_LIST: InterruptTicketMutex<FreeList<16>> =
+static PHYSICAL_FREE_LIST: InterruptTicketMutex<FreeList<16>> =
 	InterruptTicketMutex::new(FreeList::new());
 pub static TOTAL_MEMORY: AtomicUsize = AtomicUsize::new(0);
 
+pub struct FrameAlloc;
```
m-mueller678, line 21: I like the idea of having these as methods rather than free standing functions. Both to enable `PageRangeBox` and to have something natural to implement `x86_64`'s `FrameAllocator` on.

src/arch/riscv64/mm/mod.rs
mkroening: Having these operations in separate blocks was on purpose. See [`clippy::multiple_unsafe_ops_per_block`](https://rust-lang.github.io/rust-clippy/stable/index.html#multiple_unsafe_ops_per_block). Arguably, the situation is far from ideal in the kernel still, but nevertheless, I think this is a step in the right direction.

This function itself should be unsafe too, I just noticed. Let's change that.

src/mm/page_range_alloc.rs
mkroening: Makes sense. I added it. 👍

PR #2007, opened 2025-10-22
perf(tls): use a TLS-based TCB

mkroening:


mkroening:
I don't think there is a benefit over using compiler-generated TLS here.

PR #2008, opened 2025-10-22
fix(mman): don't deallocate PROT_NONE frames

mkroening:
Wasmtime allocates 4 GiB of `PROT_NONE` (Data cannot be accessed) memory.

Since we don't allocate any backing physical frames in `mmap` for `PROT_NONE`, we should neither try to deallocate any physical frames in `munmap` in those cases.

### DIFF
--- a/src/syscalls/mman.rs
+++ b/src/syscalls/mman.rs
@@ -1,7 +1,8 @@
 use core::ffi::{c_int, c_void};
 
 use align_address::Align;
-use free_list::{PageLayout, PageRange};
+use free_list::{FreeList, PageLayout, PageRange};
+use hermit_sync::SpinMutex;
 use memory_addresses::{PhysAddr, VirtAddr};
 
 use crate::arch;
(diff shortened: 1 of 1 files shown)


m-mueller678:
This looks like an ok fix for the issue at hand to me. However, I think `mmap` has a bunch of problems and this is more of a band-aid solution. For example, this implementaiton breaks if the affected area contains both pages mapped with `PROT_NONE` and pages with some access allowed. From my reading of the man pages, it is perfectly fine to use linux `mmap` like that.

For a proper solution, we should decide on which level `mmap` should operate on:

If it is a high level api, we should track the state of all pages mapped via `mmap`. `munmap` should refuse to unmap pages which have not been allocated via `mmap`. This could be done via a hashtable or using the extra bits available in the pagetable. As an example of why this is needed, you could use `munmap` to unmap the identity mappings with the current implementation.

If it is a low level api, it should interact directly with the free lists and the page table. In that case, `munmap` should just walk the page table and add anything it finds to the physical free list. We should specify incorrect use of `mmap` to be UB, as we can not really detect it reliably. This would make our mmap incompatible with the linux one.

I would prefer `mmap` to be an optional high level feature and to have an additional set of syscalls that allows direct access to the free lists.

PR #2011, opened 2025-10-23
perf(x86_64): don't remap ACPI tables

mkroening:


mkroening:
This was done in https://github.com/hermit-os/kernel/pull/2615.

PR #2013, opened 2025-10-25
feat(uhyve): use v2 interface, MmioWrite exits on x86_64

n0toose:
Companion PR to https://github.com/hermit-os/uhyve/pull/1145

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,7 +135,7 @@ thiserror = { version = "2", default-features = false }
 time = { version = "0.3", default-features = false }
 volatile = "0.6"
 zerocopy = { version = "0.8", default-features = false }
-uhyve-interface = "0.1.3"
+uhyve-interface = "0.2.0"
 
 [dependencies.smoltcp]
 version = "0.12"
@@ -204,6 +204,7 @@ exclude = [
 ]
 
 [patch.crates-io]
+uhyve-interface = { git = "https://github.com/n0toose/uhyve", branch = 'uhyve_if_v2-mmiowrite' }
 # FIXME: remove once merged: https://github.com/rcore-os/trapframe-rs/pull/16
 trapframe = { git = "https://github.com/hermit-os/trapframe-rs", branch = "global_asm" }
 safe-mmio = { git = "https://github.com/hermit-os/safe-mmio", branch = "be" }
--- a/src/syscalls/interfaces/uhyve.rs
+++ b/src/syscalls/interfaces/uhyve.rs
@@ -1,14 +1,14 @@
 use core::ptr;
 
 use memory_addresses::VirtAddr;
-use uhyve_interface::parameters::{ExitParams, SerialWriteBufferParams};
-use uhyve_interface::{Hypercall, HypercallAddress};
+use uhyve_interface::v2::parameters::SerialWriteBufferParams;
+use uhyve_interface::v2::{Hypercall, HypercallAddress};
 
 use crate::arch;
 use crate::arch::mm::paging::{self, virtual_to_physical};
 use crate::syscalls::interfaces::SyscallInterface;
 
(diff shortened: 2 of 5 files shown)


n0toose:
Small status update: This can't work as-is yet, because we use `GuestVirtAddr` for the `write` and `read` implementations inside `src/fd/stdio.rs` and `src/fs/uhyve.rs`, whereas uhyve-interface v2 should now use `GuestPhysAddr` instead.

Cargo.toml
```diff
@@ -215,6 +215,7 @@ exclude = [
 ]
 
 [patch.crates-io]
+uhyve-interface = { git = "https://github.com/n0toose/uhyve", branch = 'uhyve_if_v2-mmiowrite' }
```
fogti, line 207:
```suggestion
uhyve-interface = { git = "https://github.com/n0toose/uhyve", branch = "uhyve_if_v2-mmiowrite" }
```

n0toose: Will be removed, this was added to 'declare' what branch must be used and allow for easier compilation (in theory...)

src/syscalls/interfaces/uhyve.rs
```diff
@@ -31,9 +31,7 @@ fn data_addr<T>(data: &T) -> u64 {
 #[inline]
 fn hypercall_data(hypercall: &Hypercall<'_>) -> u64 {
 	match hypercall {
-		Hypercall::Cmdsize(data) => data_addr(*data),
-		Hypercall::Cmdval(data) => data_addr(*data),
-		Hypercall::Exit(data) => data_addr(*data),
+		Hypercall::Exit(exit_code) => u64::try_from(*exit_code).unwrap(),
```
fogti: I think `(*exit_code) as u64` would be a better idea (the current version mangles the sign bit / panics on wrong sign)

n0toose: Thanks for helping me out with this.

n0toose:
```
---- fd_open_remove_before_and_after_closing stdout ----
Building test kernel: fs_tests

thread '<unnamed>' (166136) panicked at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/memory_addresses-0.2.3/src/arch/x86_64.rs:317:23:
physical addresses must not have any bits in the range 52 to 64 set
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

woops, I did not consider that either :D

fogti: you can clear that part with a `& 0xffff_ffff_ffff`, given that the mapping, even with that, remains injective.

fogti: alternatively, just do `(*exit_code) as u32 as u64`, which should prevent the sign-extension.

n0toose: already done :D

n0toose:
The issue causing the CI to fail was technically fixed in https://github.com/hermit-os/kernel/pull/2066/files, but we don't expect a new release for v1 of uhyve-interface (in which https://github.com/hermit-os/kernel/pull/2066/files would've helped.)

We can work around this by using @fogti's work that effectively shifts this responsibility onto `hermit-entry`:
- https://github.com/hermit-os/hermit-entry/pull/64
- https://github.com/hermit-os/kernel/pull/2066

Alternatively, we can use `[patch.crates-io]` and force the kernel to use the latest commit for Uhyve.

CC: @jounathaen, @mkroening

fogti:
> but we don't expect a new release for v1 of `uhyve-interface`

Note that we don't need a new release of `uhyve-interface` because the error itself only manifests if we use the faulty macro from `uhyve-interface`.

PR #1995, opened 2025-10-20
Draft: Add virtio balloon

hcsch:
Description and more explanatory comments are to follow

PR #2017, opened 2025-10-27
feat: support regularly printing allocation stats

mkroening:
This was proposed by @hcsch in https://github.com/hermit-os/kernel/pull/1995 (https://github.com/hermit-os/kernel/pull/1995/commits/858b1e6550b502bea2b9f1148778e5b45c01f88c).

Is this useful to us without the context of that PR?


Gelbpunkt (approved):
This code doesn't hurt and we can always remove it in case we decide against talc in the future.

PR #2047, opened 2025-11-03
refactor(pvirtq): combine `u15` offset and `u1` wrap counter in more places

mkroening:
This PR makes use of the offset + wrap bitfield for saving one byte per index. I am not sure if there is any performance change, but this also makes the code more readable by combining these related fields.

Depends on https://github.com/hermit-os/kernel/pull/2046.


mkroening:
Makes sense. I renamed the commits to imply only a refactor. Merging this once CI passes.

PR #2049, opened 2025-11-04
perf(virtqueue): replace vec-based `MemPools` with bitmap-based `IndexAlloc`

mkroening:
This replaces the vec-based `MemPool` with a bitmap-based `IndexAlloc`. To track 256 indexes, we now need 32 bytes instead of 512 bytes.

These are measurements from an Apple M2 of creating the allocator, allocating all indices, and then deallocating them again:

| len  | size old | size new | time old  | time new  |
| ----:| --------:| --------:| ---------:| ---------:|
|  256 |      512 |       32 |  1.242 µs |  1.370 µs |
| 1024 |     2048 |       64 |. 5.505 µs |  7.293 µs | 
| 2048 |     4096 |      128 | 10.460 µs | 21.273 µs |

While this is a strict slowdown in this case, I think it is still worth it.

### DIFF
--- a/src/drivers/virtio/virtqueue/mod.rs
+++ b/src/drivers/virtio/virtqueue/mod.rs
@@ -512,21 +512,58 @@ pub enum BufferType {
 	Indirect,
 }
 
-/// MemPool allows to easily control, request and provide memory for Virtqueues.
-struct MemPool {
-	pool: Vec<u16>,
-}
+mod index_alloc {
+	use alloc::boxed::Box;
 
-impl MemPool {
-	/// Returns a given id to the id pool
-	fn ret_id(&mut self, id: u16) {
-		self.pool.push(id);
+	/// This type allows allocating indices.
+	///
+	/// The indices can be used as descriptor IDs.
+	pub struct IndexAlloc {
+		/// Zero bits are available.
+		bits: Box<[usize]>,
 	}
 
-	/// Returns a new instance, with a pool of the specified size.
-	fn new(size: u16) -> MemPool {
-		MemPool {
-			pool: (0..size).collect(),
+	const USIZE_BITS: usize = usize::BITS as usize;
+
+	impl IndexAlloc {
+		pub fn new(len: usize) -> Self {
+			let usizes = len.div_ceil(USIZE_BITS);
+			let extra_bits = len % USIZE_BITS;
+
+			let mut bits = vec![0; usizes].into_boxed_slice();
+
+			if extra_bits != 0 {
+				*bits.last_mut().unwrap() = usize::MAX >> extra_bits;
+			}
+
+			Self { bits }
+		}
+
+		#[inline]
+		pub fn allocate(&mut self) -> Option<usize> {
+			for (word_index, word) in self.bits.iter_mut().enumerate() {
+				let trailing_ones = word.trailing_ones();
+				if trailing_ones < usize::BITS {
+					let mask = 1 << trailing_ones;
+					*word |= mask;
+					let index = word_index * USIZE_BITS + usize::try_from(trailing_ones).unwrap();
+					return Some(index);
+				}
+			}
+
+			None
+		}
+
+		#[inline]
+		pub unsafe fn deallocate(&mut self, index: usize) {
+			let word_index = index / USIZE_BITS;
+			let bit = index % USIZE_BITS;
+			let mask = 1 << bit;
+
+			debug_assert!(self.bits[word_index] & mask == mask);
+			unsafe {
+				*self.bits.get_unchecked_mut(word_index) &= !mask;
+			}
 		}
 	}
 }
--- a/src/drivers/virtio/virtqueue/packed.rs
+++ b/src/drivers/virtio/virtqueue/packed.rs
@@ -22,9 +22,8 @@ use super::super::transport::mmio::{ComCfg, NotifCfg, NotifCtrl};
 #[cfg(feature = "pci")]
 use super::super::transport::pci::{ComCfg, NotifCfg, NotifCtrl};
 use super::error::VirtqError;
-use super::{
-	AvailBufferToken, BufferType, MemPool, TransferToken, UsedBufferToken, Virtq, VirtqPrivate,
-};
+use super::index_alloc::IndexAlloc;
+use super::{AvailBufferToken, BufferType, TransferToken, UsedBufferToken, Virtq, VirtqPrivate};
 use crate::arch::mm::paging::{BasePageSize, PageSize};
 use crate::mm::device_alloc::DeviceAlloc;
 
@@ -69,9 +68,8 @@ struct DescriptorRing {
 	/// See Virtio specification v1.1. - 2.7.1
 	drv_wc: bool,
 	dev_wc: bool,
-	/// Memory pool controls the amount of "free floating" descriptors
-	/// See [MemPool] docs for detail.
-	mem_pool: MemPool,
+	/// This allocates available descriptors.
+	indexes: IndexAlloc,
 }
 
 impl DescriptorRing {
@@ -92,7 +90,7 @@ impl DescriptorRing {
 			poll_index: 0,
 			drv_wc: true,
 			dev_wc: true,
-			mem_pool: MemPool::new(size),
+			indexes: IndexAlloc::new(size.into()),
 		}
 	}
 
@@ -186,13 +184,13 @@ impl DescriptorRing {
 	/// Returns an initialized write controller in order
 	/// to write the queue correctly.
 	fn get_write_ctrler(&mut self) -> Result<WriteCtrl<'_>, VirtqError> {
-		let desc_id = self.mem_pool.pool.pop().ok_or(VirtqError::NoDescrAvail)?;
+		let desc_id = self.indexes.allocate().ok_or(VirtqError::NoDescrAvail)?;
 		Ok(WriteCtrl {
 			start: self.write_index,
 			position: self.write_index,
 			modulo: u16::try_from(self.ring.len()).unwrap(),
 			first_flags: DescF::empty(),
-			buff_id: desc_id,
+			buff_id: u16::try_from(desc_id).unwrap(),
 
 			desc_ring: self,
 		})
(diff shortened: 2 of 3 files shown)


src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -512,21 +512,65 @@ pub enum BufferType {
...
+				}
+			}
+
+			None
+		}
+
+		#[inline]
+		pub unsafe fn deallocate(&mut self, index: usize) {
+			let word_index = index / 64;
+			let bit = index % 64;
```
cagatay-y: For consistency we can use `USIZE_BITS` here instead of hard coding 64.

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -512,27 +512,65 @@ pub enum BufferType {
...
+
+	impl IndexAlloc {
+		pub fn new(len: usize) -> Self {
+			let usizes = len.align_up(USIZE_BITS) / USIZE_BITS;
+			let extra_bits = len % USIZE_BITS;
+
+			let mut bits = vec![0; usizes].into_boxed_slice();
+
+			if extra_bits != 0 {
+				*bits.last_mut().unwrap() |= usize::MAX >> extra_bits;
```
cagatay-y:
```suggestion
				*bits.last_mut().unwrap() = usize::MAX >> extra_bits;
```
Should not change the result but I think it makes the intention clearer.

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -512,27 +512,65 @@ pub enum BufferType {
...
-	/// Returns a new instance, with a pool of the specified size.
-	fn new(size: u16) -> MemPool {
-		MemPool {
-			pool: (0..size).collect(),
-			limit: size,
+	const USIZE_BITS: usize = usize::BITS as usize;
+
+	impl IndexAlloc {
+		pub fn new(len: usize) -> Self {
+			let usizes = len.align_up(USIZE_BITS) / USIZE_BITS;
```
cagatay-y:
```suggestion
			let usizes = len.div_ceil(USIZE_BITS);
```

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -512,27 +512,65 @@ pub enum BufferType {
...
+				let trailing_ones = word.trailing_ones();
+				if trailing_ones < usize::BITS {
+					let mask = 1 << trailing_ones;
+					*word |= mask;
+					let index = word_index * USIZE_BITS + usize::try_from(trailing_ones).unwrap();
+					return Some(index);
+				}
+			}
+
+			None
```
cagatay-y, line 554:
```suggestion
			let (word_index, trailing_ones) = self
				.bits
				.iter()
				.copied()
				.map(usize::trailing_ones)
				.enumerate()
				.find(|(_, trailing_ones)| *trailing_ones < usize::BITS)?;

			let mask = 1 << trailing_ones;
			self.bits[word_index] |= mask;
			let index = word_index * USIZE_BITS + usize::try_from(trailing_ones).unwrap();
			Some(index)
```
I am not sure if it would be an improvement but wanted to offer it as an option. It would save us from some nesting.

src/drivers/virtio/virtqueue/mod.rs
```diff
@@ -512,27 +512,65 @@ pub enum BufferType {
...
+
+			if cfg!(debug_assertions) {
+				let old = self.bits[word_index];
+				assert!(old & mask == mask);
+				self.bits[word_index] = old & !mask;
+			} else {
+				unsafe {
+					*self.bits.get_unchecked_mut(word_index) &= !mask;
+				}
+			}
```
cagatay-y: Would it make sense to share the masking operation in the debug and the release code paths and remove the conditional? Something like:

```rust
debug_assert!(self.bits[word_index] & mask == mask);
unsafe {
	*self.bits.get_unchecked_mut(word_index) &= !mask;
}
```

The existence of an item at the specified index will be checked by the read for the assertion anyway.

mkroening: Good idea! 👍

src/drivers/virtio/virtqueue/mod.rs
mkroening: Interesting! I have looked into this, and the compiler fails to optimize the bounds check when setting the bit. Also, maybe because the trailing ones calculation is too far away now, the compiler no longer optimizes the masking from `shl` and `or` to `bts`.

For details, see [Compiler Explorer](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywYSNpRwBk8DJgAcu4ARpjEIACsAOykAA6oCoR2DC5uHiAAzACcCUkpAv6BISzhkVHSVpg2qUIETMQE6e6eXN7VtQL1jQTFwWER0XHmvS2ZuZYNTf2l5dGSAJSWqCbEyOwcAPRbANQAKgCe8Zi7h6vEu2hYuwgRmKS7JLu0qEzou4a7mKqs8fQAOgApBoAIJmU7mdAgFACABuEQIMP2xEOADFiKgWECsgAhEGggk7Pb7BB4BS7AjHU5iV4AdwptNQolsRl2AXwGwUwLBxKJOwOd3ZDE5mApogYu3CuwhHyYFKwCmQxDw8SIlwAkgARbkE%2BImUK7cymGy7DUin6g2ivZC7IExfFg3bO3bE3YALQiqClhEZxBpcKYdCYoUBBJdPoIChAu1xqFUOKcQKiuLMeAAXphk1qcdhSAT7TmwQS0AxzLs5EINe7sAB9XEa/ZCGNpzN2rJamXJTMwhtNz4U1tZvEFsF4Fj/M0W1RWm12h3hl1AgBMWWTuICtACWaiRdBEf1hqoksCdIg9E8XYzmEWuwAtLndkJMLQqPPHfuIy76AQr5mKTinYXq6FZVjW9aNkIOIfl%2B36YL%2BPwEMQTC1qEvrtkBjB2suUSgdWdZ9lBI7Fk6sEvPBuwsCYv5oVGGG7AiyBgGA64aNBf5itmAIBEQqHxpg6C1goW4bBAizQaOn6wXgb6IchqHocxgG7Bo76LmRLoAFS0dyBjmLWVEEGJAImAwdLIfEYnzkmHYcTCACyoIABrttgj5yShOkSaRsGFpJGnPq%2B76RgBMR7hGfkkVJ2FrimpkKEwVCYBADCoAZhjAPQ4m7upuyHrsx6fNazJMAQyUrgAbIZRovlQt4Plk2C7AA8mqqSJkOj72jBZH8JcEB0iQgkcj8jyDcQ6C3gENWvgCOncWVxAGdRxmMO4ESlclt7dblGk/pSyF0AEwC1gIYr0eN6AAkhQZbkYp2BAoYnedFGkyQdt3HQ952JnZICEWpPkaV%2B%2B0sPKADW9FcO2NlOB9R33WdCgvcDZGaZd1nKWDCjgyjqMgxRI2qBdQ21kTuyaXh4EAyuuJ/Uhhy1lQmIsBAN0IydSOLCZZkWc9xGvfj/oEGskpCFiyVE%2BJAv4/O4VkZFhJAxGQRnblivq6u67xYlyWpelRhZdmuX5YVTIspgy4QJV1VWHV96Pq1tgCB13bDk1O3K3Bv4DaTROPOzd2c49222Xbu1kXNvoR7BC0RM9Xux2g8R4AJCeC7HYOWUOMKB19XMx1%2BAJrWwyFlenssAjQIrWzETgQLWAeHUH30KOJdcU3niOPTDf2EdLaIvYXoMQ1DMO/V3wecTLZE60lgMZ1%2Bml21HUYAsA8G1qZyB3Mg4MCctRmXWT07bXXWMQ3jvlhcPhPTiTE0n1gxOU5W%2BEQf2tP06iTMs2zzf5xDjzcyTBLLSx6rBcWbAIBS3VjfKKEUtYpk3NuY2QN8pz1OIVLA5tNrW2XFVaiM0qCPCJi2N2Z8IEE1/MfcmylyZ7AqpIK%2B1DIz0ToThXYTCWERhHjjMeiZfq0SHondkb5kBUGAMxCAWBQgmBOvKJQTRUhtwXrLfargPjKRXl5FMtDpzGxnqjRRiJpGaOwhVSio9AIXxxuAwuEYdHRz0X7Axu56LmMqrsZi2NcZGK/IWb4tAlBqPxpg0JstnTL1qqvbkG8CBbwYDvGo%2B9BKGV9o/KWFjlI%2BMvv4jSisCnwOihrYpHBli0E4FEXgngOBaFIKgTgAAlMwv4FAXA2DFHgpAkR1PKcscG0RvCVI4JIGpmheCNI4LwaM3helaGWHAWASA0ATjoBEcglBVkp3oJEYgXAcjtD4HQRa0YIChAmaQNCzBUScG6dcxohxmqhG0DUXp3TVlsEEM1BgtBDiXNkfIpwtJozcF4FgMGhtNj1PwP6WoCJQX1J%2BDUaimxuk8RfJcrcoRkKohcFgS5SFxx3N4AiYgoQkiYC1JgSFmVjoTOWFQAwwAFAADVU50laowElMhBAiDEOwKQvL5BKDUJc3QXB9AZRAKYcw%2Bg8ChGjJAZYqA2oCFBXeFwao7z0ARLQQCWRJlkpVIqeAyxOjO08BARw4xPDLmkH4bccwhjLm8IkZIlrbUgHtfkD1qRZiDEiK6ywL43l1GmM0VwrRvWSoteG3oAayguo6BGr1y5Y0RsTfMV15qOmCoqVU8ZfSGmcF2LKggtouAAkOQCVSEBcCEGeCuLIXBFi8Hmf00ggyojDM4GM0gtT6lTJmSAOZDKC0cGXEWodnB23jtIGS5I9hJBAA%3D).

So I'd keep it as is, even though the performance difference is small, of course (about 5%). :D

PR #2057, opened 2025-11-04
using stp/ldp to save/restore the FPU state

stlankes:
In addition, all FPU registers will be saved/restored. Currently, a premptive scheduling is supported and a context switch should not depend on the AARCH64 ABI.

### DIFF
--- a/src/arch/aarch64/kernel/processor.rs
+++ b/src/arch/aarch64/kernel/processor.rs
@@ -10,32 +10,17 @@ use crate::env;
 /// Current FPU state. Saved at context switch when changed.
 ///
 /// AArch64 mandates 32 NEON SIMD registers, which are named v0-v32.
-/// Only the lower 64 bits of v8-v15 must be saved (the d parts).
 ///
-/// See the ABI documentation for aarch64 on this topic:
-/// <https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#612simd-and-floating-point-registers>
+/// See the Arm documentation for more information:
+/// <https://developer.arm.com/documentation/102374/0103/Registers-in-AArch64---general-purpose-registers>
 ///
 /// FPCR is the floating point control register and controls things like NaN
 /// propagation, FPSR contains info like carry condition and over overflow
 /// condition. These are callee-saved bits.
 #[derive(Clone, Copy, Debug)]
 pub struct FPUState {
-	/// d8 register
-	d8: u64,
-	/// d9 register
-	d9: u64,
-	/// d10 register
-	d10: u64,
-	/// d11 register
-	d11: u64,
-	/// d12 register
-	d12: u64,
-	/// d13 register
-	d13: u64,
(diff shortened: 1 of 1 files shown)


src/arch/aarch64/kernel/processor.rs
```diff
@@ -9,33 +9,78 @@ use crate::env;
 
 /// Current FPU state. Saved at context switch when changed.
 ///
-/// AArch64 mandates 32 NEON SIMD registers, which are named v0-v32.
-/// Only the lower 64 bits of v8-v15 must be saved (the d parts).
-///
-/// See the ABI documentation for aarch64 on this topic:
-/// <https://github.com/ARM-software/abi-aa/blob/main/aapcs64/aapcs64.rst#612simd-and-floating-point-registers>
+/// AArch64 mandates 16 NEON 128 bit SIMD registers, which are named q0-q15.
```
Gelbpunkt: Why was this changed from 32 to 16? It's clearly 32, see [the docs](https://developer.arm.com/documentation/102374/0103/Registers-in-AArch64---general-purpose-registers) or the [NEON quick reference](https://developer.arm.com/community/arm-community-blogs/b/operating-systems-blog/posts/arm-neon-programming-quick-reference). Only in 32-bit ARM the SIMD registers are 16x 128-bit or 32x 64-bit depending on how they're viewed, 64-bit is always 32x 128-bit.
`v` and `q` suffixes are synonymous and depend on whether they're treated as floating point (`q`) or vector (`v`) registers.

mkroening: Yes, it is 32, and the code reflects that. We just need to fix the doc comment.

You got v and q mixed up, I think: https://developer.arm.com/documentation/dui0801/l/Overview-of-AArch64-state/Predeclared-extension-register-names-in-AArch64-state

Gelbpunkt: Hmm maybe I phrased it a bit misleading: `v` is the name of the register, `q` is the form used to refer to 128-bit scalar contents inside. I find this graphic from the docs quite helpful:

image

Therefore I think we should refer to them with the `v` prefix.

mkroening: [STP](https://developer.arm.com/documentation/ddi0602/2025-09/SIMD-FP-Instructions/STP--SIMD-FP---Store-pair-of-SIMD-FP-registers-) does not accept v but q, though. The instruction requires being explicit about this.


```
error: invalid operand for instruction
  --> src/arch/aarch64/kernel/processor.rs:67:6
   |
67 |                 "stp     v0,  v1,  [sp, #16*0]",
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: instantiated into assembly here
  --> <inline asm>:2:9
   |
 2 | stp     v0,  v1,  [sp, #16*0]
   |         ^
```

Gelbpunkt (approved):

PR #2066, opened 2025-11-14
refactor: use uhyve version define macro from hermit_entry

fogti:
Depends on https://github.com/hermit-os/hermit-entry/pull/64, and subsequent new `hermit-entry` release.


mkroening (approved):
Thanks! :)

Can you bump this to version 0.10.7 (with the docs.rs fix)?

mkroening (approved):

PR #2075, opened 2025-11-20
fix(pcie): include the end bus number in scan

cagatay-y:
According to the sections 4.1.1 and 4.1.2 of the PCI firmware specification v3.0, PCI segments support up to 256 buses, which would mean that the value in the 8-bit end bus number field is included in the range of valid bus numbers. Previous behavious caused no busses to be scanned when only one was available.


mkroening (approved):
Thanks! :)

PR #2076, opened 2025-11-21
fix(syscall/socket): follow POSIX and Linux for errno

cagatay-y:
Return error codes that follow the POSIX specification for socket (for EAFNOSUPPORT, EPROTONOSUPPORT, EPROTOTYPE, ESOCKTNOSUPPORT) or the Linux IPv4 implementation as linked by the Linux socket manpage (for ENETDOWN).


mkroening (approved):
Thanks! :)

PR #2077, opened 2025-11-22
feat(fs/mem): tarfs implementation

### LINKED ISSUE #2009: Investigate `tarfs` support (as part of image support)
This came up during work on https://github.com/hermit-os/uhyve/issues/1015.

This would be a way to reduce possible future code duplication / duplicated work between `loader` and `uhyve`.
The main alternative to this is:
* `uhyve` implements "virtual file descriptors" to refer to read-only segments / blobs / slices of (decompressed tar) files.
* `loader` implements a way to parse tar files and pass the file tree via FDT (device tree) to the kernel, the kernel would then handle this as a read-only file system.

cc @n0toose

fogti:
Fixes #2009.

### DIFF
--- a/src/fs/mem.rs
+++ b/src/fs/mem.rs
@@ -12,6 +12,7 @@ use core::mem::{MaybeUninit, offset_of};
 use align_address::Align;
 use async_lock::{Mutex, RwLock};
 use async_trait::async_trait;
+use hermit_entry::ThinTree;
 
 use crate::errno::Errno;
 use crate::executor::block_on;
@@ -302,6 +303,10 @@ impl RomFile {
 	pub fn new(data: &'static [u8], mode: AccessPermission) -> Self {
 		let microseconds = arch::kernel::systemtime::now_micros();
 		let t = timespec::from_usec(microseconds as i64);
+		Self::new_with_timestamp(data, mode, t)
+	}
+
+	pub fn new_with_timestamp(data: &'static [u8], mode: AccessPermission, t: timespec) -> Self {
 		let attr = FileAttr {
 			st_size: data.len().try_into().unwrap(),
 			st_mode: mode | AccessPermission::S_IFREG,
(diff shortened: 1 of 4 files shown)


src/fs/mem.rs
```diff
@@ -484,6 +489,32 @@ pub(crate) struct MemDirectory {
 	attr: FileAttr,
 }
 
+fn thin_tree_to_vfs_node(thin_tree: ThinTree<'static>, t: timespec) -> io::Result<Box<dyn VfsNode + Send + Sync>> {
+	Ok(match thin_tree {
+		ThinTree::File { content, metadata } => {
+			let mut st_mode = AccessPermission::S_IRUSR;
```
fogti, line 470: TODO: what file permissions should we actually set?

src/fs/mod.rs
```diff
@@ -340,18 +341,51 @@ pub(crate) fn init() {
...
+
+	// Handle optional Hermit Image specified in FDT.
+	let mut tar_image: Option<&'static [u8]> = None;
+	if let Some(fdt) = fdt() {
+		if let Some(chosen) = fdt.find_node("/chosen") {
+			if let Some(fdt::node::NodeProperty {
+				value: image_reg, ..
+			}) = chosen.property("image_reg")
+			{
+				let cell_sizes = fdt.cell_sizes();
```
fogti: TODO: This line is broken.

Use `fdt.root().cell_sizes()` instead.

src/fs/mem.rs
```diff
@@ -472,11 +474,40 @@ impl ObjectInterface for MemDirectoryInterface {
...
+			if metadata.is_exec {
+				st_mode |= AccessPermission::S_IXUSR;
+			}
+			Box::new(RomFile::new_with_timestamp(content, st_mode, t))
+				as Box<dyn VfsNode + Send + Sync>
+		}
+		ThinTree::Directory(d) => Box::new(MemDirectory {
+			inner: Arc::new(RwLock::new(
+				d.into_iter()
+					.map(|(key, value)| Ok((key.to_string(), thin_tree_to_vfs_node(value, t)?)))
```
fogti: This doesn't work because `ThinTree` filenames are `Vec<u8>`, not strings, which makes stuff a bit complicated.

PR #2085, opened 2025-11-26
feat: add `HERMIT_RUSTFLAGS` for allowing LTO kernels and more

mkroening:
See https://github.com/hermit-os/hermit-rs/pull/864 for Rust support.

### DIFF
--- a/xtask/src/build.rs
+++ b/xtask/src/build.rs
@@ -86,22 +86,25 @@ impl Build {
 	}
 
 	fn cargo_encoded_rustflags(&self) -> Result<String> {
-		assert_eq!(
-			env::var("CARGO_ENCODED_RUSTFLAGS"),
-			Err(VarError::NotPresent)
-		);
-		let mut rustflags = Vec::new();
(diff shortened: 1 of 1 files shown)


xtask/src/build.rs
```diff
@@ -122,3 +122,39 @@ impl Build {
...
+		Err(VarError::NotUnicode(s)) => {
+			panic!("HERMIT_ENCODED_RUSTFLAGS did not contain valid unicode data: {s:?}")
+		}
+	}
+
+	match env::var("HERMIT_RUSTFLAGS") {
+		Ok(s) => {
+			let hermit_rustflags = s
+				.split(' ')
+				.filter(|s| !s.is_empty())
```
fogti, line 148: Is there any justification to not have `.map(str::trim)` before this step?

fogti (approved):

xtask/src/build.rs
mkroening: No, thanks for pointing that out! :)

PR #2086, opened 2025-11-27
fix: Proper network wakers

Gelbpunkt:
This refactors our timer interrupt handling into a separate module that all timer interrupts are configured through and handled by. It allows us to track why a timer interrupt was fired.

We make use of this to now only conditionally wake the network task's waker when necessary. This is either because we sent some network packets, a network device driver received an interrupt, or because a timer interrupt that we configured so we can poll smoltcp in the future was handled.

Closes https://github.com/hermit-os/kernel/pull/1933.
Closes https://github.com/hermit-os/kernel/issues/2126.

### DIFF
--- a/src/arch/aarch64/kernel/interrupts.rs
+++ b/src/arch/aarch64/kernel/interrupts.rs
@@ -25,7 +25,7 @@ use crate::drivers::pci::get_interrupt_handlers;
 use crate::drivers::{InterruptHandlerQueue, InterruptLine};
 use crate::kernel::serial::handle_uart_interrupt;
 use crate::mm::{PageAlloc, PageRangeAllocator};
-use crate::scheduler::{self, CoreId};
+use crate::scheduler::{self, CoreId, timer_interrupts};
 use crate::{core_id, core_scheduler, env};
 
 /// The ID of the first Private Peripheral Interrupt.
@@ -94,10 +94,7 @@ pub(crate) fn install_handlers() {
 
 	fn timer_handler() {
 		debug!("Handle timer interrupt");
-
-		// disable timer
-		CNTP_CVAL_EL0.set(0);
-		CNTP_CTL_EL0.write(CNTP_CTL_EL0::ENABLE::CLEAR);
+		timer_interrupts::clear_active_and_set_next();
 	}
 
 	for (key, value) in get_interrupt_handlers().into_iter() {
--- a/src/drivers/net/virtio/mod.rs
+++ b/src/drivers/net/virtio/mod.rs
@@ -43,6 +43,7 @@ use crate::drivers::virtio::virtqueue::{
 	AvailBufferToken, BufferElem, BufferType, UsedBufferToken, VirtQueue, Virtq,
 };
 use crate::drivers::{Driver, InterruptLine};
+use crate::executor::network::wake_network_waker;
 use crate::mm::device_alloc::DeviceAlloc;
 
 /// A wrapper struct for the raw configuration structure.
@@ -418,6 +419,8 @@ impl NetworkDriver for VirtioNetDriver<Init> {
 			info!("Configuration changes are not possible! Aborting");
 			todo!("Implement possibility to change config on the fly...")
 		}
+
+		wake_network_waker();
 	}
 }
 
--- a/src/executor/mod.rs
+++ b/src/executor/mod.rs
@@ -18,15 +18,11 @@ use core::time::Duration;
 
 use crossbeam_utils::Backoff;
 use hermit_sync::without_interrupts;
-#[cfg(feature = "net")]
-use smoltcp::time::Instant;
 
 use crate::arch::core_local;
 use crate::errno::Errno;
 use crate::executor::task::AsyncTask;
 use crate::io;
-#[cfg(feature = "net")]
-use crate::scheduler::PerCoreSchedulerExt;
 use crate::synch::futex::*;
 
 /// WakerRegistration is derived from smoltcp's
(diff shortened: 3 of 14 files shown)


src/scheduler/timer_interrupts.rs
```diff
@@ -0,0 +1,137 @@
...
+
+/// A slot in the timer list. Each source is represented once. This is so that
+/// we can have multiple timers at the same time with only one hardware timer.
+#[derive(Debug)]
+pub struct Slot {
+	/// Timer source.
+	source: Source,
+	/// Point in time at which to wake up (in microsecond precision).
+	/// A value of [`u64::MAX`] means the timer is not set.
+	wakeup_time: u64,
```
mkroening, line 25: Would an `Option<NonZero<u64>>` also work? Wanting to fire at absolute time `0` would be a bug, no?

src/arch/timer_interrupts.rs
```diff
@@ -0,0 +1,137 @@
...
+	fn deref(&self) -> &Self::Target {
+		&self.0
+	}
+}
+
+impl DerefMut for TimerList {
+	fn deref_mut(&mut self) -> &mut Self::Target {
+		&mut self.0
+	}
+}
```
mkroening: I don't think we should implement `Deref` and `DerefMut` here in case we want to add methods to `TimerList` later on. That would also allow us to keep the inner list private.

src/scheduler/task/mod.rs
```diff
@@ -594,35 +577,20 @@ impl BlockedTaskQueue {
...
-					set_oneshot_timer();
+					create_timer_abs(Source::Scheduler, wt);
 					return;
 				}
 
 				cursor.move_next();
 			}
 
-			set_oneshot_timer();
+			create_timer_abs(Source::Scheduler, wt);
```
mkroening, line 581: The docs say that `wakeup_time` is relative. This was also done wrongly before, but maybe we should follow up to fix this.

src/arch/timer_interrupts.rs
```diff
@@ -0,0 +1,137 @@
...
+		source,
+		crate::arch::processor::get_timer_ticks() + wakeup_micros,
+	);
+}
+
+/// Crete a new timer, but with an absolute wakeup time.
+pub fn create_timer_abs(source: Source, wakeup_time: u64) {
+	let timers = &mut core_scheduler().timers;
+
+	// SAFETY: Our timer list has an entry for every possible source
```
mkroening: Why does it say `SAFETY` when there is no unsafe block here and further down?

src/arch/timer_interrupts.rs
```diff
@@ -0,0 +1,137 @@
...
+		.unwrap();
+
+	assert!(lowest_timer.wakeup_time != u64::MAX);
+
+	// TODO: Do we really want to do this here?
+	match lowest_timer.source {
+		#[cfg(feature = "net")]
+		Source::Network => NETWORK_WAKER.lock().wake(),
+		_ => {} // no-op, we always poll after a timer interrupt
+	}
```
mkroening: Did you find an answer whether to do this here? :D

src/arch/aarch64/kernel/interrupts.rs
```diff
@@ -93,10 +94,7 @@ pub(crate) fn install_handlers() {
 
 	fn timer_handler() {
 		debug!("Handle timer interrupt");
-
-		// disable timer
-		CNTP_CVAL_EL0.set(0);
-		CNTP_CTL_EL0.write(CNTP_CTL_EL0::ENABLE::CLEAR);
```
mkroening, line 100: Was this unnecessary before, or why don't we disable the timer anymore?

src/scheduler/timer_interrupts.rs
mkroening, line 1: Since this is not arch-specific, it should not be in `arch`. `arch` should only forward to the real archs.

src/drivers/net/virtio/mod.rs
```diff
@@ -417,6 +418,9 @@ impl NetworkDriver for VirtioNetDriver<Init> {
 		}
 
 		self.isr_stat.acknowledge();
+
+		trace!("Waking network waker");
```
mkroening: An alternative to duplicating the trace statement everywhere could be using [`#[track_caller]`](https://doc.rust-lang.org/reference/attributes/codegen.html#the-track_caller-attribute) to log the caller inside `WakerRegistration::wake`. I am not sure about the performance implications, though.

src/executor/mod.rs
```diff
@@ -155,101 +151,27 @@ where
...
-						delay.map(|d| crate::arch::processor::get_timer_ticks() + d),
-					);
-				}
-			}
-
 			return Err(Errno::Time);
 		}
 
-		#[cfg(feature = "net")]
+		// TODO: I have no idea whether this is correct
```
mkroening: Do you have an idea now, or should we discuss this? :)

src/executor/network.rs
```diff
@@ -189,14 +189,35 @@ async fn dhcpv4_run() {
...
 		if let Some(mut guard) = NIC.try_lock() {
 			match &mut *guard {
 				NetworkState::Initialized(nic) => {
-					nic.poll_common(now());
-					// FIXME: only wake when progress can be made
-					cx.waker().wake_by_ref();
+					let now = now();
+
+					// TODO: smoltcp is probably not exposing enough information here
+					// Well, how could it! Impossible :)
```
mkroening: Is this `TODO` still up to date?

src/executor/network.rs
```diff
@@ -255,12 +276,12 @@ pub(crate) fn init() {
 	*guard = NetworkInterface::create();
 
 	if let NetworkState::Initialized(nic) = &mut *guard {
+		// TODO: Is this really necessary?
```
mkroening: Is this `TODO` still up to date?

src/scheduler/timer_interrupts.rs
Gelbpunkt: I went with this because it has very nice properties when sorting ascending, the implementation of `Ord` for `Option` is exactly the opposite of what we'd want (`None` is always lower than `Some`). Will add a comment.

mkroening:
As of https://github.com/hermit-os/hermit-rs/pull/914, the CI should be less flaky now.

src/scheduler/timer_interrupts.rs
```diff
@@ -0,0 +1,150 @@
...
+			.iter_mut()
+			.min_by(|a, b| a.wakeup_time.cmp(&b.wakeup_time))
+			.unwrap()
+	}
+
+	/// Adjust all wakeup times by a specific offset.
+	pub fn adjust_by(&mut self, offset: u64) {
+		for timer in self.0.iter_mut() {
+			if timer.wakeup_time != u64::MAX {
+				timer.wakeup_time -= offset;
```
mkroening, line 77: [CI found](https://github.com/hermit-os/kernel/actions/runs/21710103438/job/62611196355?pr=2086) this reliably overflows:

```
$ cargo xtask ci rs --arch x86_64 --package mioudp --features hermit/dhcpv4,hermit/virtio-net qemu --devices virtio-net-pci
[..]
[    0.450189][0][WARN  entropy   ] Unable to read entropy! Fallback to a naive implementation!
You can connect to the server using `nc`:
 $ nc -u 127.0.0.1 9975
Anything you type will be echoed back to you.
[CI] send "exit" via UDP to 127.0.0.1:9975
[CI] receive: exit
[0][PANIC] panicked at src/scheduler/timer_interrupts.rs:75:17:
attempt to subtract with overflow

[    5.713705][0][INFO  interrupts] Number of interrupts
[    5.715990][0][INFO  interrupts] [0][Reschedule]: 3
[    5.716729][0][INFO  interrupts] [0][Timer]: 2
[    5.717296][0][INFO  interrupts] [0][virtio]: 3
exit status 1
[    5.718251][0][INFO  processor ] Shutting down system
```

The same happens on AArch64. I can reproduce this consistently locally as well.

Good thing we actually test sleeping in our CI now (https://github.com/hermit-os/hermit-rs/pull/914). :D

Gelbpunkt: I cannot reproduce this locally, no matter whether I target x86_64 or aarch64... Quite weird

Gelbpunkt: Anyways, I think `saturating_sub` is acceptable to use here, let's see whether it passes CI

Gelbpunkt: Okay, turns out this was a logic error I introduced in some revision. It's fixed now.

PR #2087, opened 2025-11-28
refactor(fd): remove `Debug` bound from `ObjectInterface`

mkroening:
Extracted and extended from https://github.com/hermit-os/kernel/pull/2078.

### DIFF
--- a/src/scheduler/task/mod.rs
+++ b/src/scheduler/task/mod.rs
@@ -464,7 +464,9 @@ impl Task {
 				>::with_hasher(
 					RandomState::with_seeds(0, 0, 0, 0),
 				))))
-				.unwrap();
+				// This function is called once per core and thus only once on core 0.
+				// Thus, this is the only place where we set OBJECT_MAP.
+				.unwrap_or_else(|_| unreachable!());
 			let objmap = OBJECT_MAP.get().unwrap().clone();
 			let mut guard = objmap.write();
 			if env::is_uhyve() {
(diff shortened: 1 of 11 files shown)


src/scheduler/task/mod.rs
```diff
@@ -464,7 +464,7 @@ impl Task {
 				>::with_hasher(
 					RandomState::with_seeds(0, 0, 0, 0),
 				))))
-				.unwrap();
+				.unwrap_or_else(|_| unreachable!());
```
fogti, line 469: If we declare this as `unreachable!()`, we should state a reason why that should be.

fogti (approved):

src/scheduler/task/mod.rs
mkroening: I was hesitant to include a string in the binary that we know will never be printed. Are you happy with the comment I added?

fogti: okay, the assumption / precondition of the function is that it's only called at most once for each possible argument value?

mkroening: Do you mean `Task::new_idle`? That function is called once per core and thus only once where `core_id == 0`.

fogti: Yes, that's what I concluded, but it would be nice to state it in a comment on the function.

mkroening: Are you happy now? :D

fogti: Yes.

fogti (approved):

PR #2090, opened 2025-11-28
perf(fs/mem): get rid of unnecessary `String` allocations

fogti:


src/fs/mem.rs
```diff
@@ -11,7 +11,7 @@
 
 use alloc::boxed::Box;
 use alloc::collections::BTreeMap;
-use alloc::string::String;
+use alloc::string::{String, ToString as _};
```
mkroening: Can you instead use `ToOwned` and not do an underscore import here? :)

mkroening (changes requested):
Thanks! This is great. I just have one request.

src/fs/mem.rs
fogti: Is it better now?

PR #2091, opened 2025-11-28
refactor: consistently use `to_owned` over `to_string` for `str`

mkroening:


src/fs/fuse.rs
```diff
@@ -980,9 +979,9 @@ impl FuseDirectoryHandle {
 impl ObjectInterface for FuseDirectoryHandle {
 	async fn getdents(&self, buf: &mut [MaybeUninit<u8>]) -> io::Result<usize> {
 		let path: CString = if let Some(name) = &self.name {
-			CString::new("/".to_string() + name).unwrap()
+			CString::new("/".to_owned() + name).unwrap()
 		} else {
-			CString::new("/".to_string()).unwrap()
+			CString::new("/".to_owned()).unwrap()
```
fogti:
```suggestion
			CString::new("/").unwrap()
```

mkroening: Wow, your code is not just shorter but actually better, which can't be guessed from the outside (`T: Into<Vec<u8>>`).

There is additional internal specialization: https://github.com/rust-lang/rust/blob/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/alloc/src/ffi/c_str.rs#L271-L291

fogti: incredible. TIL that it's even better than I expected.

PR #2101, opened 2025-12-01
MSI-X support

cagatay-y:
cloud-hypervisor only supports MSI-X interrupts for PCI devices, so support for MSI-X is needed to support running on it. Additionally, MSI-X  allows us to set separate interrupt handlers for the configuration change interrupts and queue updates (even per-queue handlers once we have multiple queues in the future).

~The implementation does not work on QEMU with the TAP network device (tested with `cargo xtask ci rs --arch x86_64 --profile dev  --package httpd --features ci,hermit/virtio-net,hermit/msix qemu --accel --sudo --devices virtio-net-pci --tap`) and possibly other platforms as well. It works with QEMU user network device in addition to cloud-hypervisor, by adding the `hermit/msix` feature to the regular run command.~

### DIFF
--- a/src/drivers/fs/mod.rs
+++ b/src/drivers/fs/mod.rs
@@ -36,10 +36,11 @@ use crate::drivers::mmio::get_filesystem_driver;
 use crate::drivers::pci::get_filesystem_driver;
 use crate::drivers::virtio::ControlRegisters;
 use crate::drivers::virtio::error::VirtioFsInitError;
+use crate::drivers::virtio::transport::InterruptCapability;
 #[cfg(not(feature = "pci"))]
-use crate::drivers::virtio::transport::mmio::{ComCfg, IsrStatus, NotifCfg};
+use crate::drivers::virtio::transport::mmio::{ComCfg, NotifCfg};
 #[cfg(feature = "pci")]
-use crate::drivers::virtio::transport::pci::{ComCfg, IsrStatus, NotifCfg};
+use crate::drivers::virtio::transport::pci::{ComCfg, NotifCfg};
 use crate::drivers::virtio::virtqueue::error::VirtqError;
 use crate::drivers::virtio::virtqueue::split::SplitVq;
 use crate::drivers::virtio::virtqueue::{
@@ -67,7 +68,7 @@ pub(crate) struct FsDevCfg {
 pub(crate) struct VirtioFsDriver {
 	pub(super) dev_cfg: FsDevCfg,
 	pub(super) com_cfg: ComCfg,
-	pub(super) isr_stat: IsrStatus,
+	pub(super) isr_stat: InterruptCapability,
 	pub(super) notif_cfg: NotifCfg,
 	pub(super) vqueues: Vec<VirtQueue>,
 }
@@ -159,11 +160,34 @@ impl VirtioFsDriver {
 			self.vqueues.push(vq);
 		}
 
-		handlers.entry(irq.unwrap()).or_default().push_back(|| {
-			if let Some(driver) = get_filesystem_driver() {
-				driver.lock().handle_interrupt();
-			};
-		});
+		match &mut self.isr_stat {
+			InterruptCapability::IsrStatus(_) => {
+				let irq = irq.unwrap();
+				handlers.entry(irq).or_default().push_back(|| {
+					if let Some(driver) = get_filesystem_driver() {
+						driver.lock().handle_interrupt();
+					};
+				});
+				crate::arch::kernel::interrupts::add_irq_name(irq, "virtio");
+				info!("Virtio interrupt handler at line {irq}");
+			}
+			#[cfg(all(feature = "pci", target_arch = "x86_64"))]
+			InterruptCapability::Msix(msix_table) => {
+				use core::iter;
+
+				self.com_cfg.register_msix_vectors(
+					msix_table,
+					handlers,
+					|| {
+						if let Some(driver) = get_filesystem_driver() {
+							driver.lock().handle_device_configuration_interrupt();
+						};
+					},
+					iter::empty::<(iter::Empty<_>, _)>(),
+					0..vqnum as u16,
+				);
+			}
+		}
 
 		// At this point the device is "live"
 		self.com_cfg.drv_ok();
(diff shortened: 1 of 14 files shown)


src/drivers/pci.rs
```diff
@@ -537,6 +537,30 @@ pub(crate) fn init() {
 	});
 }
 
+#[repr(C)]
+pub(crate) struct MsixEntry {
+	addr_low: u32,
+	addr_high: u32,
+	data: u32,
+	control: u32,
+}
```
mkroening: This is an entry in the MSI-X table, yes? It might make sense to document that more explicitly:


```suggestion
/// MSI-X Table entry.
#[repr(C)]
pub(crate) struct MsixTableEntry {
    /// Message Address
	message_address: u32,
	
	/// Message Upper Address
	message_upper_address: u32,
	
	/// Message Data
	message_data: u32,
	
	/// Vector Control
	vector_control: u32,
}
```

mkroening: If we need volatile access here, we could also use [`volatile::VolatileFieldAccess`](https://docs.rs/volatile/0.6.1/volatile/derive.VolatileFieldAccess.html) to make accessing fields more ergonomic.

mkroening: Would it make sense to upstream this to [`pci_types`](https://github.com/rust-osdev/pci_types)?

That crate currently does not use any volatile operations at all, though. Is volatile necessary here? Should `pci_types` use volatile for some current operations? Or does this live in the PCI config region and should be accessed via [`pci_types::ConfigRegionAccess`](https://docs.rs/pci_types/latest/pci_types/trait.ConfigRegionAccess.html) instead?

src/drivers/net/virtio/mod.rs
```diff
@@ -737,11 +741,33 @@ impl VirtioNetDriver<Uninit> {
 		}
 		debug!("{:?}", self.checksums);
 
+		// If self.irq is some, it was filled with a legacy interrupt number which we prefer over MSI-X.
```
mkroening: In a world where MSI-X works everywhere, would we still want to prefer legacy interrupts?

src/drivers/net/virtio/pci.rs
```diff
@@ -44,14 +46,21 @@ impl VirtioNetDriver<Uninit> {
 			return Err(error::VirtioNetError::NoDevCfg(device_id));
 		};
 
+		let irq = device.get_irq();
+		if irq.is_none() {
+			warn!("No interrupt lanes found for virtio-net.");
```
mkroening: Should this not be an error instead?

src/drivers/virtio/transport/pci.rs
```diff
@@ -256,6 +262,19 @@ impl VqCfgHandler<'_> {
 			.write(addr.as_u64().into());
 	}
 
+	#[cfg(all(
+		feature = "virtio-net",
+		not(feature = "rtl8139"),
+		target_arch = "x86_64"
+	))]
```
mkroening: Why gate this on `not(feature = "rtl8139")`?
And why gate this on `feature = "virtio-net"`? Should we not enable it for all virtio devices?

mkroening (changes requested):
Thanks! :)

Could you also add how to run this with QEMU user networking and cloud hypervisor to the PR description?

Is this tested in CI already?

src/drivers/net/virtio/pci.rs
cagatay-y: According to the section 7.5.1.1.13 of the PCIe specification version 6.0, "A [Interrupt Pin Register] value of 00h indicates that the Function uses no legacy interrupt Message(s)," so a device may be well-behaving and still not have an interrupt pin and line, in which case we may still be able to use MSI-X. There are reserved and unknown values and values that indicate no connection to the interrupt controller but for `get_irq` we do not differentiate between these and return None just like we do for the _no legacy interrupt message(s)_ case.

src/drivers/pci.rs
cagatay-y: It is part of the PCI-e specification, so I may make sense to.

Volatile is necessary and `ConfigRegionAccess` does not apply as it is an object that is mapped to the BAR space. I am not aware of any `pci_types` operations that are missing volatile accesses.

src/drivers/net/virtio/pci.rs
cagatay-y: We could change it to also check for the existence of the MSI-X capability and print an error if both are unavailable.

src/drivers/pci.rs
```diff
@@ -463,6 +463,67 @@ pub(crate) fn init(handlers: &mut InterruptHandlerMap) {
...
+		) {
+			use bit_field::BitField;
+
+			// Intel 64 and IA-32 Architectures Software Developer’s Manual volume 2 section 12.11.2.1.
+			core::assert_matches!(vector, 0x10..=VECTOR_MAX);
+
+			let msix_entry = unsafe {
+				msix_table
+					.as_mut_ptr()
+					.map(|table| table.cast::<_>().offset(index.try_into().unwrap()))
```
mkroening: Either specify the type or remove the turbofish:

```suggestion
					.map(|table| table.cast().offset(index.try_into().unwrap()))
```

src/drivers/fs/mod.rs
```diff
@@ -159,11 +160,32 @@ impl VirtioFsDriver {
...
+			InterruptCapability::Msix(msix_table) => {
+				self.com_cfg.register_msix_vectors(
+					msix_table,
+					handlers,
+					|| {
+						if let Some(driver) = get_filesystem_driver() {
+							driver.lock().handle_device_configuration_interrupt();
+						};
+					},
+					super::pci::msix::NO_QUEUE_HANDLERS.into_iter(),
```
mkroening: I think we can remove the constant and use a zero-sized iterator instead:

```suggestion
					iter::empty(),
```

src/drivers/pci.rs
```diff
@@ -463,6 +463,67 @@ pub(crate) fn init(handlers: &mut InterruptHandlerMap) {
...
+		pub(crate) vector_control: u32,
+	}
+
+	impl TableEntry {
+		#[cfg(target_arch = "x86_64")]
+		pub fn configure(
+			msix_table: &mut volatile::VolatileRef<'_, [Self]>,
+			index: u16,
+			vector: u8,
+		) {
```
mkroening: Could you add a small doc comment?

mkroening (changes requested):
Thanks, this looks great!

Please ensure that MSI-X is written consistently in code (`MsiX` and `msi_x`).

Also, this seems to be tested successfully with the current CI, right? Could you also add an info log that says whether we are currently using legacy interrupts or MSI-X? :)

src/drivers/pci.rs
mkroening: Maybe it would make sense to make this a method taking `self` instead, by implementing an extension trait over `VolatilePtr<'_, [TableEntry], WriteOnly>`?

src/drivers/fs/mod.rs
cagatay-y: This still requires something like `iter::empty::<(iter::Empty<_>, _)>()` because the compiler cannot deduce a type.

mkroening: I see. Let's do that, then.

mkroening (approved):
Thank you! 🥳 

The feature and the speedup are great!

PR #2132, opened 2025-12-11
build(`xtask`): reduce dependencies in non-CI usecases

fogti:
This puts the `ci` subcommand behind a feature flag (enabled by default for backward compatibility). In particular, this should make it possible to massively reduce the amount of dependencies which have to be built in order to build the hermit kernel as part of the `hermit` build script (previously 101, now 33, for `xtask`).

This also gets rid of the `flate2` dependency entirely.

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -3,13 +3,17 @@ name = "xtask"
 version = "0.1.0"
 edition = "2024"
 
+[features]
+default = ["ci"]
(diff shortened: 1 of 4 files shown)


fogti:
(this would also subsequently massively reduce my concerns about the kernel `xtask` package not being a library which can be linked into build scripts)

mkroening (approved):
Thanks! :)

PR #2138, opened 2025-12-16
CITATION.cff: Removed license

jounathaen:
The goal is to work around the Zenodo limitation on citation.cff parsing: https://github.com/zenodo/zenodo/issues/2515

This does obviously not affect the license of the project, and we can still set the license in Zenodo by hand.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -66,9 +66,6 @@ keywords:
   - operating system
   - high-performance computing
   - cloud computing
-license:
-  - MIT
-  - Apache-2.0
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
This would mean for every version we would need to add both licenses, right?

Would it make sense to go with MIT only instead for this file? Then, we have less to add, and the default would still be usable.

mkroening:
> I'm not 100% sure, but I think that Zenodo pre-fills missing metadata from the previous release.

Ah, and it would not do that if we set a license here explicitly?

PR #2139, opened 2025-12-21
drivers: Modified code in virtio transport

udaykiriti:
I've made some small, improvements to the virtio drivers. :| This commit optimizes queue selection for both MMIO and PCI to skip redundant register writes.


mkroening (changes requested):
This does not work at this level, since only one virtq can be selected at the same time, but `self.selected` is not set to `false` when other virtqs are selected.

mkroening:
The refactor required to properly support this would be rather big and nontrivial. I am not sure if it would be worth it. Are you okay with closing the PR or what would you prefer? :)

PR #2172, opened 2026-01-10
refactor(fs): refactor path traversal handling

fogti:
I realized that by "just" factoring out all path traversal handling, I don't even need to introduce an `std::path::Path` equivalent.

The afaik only disadvantage that this has is the re-allocations during traversal into `fuse` and `uhyve` filesystems.

### DIFF
--- a/src/fs/fuse.rs
+++ b/src/fs/fuse.rs
@@ -1,5 +1,6 @@
 use alloc::borrow::ToOwned;
 use alloc::boxed::Box;
+use alloc::collections::btree_set::BTreeSet;
 use alloc::ffi::CString;
 use alloc::string::String;
 use alloc::sync::Arc;
@@ -1093,9 +1094,13 @@ impl ObjectInterface for FuseDirectoryHandle {
 	}
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
+// The `original_prefix` is the one the directory was originally initialized with,
+// `prefix` is the current prefix (normally, `prefix` has `original_prefix` as its prefix).
+// This distinction is used for symlink resolution and directory traversal.
 pub(crate) struct FuseDirectory {
-	prefix: Option<String>,
+	original_prefix: Arc<str>,
+	prefix: String,
 	attr: FileAttr,
 }
 
@@ -1105,7 +1110,8 @@ impl FuseDirectory {
 		let t = timespec::from_usec(microseconds as i64);
 
 		FuseDirectory {
-			prefix,
+			original_prefix: Arc::from(prefix.as_deref().unwrap_or_default()),
+			prefix: prefix.unwrap_or_default(),
 			attr: FileAttr {
 				st_mode: AccessPermission::from_bits(0o777).unwrap() | AccessPermission::S_IFDIR,
 				st_atim: t,
@@ -1116,17 +1122,22 @@ impl FuseDirectory {
 		}
 	}
 
-	fn traversal_path(&self, components: &[&str]) -> CString {
-		let prefix_deref = self.prefix.as_deref();
-		let components_with_prefix = prefix_deref.iter().chain(components.iter().rev());
-		let path: String = components_with_prefix
-			.flat_map(|component| ["/", component])
-			.collect();
-		if path.is_empty() {
-			CString::new("/").unwrap()
-		} else {
-			CString::new(path).unwrap()
+	fn traversal_this(&self) -> CString {
+		let mut path = String::new();
+		path.push('/');
+		path.push_str(&self.prefix);
+		CString::new(path).unwrap()
+	}
+
+	fn traversal_path(&self, component: &str) -> CString {
+		let mut path = String::new();
+		if !self.prefix.is_empty() {
+			path.push('/');
+			path.push_str(&self.prefix);
 		}
+		path.push('/');
+		path.push_str(component);
+		CString::new(path).unwrap()
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/fs/fuse.rs
```diff
@@ -1093,61 +1093,78 @@ impl ObjectInterface for FuseDirectoryHandle {
...
-#[derive(Debug)]
+#[derive(Clone, Debug)]
+// The `original_prefix` is the one the directory was originally initialized with,
+// `prefix` is the current prefix (normally, `prefix` has `original_prefix` as its prefix).
+// This distinction is used for symlink resolution and directory traversal.
 pub(crate) struct FuseDirectory {
-	prefix: Option<String>,
-	attr: FileAttr,
+	original_prefix: Arc<str>,
+	prefix: String,
```
fogti, line 1103: It might be a good idea to instead only store the part in `prefix` that is actually "beyond" `original_prefix`.

src/fs/fuse.rs
```diff
@@ -1226,36 +1243,54 @@ impl VfsNode for FuseDirectory {
 		Ok(entries)
 	}
 
-	fn traverse_stat(&self, components: &mut Vec<&str>) -> io::Result<FileAttr> {
-		let path = self.traversal_path(components);
+	async fn stat(&self) -> io::Result<FileAttr> {
```
fogti: I had to change this method from a recursion to a loop to avoid having to allocate a `Box` for the `.stat` (because otherwise calling `.await` on it inside of the function would lead to an infinitely-sized closure).

src/fs/mod.rs
```diff
@@ -50,77 +50,73 @@ pub(crate) enum NodeKind {
...
 		Err(Errno::Nosys)
 	}
 
-	/// Determine the syscall interface
-	fn get_object(&self) -> io::Result<Arc<async_lock::RwLock<dyn ObjectInterface>>> {
+	/// Duplicate an object, if possible
+	fn dup(&self) -> Box<dyn VfsNode>;
+
+	/// Traverse into a subdirectory or file
+	async fn traverse_once(&self, _component: &str) -> io::Result<Box<dyn VfsNode>> {
```
fogti: TODO: introduce `traverse_multiple`.

src/fs/fuse.rs
```diff
@@ -1093,61 +1093,99 @@ impl ObjectInterface for FuseDirectoryHandle {
...
+			if let Some(x) = path.strip_prefix("/") {
+				path = x;
+			} else {
+				return Err(Errno::Nosys);
+			}
+			if !prefix.is_empty() {
+				prefix.push('/');
+			}
+		}
+		prefix.push_str(path);
```
fogti, line 1195: This is new, and should be specifically reviewed.

mkroening (changes requested):
Thanks for the PR! :)

PRs are very difficult to review when one commit changes so much at once. I amended two commits to make discussing the main change of this more reviewable. I also extracted a few changes into another PR (https://github.com/hermit-os/kernel/pull/2229) to merge them separately. Small commits not only help the reviewer with comprehending the diff but also help Git show nicer diffs, since each diff has more anchor points that did not change.

I'll try to summarize the change to discuss it. Please add this summary or a corrected version if I got something wrong to the PR description. This helps immensely with understanding the changes and their motivation and is a necessary foundation for discussing the overall approach. Any alternatives that have been considered before opening the PR are also important to put into the PR description. Ultimately, PR authors should do a self-review before submitting a PR to ensure reviewability.

Extracting the core change, understanding it, and guessing the motivation took me a lot of time. For you, providing this info would have been much easier, since you know the ins and outs of the changes and your motivation. Please make sure to include these things in future PRs. Otherwise they will not get reviewed or merged.

Something like this should go into the PR description:

### PR Description

Currently, `VfsNode` has many functions that do recursive traversal as well as their respective operation:

- `traverse_mkdir`
- `traverse_rmdir`
- `traverse_unlink`
- `traverse_readdir`
- `traverse_lstat`
- `traverse_stat`
- `traverse_mount`
- `traverse_open`
- `traverse_create_file`

This PR makes these functions more maintainable by separating VFS traversal from the actual operations, heavily deduplicating path traversal logic.

This is done by adding three new methods to `VfsNode`: 

- ```rust
  fn traverse_once(&self, component: &str) -> io::Result<Box<dyn VfsNode>>;
  ```
- ```rust
  fn traverse_multiple(&self, component: &str) -> io::Result<Box<dyn VfsNode>>;
  ```
- ```rust
  fn dup(&self) -> Box<dyn VfsNode>;
  ```

`dup` clones the object like `Clone::clone` but returns `Box<dyn VfsNode>` instead of `Self` to remain dyn-compatible.

`dup` is required since subnodes may be a different `dyn VfsNode` that we return by value in the traversal functions.

This allows us for each operation to first traverse the VFS and then do the operation separately, successfully decoupling the two. Thus, this PR correspondingly changes all operation methods in the following way:

```rust
fn traverse_operation(&self, components: &mut Vec<&str>) -> io::Result<ReturnValue>
```

becomes

```rust
fn operation(&self, component: &str) -> io::Result<ReturnValue>
```

### PR Review

Is `dup` not the same as `Clone::clone` in all current implementations? Why invent a new name? If this really is necessary, using the [`dyn_clone`](https://docs.rs/dyn-clone) crate would be preferable.

More generally, I would like to avoid the reallocations and not have a `dup` method or use `dyn_clone`. This means not returning something via ownership but working on references.

What about still doing traversal recursively like before but also decoupling traversal from the operations?

Ideally, this would look something like this:

```rust
fn traverse_with<F, R>(&self, path: &str, f: F) -> io::Result<R>
where
    F: FnOnce(&dyn VfsNode) -> R;
```

Unfortunately, this would not be dyn-safe. This could be viable in the future, though if we chose to abandon `dyn VfsNode` in favor of an `enum VfsNode`, similar to what is currently proposed in https://github.com/hermit-os/kernel/pull/2096. I need to discuss that PR internally to see whether this would also be viable here.

Alternatively, we could create an `enum VfsOp` and `enum VfsOpRet` for VFS operations that we could use while not migrating away from `dyn VfsNode`. This does not feel very ergonomic to me, though.

If `components: &mut Vec<&str>` is the main thing that you want to clean up, we can easily merge a PR that migrates this to `components: &[&str]` or `path: &str` instead.

src/fs/fuse.rs
```diff
@@ -1142,12 +1153,56 @@ impl VfsNode for FuseDirectory {
...
+
+		Ok(Box::new(Self {
+			original_prefix: Arc::clone(&self.original_prefix),
+			prefix,
+			attr: todo!(),
```
fogti, line 1200: Why not copy the original `attr` down?

Otherwise, we never really use this `attr` anyways, so why keep it around?

mkroening: This commit is not proposing we do it like this. I just reduced your changes so I can understand what and why you are doing them. You are welcome to modify as necessary. Just keep your individual commits minimal, please.

fogti:
That is a good point (`Arc` instead of `Box`). Doing so would be even more complicated because for that we currently have the `Arc`'s in the hierarchy at the wrong places, so in contrast to the current state of this PR, it would involve actually restructuring the involved directory structures to a large extent.

I can do that.

I want to primarily avoid 2 things:
1. Holding of a lock guard per traversed directory along the path simultaneously. This is imo more important than avoiding recursion itself, particularly because the factoring-out of `block_on` to the top-level filesystem entry points have alleviated concerns about potential scheduling churn, etc.
2. Avoiding the recursion itself: This is beneficial because it is easy to do and imo makes the code more readable, and it allows omitting an allocation for the recursive async closures, which would need boxing to be able to `await` them.

I also want to note that albeit this introduces an allocation for `dup`-in (cloning) references to directory / file nodes, this replaces (via 2.) a stack of nested allocations (not even tail-call optimization can help with that, given how convoluted some of the traversal functions were, and that they need to refer to the stack-bound lock guards which have non-trivial destructors) with short-lived allocations for `dup`-ed references.

PR #2185, opened 2026-01-20
refactor(arch/*/kernel): factor out duplicated code

fogti:
This is just a suggestion, but I found it way easier to understand code when the parts that are actually 100% identical aren't duplicated in per-architecture code.


fogti:
Might be worth to revisit later, but afaik needs more coordination about what to do further first.

PR #2190, opened 2026-01-23
feat(uhyve): use v2 interface

n0toose:
On x86_64, the data address will be stored in the rcx register to support 64-bit addresses.

This also removes the usage of ExitArgs, as it is a single integer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -338,7 +338,7 @@ talc = { version = "5" }
 thiserror = { version = "2", default-features = false }
 time = { version = "0.3", default-features = false }
 volatile = "0.6"
-uhyve-interface = "0.1.4"
+uhyve-interface = "0.2"
 
 [dependencies.smoltcp]
 version = "0.13"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -211,6 +211,7 @@ exclude = [
 ]
 
 [patch.crates-io]
+uhyve-interface = { git = "https://github.com/n0toose/uhyve", branch = "uhyve-if-v2_io-plus-rcx" }
```
n0toose: TODO: Remove this before merging, update `Cargo.toml` and `Cargo.lock` accordingly.

Cargo.toml
```diff
@@ -416,6 +416,7 @@ exclude = [
 ]
 
 [patch.crates-io]
+uhyve-interface = { git = "https://github.com/hermit-os/uhyve" }
```
jounathaen: I hope this is now obsolete with the release of uhyve-if 2.0

Cargo.toml
```diff
@@ -347,7 +347,7 @@ thiserror = { version = "2", default-features = false }
 time = { version = "0.3", default-features = false }
 volatile = "0.6"
 zerocopy = { version = "0.8", default-features = false }
-uhyve-interface = "0.1.4"
+uhyve-interface = "0.2.0"
```
mkroening: Nit:

```suggestion
uhyve-interface = "0.2"
```

That way, the Cargo.toml gets touched less often.

jounathaen:
I've taken the liberty to rebase this :innocent:

PR #2191, opened 2026-01-26
build(deps): update uart_16550 to 0.5

mkroening:
See https://github.com/rust-osdev/uart_16550/pull/41.

### DIFF
--- a/src/arch/x86_64/kernel/serial.rs
+++ b/src/arch/x86_64/kernel/serial.rs
@@ -2,6 +2,8 @@ use alloc::collections::VecDeque;
 
 use embedded_io::{ErrorType, Read, ReadReady, Write};
 use hermit_sync::{InterruptTicketMutex, Lazy};
+use uart_16550::backend::PioBackend;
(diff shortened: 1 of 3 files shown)


src/arch/x86_64/kernel/serial.rs
```diff
@@ -27,8 +30,10 @@ impl UartDevice {
 			.serial_port_base
 			.unwrap()
 			.get();
-		let mut uart = unsafe { uart_16550::SerialPort::new(base) };
-		uart.init();
+		let mut uart = unsafe { Uart16550::new_port(base).unwrap() };
+		uart.init(Config::default()).unwrap();
+		uart.test_loopback().unwrap();
+		uart.check_connected().unwrap();
```
phip1611: Please note that this will most likely fail on real hardware. Most null modem cables (for point-to-point communication) ignore the necessary lanes

mkroening: Yeah, thanks for pointing that out! I removed the checks entirely, since we don't have any fallback at the moment.

PR #2214, opened 2026-02-03
paging: make ALL levels writable, not just P4

zyuiop:
After investigating PF exceptions, I traced them back to a `pt.set_unused()` call, which failed because the page table was not entirely writable.


fogti:
Wait, is the title of this PR misleading and this does make all levels including-and-below P4 writable, but doesn't touch anything else?

zyuiop:
> This PR is still useful for you in the meantime, right?

I have it in my fork as well, so it's okay!

PR #2218, opened 2026-02-04
fix(socket/tcp): when polling listening, poll all sockets

zyuiop:
A connection may be established on something that is not the first in the list (because of race conditions with accept).

If that happens, `poll` will not report it and the connection will remain stuck.

Typical symptom:
A benchmark (wrk, ...) with multiple connections runs fine, but suddenly throughput drops to 0.


mkroening (approved):
Thanks! :)

PR #2249, opened 2026-02-11
refactor: do more early returns

mkroening:
This PR makes the kernel use early returns in more places as appropriate. This makes the code more readable, as we avoid indentation creep and don't put the happy path into `if`s. Handling the exceptional cases earlier and be done with them should also be easier for the brain's stack when reading this code.

This is a rather big commit because of the indentation changes. Semantically, the changes should be much smaller. All changes should be independent.


fogti:
This PR introduced at least one bug: (see fix PR #2301).

PR #2255, opened 2026-02-16
fix(mmio): don't return MMIO virtual addresses to the free list (use-after-free)

edef1c:
In check_linux_args, we return VAs to the free list while a VolatileRef to them is active. This breaks multi-device configurations. A similar bug exists in guess_device, but the current version of this patch doesn't fix it, since my reference target supplies mmio args.

### DIFF
--- a/src/arch/x86_64/kernel/mmio.rs
+++ b/src/arch/x86_64/kernel/mmio.rs
@@ -98,8 +98,8 @@ fn check_linux_args(
 	linux_mmio: &'static [String],
 ) -> Vec<(VolatileRef<'static, DeviceRegisters>, u8)> {
 	let layout = PageLayout::from_size(BasePageSize::SIZE as usize).unwrap();
-	let page_range = PageBox::new(layout).unwrap();
-	let virtual_address = VirtAddr::from(page_range.start());
+	let mut page_range = PageBox::new(layout).unwrap();
+	let mut virtual_address = VirtAddr::from(page_range.start());
 
 	let mut devices = vec![];
 	for arg in linux_mmio {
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
Thanks, this looks good! :)

Can you share how to reproduce this? Are you using Firecracker or something else?

> A similar bug exists in guess_device

Fixing that one would be good too. I can reproduce that as follows:

```bash
cargo xtask ci rs --arch x86_64 --package httpd --no-default-features --features ci,hermit/dhcpv4,hermit/tcp,hermit/virtio-net,hermit/virtio-console qemu --microvm --devices virtio-console-mmio --devices virtio-net-mmio
```

Swapping the devices makes the difference.

> FWIW, it is not totally clear to me why PageBox merely reserves the VA, and leaks the mapping / doesn't unmap on Drop; the VA use-after-free would have been dead obvious in all cases with a more RAII-style design.

That's a good point. I was planning to do that eventually, but I did not have time for that yet. I have opened https://github.com/hermit-os/kernel/issues/2338 for tracking.

Eventually, I will unify and clean up MMIO and PCI device detection across architectures, which should help with these legacy code issues in the future.

cagatay-y:
> Fixing that one would be good too. I can reproduce that as follows:
> 
> ```shell
> cargo xtask ci rs --arch x86_64 --package httpd --no-default-features --features ci,hermit/dhcpv4,hermit/tcp,hermit/virtio-net,hermit/virtio-console qemu --microvm --devices virtio-console-mmio --devices virtio-net-mmio
> ```
> 
> Swapping the devices makes the difference.

This turned out to be caused by another issue and was fixed in 0d156dec5a4cc4048bcfb3e982cae85f7b722920.

mkroening:
I fixed this in https://github.com/hermit-os/kernel/pull/2619. Thanks again for raising this issue! :)

PR #2263, opened 2026-02-18
Batch console output to one VM exit per print call

stlankes:
Increase serial buffer from 256 to 2048 bytes and gate newline-triggered flush on debug_assertions. In release builds, data is only flushed when the buffer is full or explicitly at the end of each _print()/_panic_print() call. This reduces VM exits from ~28 per multi-line info!() to just 1.

I cherry-pick a PR from @jbreitbart's repository.

### DIFF
--- a/src/console.rs
+++ b/src/console.rs
@@ -14,7 +14,7 @@ use crate::executor::WakerRegistration;
 #[cfg(not(target_arch = "riscv64"))]
 use crate::syscalls::interfaces::serial_buf_hypercall;
 
-const SERIAL_BUFFER_SIZE: usize = 256;
+const SERIAL_BUFFER_SIZE: usize = 2048;
 
 pub(crate) enum IoDevice {
 	#[cfg(not(target_arch = "riscv64"))]
@@ -162,6 +162,7 @@ impl Write for Console {
 		if SERIAL_BUFFER_SIZE - self.buffer.len() >= buf.len() {
 			// unwrap: we checked that buf fits in self.buffer
 			self.buffer.extend_from_slice(buf).unwrap();
+			#[cfg(debug_assertions)]
 			if buf.contains(&b'\n') {
 				self.flush()?;
 			}
@@ -173,6 +174,7 @@ impl Write for Console {
 			} else {
 				// unwrap: we checked that buf fits in self.buffer
 				self.buffer.extend_from_slice(buf).unwrap();
+				#[cfg(debug_assertions)]
 				if buf.contains(&b'\n') {
 					self.flush()?;
 				}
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
I have investigated how buffering works in glibc and Rust.

```c
// test.c
#include <stdio.h>
#include <unistd.h>

void wait(void) {
    size_t n = 0x40000000;
    for (size_t i = 0; i < n; ++i) {
        asm("");
    }
}

int main(void) {
    // Does buffer
    printf("printf");
    wait();
    printf(" does buffer\n");
    wait();
    
    // Does buffer
    fprintf(stdout, "fprintf(stdout)");
    wait();
    fprintf(stdout, " does buffer\n");
    wait();

    // Does not buffer
    fprintf(stderr, "fprintf(stderr)");
    wait();
    fprintf(stderr, " does not buffer\n");
    wait();

    // Does not buffer
    write(STDOUT_FILENO, "write(STDOUT_FILENO)", 20);
    wait();
    write(STDOUT_FILENO, " does not buffer\n", 17);
    wait();

    // Does not buffer
    write(STDERR_FILENO, "write(STDERR_FILENO)", 20);
    wait();
    write(STDERR_FILENO, " does not buffer\n", 17);
    wait();
}
```

Run with `gcc test.c && ./a.out`.

```rust
#!/usr/bin/env -S cargo +nightly -Zscript
---cargo
[dependencies]
log = "0.4"
env_logger = "0.11"
---

use std::io::{self, Write};

use log::error;

fn main() {
    // Does buffer
    print!("print!");
    wait();
    println!(" does buffer");
    wait();

    {
        // Does buffer
        let stdout = io::stdout();
        let mut handle = stdout.lock();

        handle.write_all(b"stdout").unwrap();
        wait();
        handle.write_all(b" does buffer\n").unwrap();
        wait();
    }

    // Does not buffer
    eprint!("eprint!");
    wait();
    eprintln!(" does not buffer");
    wait();

    {
        // Does not buffer
        let stderr = io::stderr();
        let mut handle = stderr.lock();

        handle.write_all(b"stderr").unwrap();
        wait();
        handle.write_all(b" does not buffer\n").unwrap();
        wait();
    }

    env_logger::init();

    // Does buffer
    error!(
        "log\n{}does\n{}buffer\n",
        {
            wait();
            '1'
        },
        {
            wait();
            '2'
        }
    );
}

fn wait() {
    let n = 0x4000000_u64;
    for _ in 0..n {
        std::hint::spin_loop();
    }
}
```

Run with `chmod +x test.rs && ./test.rs`.

These results seem sensible to me, and I think we should try to reproduce them with Hermit.

Specifically, this means that `sys_write` calls should _never_ buffer. Currently, `sys_write(STDOUT_FILENO)` and `sys_write(STDERR_FILENO)` are buffered because we run them through the same console buffer, which we should change. `fprintf(stdout)` (`printf`), `stdout` (`print!`), and `log` buffering should happen in userspace. All stdout and stderr output (especially non-text-based output) via `sys_write` should never be subject to text-based buffering.

Inside the kernel, we should adopt the userspace logic as necessary.

That means we should remove the buffering from the console abstraction and move them into kernel-internal stdout and stderr abstractions that buffer stdout appropriately and don't buffer stderr. For stdout buffering, we should always flush on newlines. We could use a buffer of 8192 there. Glibc defines [stdio.h's `BUFSIZ`](https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/stdio.h;h=3bf6a1f63203b111b67b6f101cc188320d3db78c;hb=f762ccf84f122d1354f103a151cba8bde797d521#l103) as 8192. Rust defines [`sys::io`'s `DEFAULT_BUF_SIZE`](https://github.com/rust-lang/rust/blob/1.93.1/library/std/src/sys/io/mod.rs#L59-L61) as `8 * 1024`. Musl defines [stdio.h's `BUFSIZ`](https://git.musl-libc.org/cgit/musl/tree/include/stdio.h?h=v1.2.5#n50) as 1024.

For logging, we should add another buffer in our logging implementation similar to [env_logger/src/writer/buffer.rs](https://github.com/rust-cli/env_logger/blob/v0.11.9/src/writer/buffer.rs).

PR #2275, opened 2026-02-20
 feat(network): support packet capture file creation

cagatay-y:
Allows creating packet capture files in the pcap format.

The feature also allows us to get nice graphs by using Wireshark's built-in tools.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,6 +164,15 @@ dns = ["net", "smoltcp", "smoltcp/socket-dns"]
 ## Enables pretty-printing the network traffic to the serial device.
 net-trace = ["smoltcp?/log", "smoltcp?/verbose"]
 
+## Enables the generation of packet capture files.
+##
+## The guest path is specified via the `HERMIT_PCAP_PATH` *runtime* environment variable. If a
+## complete path is not provided, values without a `/` are interpreted as file names and ones that
+## end with a `/` as directories. The default directory is `/root/` and the default file name is the
+## network device name. If a file already exists at the path, the file name is suffixed with
+## successive numbers.
+write-pcap-file = []
+
 #! ### Network Drivers
 #!
 #! Currently, Hermit does not support multiple network drivers at the same time.
--- a/src/executor/device.rs
+++ b/src/executor/device.rs
@@ -6,6 +6,8 @@ use smoltcp::iface::{Config, Interface, SocketSet};
 #[cfg(feature = "net-trace")]
 use smoltcp::phy::Tracer;
 use smoltcp::phy::{Device, Medium};
+#[cfg(feature = "write-pcap-file")]
+use smoltcp::phy::{PcapMode, PcapWriter};
 #[cfg(feature = "dhcpv4")]
 use smoltcp::socket::dhcpv4;
 #[cfg(all(feature = "dns", not(feature = "dhcpv4")))]
@@ -16,6 +18,8 @@ use smoltcp::wire::{IpCidr, Ipv4Address, Ipv4Cidr};
 
 use super::network::{NetworkInterface, NetworkState};
 use crate::arch;
+#[cfg(feature = "write-pcap-file")]
+use crate::drivers::Driver;
 use crate::drivers::net::NetworkDriver;
 
 cfg_select! {
@@ -42,18 +46,29 @@ impl<'a> NetworkInterface<'a> {
 				feature = "rtl8139",
 				feature = "virtio-net",
 			) => {
-				#[cfg_attr(feature = "net-trace", expect(unused_mut))]
+				#[cfg_attr(any(feature = "net-trace", feature = "write-pcap-file"), expect(unused_mut))]
 				let Some(mut device) = NETWORK_DEVICE.lock().take() else {
 					return NetworkState::InitializationFailed;
 				};
 			}
 			_ => {
-				#[cfg_attr(feature = "net-trace", expect(unused_mut))]
+				#[cfg_attr(any(feature = "net-trace", feature = "write-pcap-file"), expect(unused_mut))]
 				let mut device = LoopbackDriver::new();
 			}
 		}
 
 		let mac = device.get_mac_address();
+
+		#[cfg_attr(feature = "net-trace", expect(unused_mut))]
+		#[cfg(feature = "write-pcap-file")]
+		let mut device = {
+			let default_name = device.get_name();
+			PcapWriter::new(
+				device,
+				pcap_writer::FileSink::new(default_name),
+				PcapMode::Both,
+			)
+		};
 
 		#[cfg(feature = "net-trace")]
 		let mut device = Tracer::new(device, |timestamp, printer| trace!("{timestamp} {printer}"));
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -157,6 +157,10 @@ dns = ["net", "smoltcp", "smoltcp/socket-dns"]
 ## Enables pretty-printing the network traffic to the serial device.
 net-trace = ["smoltcp?/log", "smoltcp?/verbose"]
 
+## Enables generation of packet capture files.
```
mkroening: Can you document where the files will be stored?

src/executor/network.rs
```diff
@@ -63,13 +66,20 @@ static LOCAL_ENDPOINT: AtomicU16 = AtomicU16::new(0);
...
 
+cfg_if::cfg_if! {
+	if #[cfg(feature = "net-trace")] {
+		type InterfaceDevice = smoltcp::phy::Tracer<NetworkDevice>;
+	} else if #[cfg(feature = "pcap")] {
+		type InterfaceDevice = smoltcp::phy::PcapWriter<NetworkDevice, crate::executor::device::FileSink>;
+	} else {
+		type InterfaceDevice = NetworkDevice;
+	}
+}
```
mkroening: What if I want both `net-trace` and `pcap`?

Cargo.toml
```diff
@@ -157,6 +157,10 @@ dns = ["net", "smoltcp", "smoltcp/socket-dns"]
...
+pcap = ["net", "smoltcp", "virtio-fs"]
```
mkroening: Maybe `write-pcap-file` would be easier to understand? Since this does not enable any `libpcap`-like API.

src/executor/network.rs
```diff
@@ -45,7 +45,10 @@ pub(crate) enum NetworkState<'a> {
 	feature = "virtio-net",
 ))]
 pub(crate) fn network_handler() {
-	NIC.lock().as_nic_mut().unwrap().handle_interrupt();
+	// It is possible for us to receive interrupts before we are done with initializing the network interface.
+	if let Ok(nic) = NIC.lock().as_nic_mut() {
+		nic.handle_interrupt();
+	}
```
mkroening, line 52: Is this because virtio-fs interrupts arrive at the same interrupt lane as virtio-net interrupts? Documenting that reason would make sense.

src/executor/device.rs
```diff
@@ -130,3 +152,21 @@ impl<'a> NetworkInterface<'a> {
...
 }
+
+#[cfg(feature = "pcap")]
+pub(in crate::executor) struct FileSink(File);
+
+#[cfg(feature = "pcap")]
+impl PcapSink for FileSink {
+	fn write(&mut self, data: &[u8]) {
+		if let Some(err) = self.0.write(data).err() {
+			error!("Error while writing to the packet capture: {err}");
```
mkroening:
```suggestion
			error!("Error while writing to the pcap file: {err}");
```

src/executor/device.rs
```diff
@@ -130,3 +152,21 @@ impl<'a> NetworkInterface<'a> {
...
+		}
+	}
+
+	fn flush(&mut self) {
+		if let Some(err) = self.0.flush().err() {
+			error!("Error while flushing the packet capture: {err}");
```
mkroening:
```suggestion
			error!("Error while flushing the pcap file: {err}");
```

src/executor/device.rs
```diff
@@ -42,20 +48,36 @@ impl<'a> NetworkInterface<'a> {
...
+		cfg_if! {
+			if #[cfg(feature = "net-trace")] {
+				let mut device =
+					Tracer::new(device, |timestamp, printer| trace!("{timestamp} {printer}"));
+			} else if #[cfg(feature = "pcap")] {
+				let mut device = PcapWriter::new(
+					device,
+					FileSink(
+						File::create(
+							format!("/root/{}.pcap", crate::executor::network::now().secs())
```
mkroening: Is it very useful to automatically choose a name depending on the time?

It might be useful to make this configurable at build time via environment variables.

Cargo.toml
mkroening: Any other FS backend would also work, would it not? It might also be sensible to not enable `net` and `smoltcp` automatically. How about this?


```suggestion
pcap = []
```

src/executor/network.rs
cagatay-y: I changed the code so that both can be used at the same time.

src/executor/network.rs
cagatay-y: I thought that it was simply about timing but you are right, it is because of that. It is documented now.

src/executor/device.rs
cagatay-y: I chose it as a simple unique name source but now it tries to create and if the file already exists, a number is appended to the provided name. By default the device name is used.

Cargo.toml
cagatay-y: I think it only makes sense for filesystems that are mounted to the host but if you meant them, yes. It now just tries to write under "/root" (the mount point) and if that directory does not exist, prints an error and just ignores the writes.

src/executor/device.rs
```diff
@@ -132,3 +146,67 @@ impl<'a> NetworkInterface<'a> {
...
 	}
 }
+
+#[cfg(feature = "write-pcap-file")]
+/// Sink for packet captures. If the file Option is None, the writes are ignored.
+/// This is useful when we fail to create the sink file at runtime.
+pub(in crate::executor) struct FileSink(Option<File>);
+
+#[cfg(feature = "write-pcap-file")]
+impl FileSink {
```
mkroening: It would be preferable to put this in a separate module (can be inline). That way, we only have one outer `cfg`-attribute and can use module-global imports.

src/executor/device.rs
```diff
@@ -132,3 +146,67 @@ impl<'a> NetworkInterface<'a> {
...
+				match File::create_new(path.as_str()) {
+					Err(Errno::Exist) => {
+						file_name = format!("{file_name_base} ({i}).pcap");
+						ControlFlow::Continue(())
+					}
+					r => ControlFlow::Break(r),
+				}
+			})
+			.break_value()
+			.unwrap();
```
mkroening: I feel like a proper loop might be more readable.

src/executor/device.rs
```diff
@@ -132,3 +146,67 @@ impl<'a> NetworkInterface<'a> {
...
+		if let Err(e) = file {
+			if e == Errno::Noent {
+				error!("/root is not mounted. Are there any mount points for the VM?");
+			}
+			error!(
+				"Error {e:?} encountered while creating the pcap file. No pcap file will be written."
+			);
+		} else {
+			info!(
+				"The packet capture will be written to a file called \"{file_name}\" under the mount point."
```
mkroening: Why not print the full path? The mount point could be something other than `/root`.

src/executor/device.rs
```diff
@@ -132,3 +146,67 @@ impl<'a> NetworkInterface<'a> {
...
+		}
+	}
+
+	fn flush(&mut self) {
+		if let Some(file) = self.0.as_mut()
+			&& let Some(err) = file.flush().err()
+		{
+			error!("Error while flushing the pcap file: {err}");
+		}
+	}
```
mkroening: I think using `let else` with early returns might make this more readable than calling multiple methods in one `if` expression.

src/executor/network.rs
```diff
@@ -63,13 +67,25 @@ static LOCAL_ENDPOINT: AtomicU16 = AtomicU16::new(0);
...
+	feature = "net-trace" => {
+		smoltcp::phy::Tracer<NetworkDevice>
+	}
+	feature = "write-pcap-file" => {
+		smoltcp::phy::PcapWriter<NetworkDevice, crate::executor::device::FileSink>
+	}
+	_ => {
+		NetworkDevice
+	}
+};
```
mkroening: This might be a bit more composable:

```rust
type PcapDevice = cfg_select! {
	feature = "write-pcap-file" => {
		smoltcp::phy::PcapWriter<NetworkDevice, crate::executor::device::FileSink>
	}
	_ => NetworkDevice,
};

type TracerDevice = cfg_select! {
	feature = "net-trace" => {
		smoltcp::phy::Tracer<PcapDevice>
	}
	_ => PcapDevice,
};
```

src/executor/device.rs
cagatay-y: I don't think it is particularly useful to know where the file is placed in the guest file system tree. By mount point, I meant the one on the host, where the user will go to open the pcap file. What would be a clearer way to express that?

src/executor/network.rs
cagatay-y: What about `InnerDevice` and `InterfaceDevice`? Having types called `PcapDevice` and `TracerDevice` in a case where we don't actually have a pcap writer or a tracer seems a bit confusing to me.

src/executor/device.rs
mkroening: I think printing the full path from the point of the virtual machine makes sense. 

`-device vhost-user-fs-pci,tag=foo` would mount the share in `/foo`, for example. This is semantically very similar to mounting into Docker containers. Users should be aware that guest paths are different from host paths.

src/executor/network.rs
mkroening: I see. What about `MaybePcapDevice` and `MaybeTracerDevice`? That way it is still clear what these types are about, and this is composable in case more of these come in the future.

Cargo.toml
```diff
@@ -164,6 +164,12 @@ dns = ["net", "smoltcp", "smoltcp/socket-dns"]
 ## Enables pretty-printing the network traffic to the serial device.
 net-trace = ["smoltcp?/log", "smoltcp?/verbose"]
 
+## Enables generation of packet capture files. The files are placed under the
+## mount point on the host, with the name provided via the HERMIT_PCAP_NAME environment
+## variable, or the device name by default.
+write-pcap-file = []
+
+
```
mkroening: Nit: only use one newline.

.github/workflows/ci.yml
```diff
@@ -45,8 +45,8 @@ jobs:
         run: |
           cargo clippy --package hermit-kernel --target x86_64-unknown-none --features tcp,virtio-console,virtio-fs,virtio-net,virtio-vsock
           cargo clippy --package hermit-kernel --target x86_64-unknown-none --no-default-features --features tcp,virtio-console,virtio-fs,virtio-net,virtio-vsock
-          cargo hack clippy --package hermit-kernel --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net
-          cargo hack clippy --package hermit-kernel --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net --features pci
+          cargo hack clippy --package hermit-kernel --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net,write-pcap-file
+          cargo hack clippy --package hermit-kernel --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net,write-pcap-file --features pci
```
mkroening: Why exclude the feature?

src/executor/device.rs
```diff
@@ -13,6 +13,11 @@ use smoltcp::socket::dns;
 use smoltcp::wire::{EthernetAddress, HardwareAddress};
 #[cfg(not(feature = "dhcpv4"))]
 use smoltcp::wire::{IpCidr, Ipv4Address, Ipv4Cidr};
+#[cfg(feature = "write-pcap-file")]
+use {
+	crate::drivers::Driver,
+	smoltcp::phy::{PcapMode, PcapWriter},
+};
```
mkroening: Nit: Please split the imports.

src/executor/device.rs
```diff
@@ -132,3 +148,72 @@ impl<'a> NetworkInterface<'a> {
...
+#[cfg(feature = "write-pcap-file")]
+pub(in crate::executor) mod pcap_writer {
+	use embedded_io::Write;
+	use smoltcp::phy::PcapSink;
+
+	use crate::fs::File;
+
+	/// Sink for packet captures. If the file Option is None, the writes are ignored.
+	/// This is useful when we fail to create the sink file at runtime.
+	pub struct FileSink(Option<File>);
```
mkroening, line 163: Do you prefer this compared to panicking? An error log and continuing might be easier to miss. Both are fine with me, though.

src/executor/device.rs
```diff
@@ -132,3 +148,72 @@ impl<'a> NetworkInterface<'a> {
...
+
+	impl FileSink {
+		pub(super) fn new(default_name: &str) -> Self {
+			use crate::errno::Errno;
```
mkroening: Nit: Import this in the module instead.

.github/workflows/ci.yml
cagatay-y: I think it is a holdover from the initial version of the PR where the feature depended on the `net` feature. Removing it now.

src/executor/device.rs
cagatay-y: I also don't really have a strong preference.

src/executor/device.rs
cagatay-y: Oh, I was not aware that the guest mount point could be changed and I had misunderstood your initial comment. In that case, shouldn't we also not hardcode "/root" as the directory to write the pcap file, in addition to printing the full path?

mkroening: Yeah, that makes sense! 👍

src/executor/device.rs
mkroening: Let's keep it like it is then.

PR #2308, opened 2026-03-10
feat(net): TCP segmentation offload

### LINKED ISSUE #1861: Support segmentation offload
VIRTIO has support for drivers to offload the segmentation of large packets to MTU sized frames to the device. This can help reduce the descriptor table pressure by reducing the number of buffers that need to be provided and also relieves the CPU of some work. There is some prior work in #834 related to this. The feature requires support from the smol-tcp side.

## Notes
Some of the information below is not relevant for virtio-net but are here as they will be relevant for the changes that will be proposed for smoltcp.

- There are separate send side segmentation features for ~each combination in {IPv4, IPv6} × {TCP, UDP}~ TCP with IPv4, TCP with IPv6, UDP with either (at least for virtio-net).
  - on Linux [[7]]
  - on Windows [[8], [9]]
    Furthermore, UDP packets can be segmented (where they are sent as _unfragmented_ IPv4 packets each with their own UDP header) or fragmented (where they are sent as _fragmented_ IPv4 packets of a single UDP packet). The latter seems to be deprecated for the Linux kernel [[1]]. IP fragmentation seems to be undesirable in general [[17]].
  - Some devices may require the UDP packet to be segmented to be an exact multiple of the segment size [[18], [19]]
- For receive offloads, the structure is not symmetrical.
  - virtio separates UDP offload into one for IPv4 and one for IPv6 on the receive side [[11] (`VIRTIO_NET_F_GUEST_USO4` and `VIRTIO_NET_F_GUEST_USO6` in the v1.3 draft].
  - Linux does not seem to separate even TCP and UDP on the receive side [[12]]. There is LRO and GRO, and they seem similar [[15]].
  - Windows distinguishes between TCP and UDP coalescing [[13]] and IPv4/IPv6 for TCP [[14]]. 
- Different drivers may support different payload sizes to be segmented
  - virtio-net doesn't seem to specify the size for send but it is specified as 65536 (maximum TCP packet size) for receive offload [[2] (section 5.1.6.3)].
  - At least some Mellanox cards seem to support payloads four times as large [[3] (`max_tso` parameter)]. Mellanox Infiniband driver on the Linux kernel also references a `max_tso` parameter [[6]] and seems to be in the context of Ethernet networking.
  - One piece of Intel documentation specifies the maximum TCP payload size as 64,000 bytes [[4]] but it may be a case of 64 KiB accidentally turning into 64 KB. The same documentation specifies "64 _K_" (emphasis mine) as the limit for UDP segmentation [[5]].
  - Windows drivers support reporting the maximum offload size, which may be larger than 64K [[10]].

## smoltcp to-do (incomplete, in no particular order)
- [x] Communicate `gso_size` (the size of the segments to be created by the device) information to the device (probably through [`PacketMeta`](https://docs.rs/smoltcp/latest/smoltcp/phy/struct.PacketMeta.html))
- [x] Implement new Fragmenter that doesn't fragment IP packets
- [ ] Figure out if smoltcp needs to know about the large receive offload (a.k.a. receive segment coalescing)
    Probably, on Windows some out-of-band information needs to be provided [[16]].
- [ ] ~Deduplicate the segment size calculation in [`seq_to_transmit`](https://github.com/smoltcp-rs/smoltcp/blob/ac32e643a4b7e09161193071526b3ca5a0deedb5/src/socket/tcp.rs#L2219) and [`dispatch`](https://github.com/smoltcp-rs/smoltcp/blob/ac32e643a4b7e09161193071526b3ca5a0deedb5/src/socket/tcp.rs#L2492-L2545)~ (EDIT: seems to be unnecessary)

[1]: https://www.kernel.org/doc/html/latest/networking/segmentation-offloads.html#udp-fragmentation-offload
[2]: https://docs.oasis-open.org/virtio/virtio/v1.2/virtio-v1.2.html
[3]: https://enterprise-support.nvidia.com/s/article/raw-ethernet-programming--tso---code-example
[4]: https://edc.intel.com/content/www/us/en/design/products/ethernet/adapters-and-devices-user-guide/30.3/large-send-offload-ipv4-and-ipv6/
[5]: https://edc.intel.com/content/www/us/en/design/products/ethernet/adapters-and-devices-user-guide/30.3/udp-segmentation-offload-ipv4-and-ipv6/
[6]: https://elixir.bootlin.com/linux/v6.16.3/source/drivers/infiniband/hw/mlx5/main.c#L973
[7]: https://docs.kernel.org/networking/netdev-features.html
[8]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-segmentation-offload-uso-#:~:text=Support%20both%20IPv4%20and%20IPv6.
[9]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/offloading-the-segmentation-of-large-tcp-packets#:~:text=Support%20IPv4%20or%20IPv6%2C%20or%20both%20IPv4%20and%20IPv6.
[10]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/task-offload#:~:text=and%20segmentation%20for%20large%20TCP%20packets%20that%20are%20larger%20than%2064K.
[11]: https://docs.oasis-open.org/virtio/virtio/v1.3/virtio-v1.3.html
[12]: https://www.kernel.org/doc/html/latest/networking/segmentation-offloads.html#generic-receive-offload
[13]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-rsc-offload
[14]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/determining-the-rsc-capabilities-of-a-network-adapter
[15]: https://www.kernel.org/doc/html/latest/networking/netdev-features.html
[16]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/overview-of-receive-segment-coalescing#:~:text=Provide%20certain%20out%2Dof%2Dband%20information%20to%20the%20host%20TCP/IP%20stack.
[17]: https://www.rfc-editor.org/rfc/rfc8900
[18]: https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-segmentation-offload-uso-#:~:text=If%20the%20miniport%20driver%20doesn%27t%20set%20the%20SubMssFinalSegmentSupported%20capability%2C%20then%20each%20large%20UDP%20packet%20offloaded%20by%20the%20transport%20must%20have%20Length%20%25%20MSS%20%3D%3D%200%2E
[19]: https://docs.kernel.org/networking/netdev-features.html#:~:text=fixing%20up%20the%20last%20one%20if%20less%20than%20gso%5Fsize

cagatay-y:
Closes #1861.

Depends on:
- https://github.com/smoltcp-rs/smoltcp/pull/1148

### DIFF
--- a/src/drivers/net/virtio.rs
+++ b/src/drivers/net/virtio.rs
@@ -8,11 +8,12 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::mem::{ManuallyDrop, MaybeUninit};
+use core::num::NonZeroUsize;
 use core::str::FromStr;
 use core::{mem, slice};
 
 use smallvec::SmallVec;
-use smoltcp::phy::{Checksum, ChecksumCapabilities, DeviceCapabilities};
+use smoltcp::phy::{Checksum, ChecksumCapabilities, DeviceCapabilities, PacketMeta};
 use smoltcp::wire::{
 	ETHERNET_HEADER_LEN, EthernetFrame, IpAddress, IpProtocol, Ipv4Packet, Ipv6Packet, TcpPacket,
 	UdpPacket,
@@ -166,12 +167,28 @@ pub struct TxQueues {
 	buf_size: u32,
 }
 
+#[expect(
+	clippy::decimal_literal_representation,
+	reason = "In the VIRTIO specification the size limit is provided in decimal."
+)]
+// One possible implementation of TSO{4,6} (the transmit side feature) is the virtual network to directly copy the
+// transmitted buffer to the receive buffers of machines on the other side that implement the GUEST_TSO{4,6} feature.
+// Although VIRTIO spec. v1.4 sec. 5.1.9.3 specifies 65589 bytes to be the maximum length that may be received with
+// GUEST_TSO{4,6}, the limit in v1.2 sec. 5.1.6.3 was 65550. Because the new, higher limit may cause incompatibility
+// with receivers that use the old value, we use the lower, safer limit when sending.
+const MAX_TSO_BUFFER_SIZE: usize = 65550;
+
 impl TxQueues {
 	fn new(vqs: Vec<VirtQueue>, dev_cfg: &NetDevCfg) -> Self {
-		Self {
-			vqs,
-			buf_size: determine_mtu(dev_cfg).into(),
-		}
+		let buf_size = if dev_cfg
+			.features
+			.intersects(virtio::net::F::HOST_TSO4 | virtio::net::F::HOST_TSO6)
+		{
+			MAX_TSO_BUFFER_SIZE.try_into().unwrap()
+		} else {
+			determine_mtu(dev_cfg).into()
+		};
+		Self { vqs, buf_size }
 	}
 
 	#[allow(dead_code)]
@@ -233,6 +250,7 @@ pub struct TxToken<'a> {
 	send_vqs: &'a mut TxQueues,
 	checksums: ChecksumCapabilities,
 	send_capacity: &'a mut u32,
+	meta: PacketMeta,
 }
 
 impl Drop for TxToken<'_> {
@@ -260,15 +278,32 @@ impl smoltcp::phy::TxToken for TxToken<'_> {
 
 		let mut header = Box::new_in(<Hdr as Default>::default(), DeviceAlloc);
 
+		let is_large_send = cfg_select! {
+			feature = "tcp" => token.meta.segmentation_offload_size.is_some(),
+			_ => false,
+		};
+
 		// If a checksum calculation by the host is necessary, we have to inform the host within the header
 		// see Virtio specification 5.1.6.2
 		if let Some((ip_header_len, csum_offset)) =
-			VirtioNetDriver::should_request_checksum(&token.checksums, &mut packet)
+			VirtioNetDriver::should_request_checksum(&token.checksums, &mut packet, is_large_send)
 		{
 			header.flags = HdrF::NEEDS_CSUM;
 			header.csum_start =
 				(u16::try_from(ETHERNET_HEADER_LEN).unwrap() + ip_header_len).into();
 			header.csum_offset = csum_offset.into();
+		}
+
+		#[cfg(feature = "tcp")]
+		if let Some(gso_size) = token.meta.segmentation_offload_size {
+			header.gso_type = match EthernetFrame::new_unchecked(&packet).ethertype() {
+				smoltcp::wire::EthernetProtocol::Ipv4 => virtio::net::HdrGso::Tcpv4.into(),
+				smoltcp::wire::EthernetProtocol::Ipv6 => virtio::net::HdrGso::Tcpv6.into(),
+				_ => unreachable!(
+					"We don't advertise segmentation offload support for any other protocol."
+				),
+			};
+			header.gso_size = gso_size.get().into();
 		}
 
 		let buff_tkn = AvailBufferToken::new(
@@ -282,6 +317,10 @@ impl smoltcp::phy::TxToken for TxToken<'_> {
 			.unwrap();
 
 		result
+	}
+
+	fn set_meta(&mut self, meta: PacketMeta) {
+		self.meta = meta;
 	}
 }
 
@@ -548,6 +587,22 @@ impl smoltcp::phy::Device for VirtioNetDriver {
 		device_capabilities.max_burst_size =
 			Some(usize::try_from(self.send_capacity).unwrap() / usize::from(BUFF_PER_PACKET));
 		device_capabilities.checksum = self.checksums.clone();
+
+		// We only support segmentation offload on and enable the corresponding feature of smoltcp for TCP.
+		#[cfg(feature = "tcp")]
+		{
+			device_capabilities.segmentation.tcpv4 = self
+				.dev_cfg
+				.features
+				.contains(virtio::net::F::HOST_TSO4)
+				.then_some(NonZeroUsize::new(MAX_TSO_BUFFER_SIZE).unwrap());
+			device_capabilities.segmentation.tcpv6 = self
+				.dev_cfg
+				.features
+				.contains(virtio::net::F::HOST_TSO6)
+				.then_some(NonZeroUsize::new(MAX_TSO_BUFFER_SIZE).unwrap());
+		}
+
 		device_capabilities
 	}
 
@@ -573,6 +628,7 @@ impl smoltcp::phy::Device for VirtioNetDriver {
 				send_vqs: &mut self.send_vqs,
 				checksums: self.checksums.clone(),
 				send_capacity: &mut self.send_capacity,
+				meta: PacketMeta::default(),
 			},
 		))
 	}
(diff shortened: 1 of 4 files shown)


mkroening (approved):
Thanks a lot! :)

This looks good. Let's wait a bit if there is any movement on the upstream PR.

cagatay-y:
Apart from the rebase, since the last review I renamed `MAX_BUFFER_SIZE` to `MAX_TSO_BUFFER_SIZE` and added a comment on its value.

cagatay-y:
Pseudo-checksum calculation was adapted for #2607, so it would be nice to retest with vDPA to make sure it is still working if you have the chance.

src/drivers/net/virtio.rs
```diff
@@ -166,12 +167,28 @@ pub struct TxQueues {
...
+#[expect(
+	clippy::decimal_literal_representation,
+	reason = "In the VIRTIO specification the size limit is provided in decimal."
+)]
+// One possible implementation of TSO{4,6} (the transmit side feature) is the virtual network to directly copy the
+// transmitted buffer to the receive buffers of machines on the other side that implement the GUEST_TSO{4,6} feature.
+// Although VIRTIO spec. v1.4 sec. 5.1.9.3 specifies 65589 bytes to be the maximum length that may be received with
+// GUEST_TSO{4,6}, the limit in v1.2 sec. 5.1.6.3 was 65550. Because the new, higher limit may cause incompatibility
+// with receivers that use the old value, we use the lower, safer limit when sending.
+const MAX_TSO_BUFFER_SIZE: usize = 65550;
```
mkroening, line 179: I understand the value a bit differently. This is the maximum incoming packet length, so we can expect there to be no larget incoming packet. But there is no harm in giving a larger buffer, right? So I'd go for the new v1.4 value.

BlueField-3 testing works fine with the higher value too.

mkroening (changes requested):
Thanks, great! 🥳 

These are my BlueField-3 numbers:

`main` (a229a560c9d99312384293eb3bb7230ca28c62ae):

```
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  76.4 MBytes   640 Mbits/sec                  
[  5]   1.00-2.00   sec  76.1 MBytes   639 Mbits/sec                  
[  5]   2.00-3.00   sec  76.4 MBytes   641 Mbits/sec                  
[  5]   3.00-4.00   sec  76.2 MBytes   640 Mbits/sec                  
[  5]   4.00-5.00   sec  76.2 MBytes   640 Mbits/sec                  
[  5]   5.00-6.00   sec  76.2 MBytes   640 Mbits/sec                  
[  5]   6.00-7.00   sec  76.2 MBytes   640 Mbits/sec                  
[  5]   7.00-8.00   sec  76.8 MBytes   644 Mbits/sec                  
[  5]   8.00-9.00   sec  76.6 MBytes   643 Mbits/sec                  
[  5]   9.00-10.00  sec  76.5 MBytes   641 Mbits/sec                  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-10.04  sec   764 MBytes   638 Mbits/sec                  sender
[  5]   0.00-10.00  sec   764 MBytes   641 Mbits/sec                  receiver
```

`main` + this PR:

```
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec   331 MBytes  2.77 Gbits/sec                  
[  5]   1.00-2.00   sec   330 MBytes  2.77 Gbits/sec                  
[  5]   2.00-3.00   sec   327 MBytes  2.74 Gbits/sec                  
[  5]   3.00-4.00   sec   327 MBytes  2.75 Gbits/sec                  
[  5]   4.00-5.00   sec   325 MBytes  2.73 Gbits/sec                  
[  5]   5.00-6.00   sec   324 MBytes  2.71 Gbits/sec                  
[  5]   6.00-7.00   sec   325 MBytes  2.73 Gbits/sec                  
[  5]   7.00-8.00   sec   325 MBytes  2.72 Gbits/sec                  
[  5]   8.00-9.00   sec   324 MBytes  2.72 Gbits/sec                  
[  5]   9.00-10.00  sec   323 MBytes  2.71 Gbits/sec                  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-10.04  sec  3.18 GBytes  2.72 Gbits/sec                  sender
[  5]   0.00-10.00  sec  3.18 GBytes  2.74 Gbits/sec                  receiver
```

Fedora 40 VM:

```
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  3.94 GBytes  33.8 Gbits/sec                  
[  5]   1.00-2.00   sec  2.75 GBytes  23.6 Gbits/sec                  
[  5]   2.00-3.00   sec  2.72 GBytes  23.4 Gbits/sec                  
[  5]   3.00-4.00   sec  3.08 GBytes  26.5 Gbits/sec                  
[  5]   4.00-5.00   sec  3.09 GBytes  26.6 Gbits/sec                  
[  5]   5.00-6.00   sec  2.73 GBytes  23.5 Gbits/sec                  
[  5]   6.00-7.00   sec  2.74 GBytes  23.5 Gbits/sec                  
[  5]   7.00-8.00   sec  2.71 GBytes  23.3 Gbits/sec                  
[  5]   8.00-9.00   sec  2.78 GBytes  23.9 Gbits/sec                  
[  5]   9.00-10.00  sec  2.72 GBytes  23.4 Gbits/sec                  
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  29.3 GBytes  25.2 Gbits/sec  1445            sender
[  5]   0.00-10.00  sec  29.3 GBytes  25.2 Gbits/sec                  receiver
```

src/drivers/net/virtio.rs
cagatay-y: `MAX_TSO_BUFFER_SIZE` is used as the maximum _outgoing_ packet length. My thinking was that the receiver of the packet may have the old value as the maximum incoming length, so have insufficient buffer size. For what we receive, the size was updated to the larger value according to v1.4 with the commit 4a8c62c89dc0812647b86aaf3bb802f1850d61d7.

mkroening: Ah, yeah. Not sure why I got confused when I already understood it once. ^^

I would personally still go with the larger value. I think the patch (https://github.com/oasis-tcs/virtio-spec/commit/00806815385340dd411cc67df3f6837935bb5e26) fixes a calculation bug. I don't think the number is normative, but just a helpful calculation. I don't think people have relied on this value; else they would have been more careful with changing the value in the spec and would have given migration guidance as they do with other changes.

If you still think we should be conservative here, that's also fine with me. My gut feeling would be to go with the latest number, though. :)

mkroening: What do you think? 🤔

cagatay-y: Sorry, I could not take a look because of an exam.

Your point about the way the value in the specification was updated makes sense to me, but for what it's worth the larger value affords us just 39 bytes for a relative increase of less than 6‰. It is up to you :)

mkroening: All right. I'll merge this then, and increase the value in a follow-up PR. :)

mkroening (approved):
Thank you so much for seeing this through! :)

PR #2320, opened 2026-03-12
fix(interrupts): only handle PIC interrupts

jounathaen:


src/arch/x86_64/kernel/interrupts.rs
```diff
@@ -21,6 +21,7 @@ use crate::drivers::InterruptHandlerQueue;
 use crate::drivers::mmio::get_interrupt_handlers;
 #[cfg(feature = "pci")]
 use crate::drivers::pci::get_interrupt_handlers;
+use crate::kernel::pic::PIC1_INTERRUPT_OFFSET;
```
jounathaen: Probably `crate::arch::x86_64::kernel::pic::PIC1_INTERRUPT_OFFSET` would be more consistent here.

PR #2342, opened 2026-03-19
feat(logging): add extended log filtering

cagatay-y:
~The new environment variable `HERMIT_LOG_EXTENDED_FILTER`~ allows log filters of different levels to be applied on a per module basis. ~It also allows filtering by regex.~

The new filter depends on the `env_filter` crate that was originally developed for the `env_logger` crate. Given it is from the  Rust CLI WG and the crate is not particularly big I think it is a reasonable inclusion. Alternatively, we can consider parsing the filter ourselves. ~The PR currently enables the regex filter from the crate. Since it pulls in a larger crate with it, we may consider disabling that.~ Lastly, we can also consider making the whole thing an optional feature. I wanted to go with the simplest proof of concept for now.

Some changes were needed to the `env_filter` crate for it to work with `no_std`. I think they are non-intrusive enough to be acceptable for the upstream. EDIT: They got accepted.

~The PR also includes changes to things that I noticed while reading the module.~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -603,6 +603,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_filter"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "900d271a03799a1ee8d1ca9b19893b48ca674a9284fefcfb85f05e74ed314217"
+dependencies = [
+ "log",
+]
+
+[[package]]
 name = "errno"
 version = "0.3.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -849,6 +858,7 @@ dependencies = [
  "embedded-io",
  "endian-num",
  "enum_dispatch",
+ "env_filter",
  "fdt",
  "float-cmp",
  "free-list",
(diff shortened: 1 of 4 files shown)


mkroening (commented):
Thanks for working on this. I've been wanting to use `env_logger` for a long time. It's great that you chose to tackle this! :)

I did not have a look at the code yet but wanted to propose a different set of environment variables:
- The run-time environment variable `HERMIT_LOG` controls the logging. This works exactly like `env_logger`'s `RUST_LOG`.
- The build-time environment variable `HERMIT_LOG_DEFAULT` controls the default value of `HERMIT_LOG` if it is not set at runtime. `HERMIT_LOG_LEVEL_FILTER` is a deprecated alias to this. This should be backwards compatible if I am not mistaken.
- The build-time environment variable `HERMIT_LOG_STATIC_MAX_LEVEL` sets the corresponding `log` crate features via xtask to properly avoid any runtime overhead of statically excluded log levels.

This is a long-term plan, though and should be worked towards in small PRs. What are you thoughts on this proposal?

mkroening (approved):
Thanks! :)

PR #2366, opened 2026-04-08
feat(x86_64): add PVH boot support

mkroening:
This PR adds PVH boot support directly to the kernel. This allows the kernel to be booted in QEMU, Firecracker, and Cloud Hypervisor directly, without the loader.

The [`pvh`](https://github.com/rust-osdev/pvh) crate was successfully published.

Depends on:
- https://github.com/hermit-os/kernel/pull/2546
- https://github.com/hermit-os/kernel/pull/2645
- https://github.com/hermit-os/kernel/pull/2646

Closes https://github.com/hermit-os/kernel/issues/2373.


src/arch/x86_64/kernel/pvh/pvh_start32.s
```diff
@@ -0,0 +1,64 @@
...
+
+    # Set CR0 (PM-bit is already set)
+    mov eax, cr0
+    and eax, ~(1 << 2)      # disable FPU emulation
+    or eax, (1 << 1)        # enable FPU montitoring
+    and eax, ~(1 << 30)     # enable caching
+    and eax, ~(1 << 29)     # disable write through caching
+    and eax, ~(1 << 16)	    # allow kernel write access to read-only pages
+    or eax, (1 << 31)       # enable paging
+    mov cr0, eax
```
TSnake41, line 36: PVH ABI (https://xenbits.xen.org/docs/unstable/misc/pvh.html) define CR0 state as 
> cr0: bit 0 (PE) must be set. All the other writeable bits are cleared.

Hence you don't need to to clear some of these bits.

TSnake41, line 36: Looks like CR0 is set up twice.

TSnake41 (changes requested):

src/arch/x86_64/kernel/pvh/pvh_start32.s
mkroening: Thanks for pointing this out! :)

This is still very much work-in-progress. I will clean it up. 👍

mkroening: This is now resolved.

mkroening: This is now resolved.

PR #2381, opened 2026-04-20
fix: redundant_imports and unused_qualifications

mkroening:


src/arch/aarch64/kernel/interrupts.rs
```diff
@@ -1,8 +1,8 @@
 use alloc::collections::{BTreeMap, VecDeque};
 use core::arch::asm;
+use core::ptr;
```
fogti: It might be a good idea to merge this with the next line to:
```suggestion
use core::ptr::{self, NonNull};
```

src/drivers/virtio/transport/pci.rs
```diff
@@ -6,8 +6,8 @@
 //! [Virtio Over PCI Bus]: https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-1150001
 
 use alloc::vec::Vec;
+use core::ptr;
```
fogti:
```suggestion
use core::ptr::{self, NonNull};
```

fogti (approved):

PR #2389, opened 2026-04-21
refactor(fs): Prepare root filesystem in big leap before applying it globally

fogti:
This is in preparation of implementing Hermit image later.

This is split from #2077 because I'm going to rewrite the rest of it later anyways, and want to integrate the already "fixed" parts beforehand.


mkroening (approved):
Thanks! :)

PR #2390, opened 2026-04-21
feat(fs/mem): Tarfs for Hermit images implementation

fogti:
Supersedes #2077.

Depends on #2540.


fogti:
@mkroening Would it be possible to merge just part of this (namely the changes to `src/fs/mem.rs`, but not to other files in `src`)? That part is basically done. (And the rest suffers from memory management issues which I don't really know yet how to solve.) It would be nice to have a test for that, could I just add an example uncompressed `tar` archive to the repository for that and basically do something that amount essentially to snapshot testing (tho manual)?

mkroening:
> Currently untested. Supersedes #2077.

Is this tested now?

> Would it be possible to merge just part of this (namely the changes to src/fs/mem.rs, but not to other files in src)? That part is basically done. (And the rest suffers from memory management issues which I don't really know yet how to solve.)

Is this still up to date?

> It would be nice to have a test for that, could I just add an example uncompressed tar archive to the repository for that and basically do something that amount essentially to snapshot testing (tho manual)?

I would like to avoid binary files in the repo. Generating a file for the test is fine, though.

Cargo.toml
```diff
@@ -326,6 +326,7 @@ simple-shell = { version = "0.0.1", optional = true }
 smallvec = { version = "1", features = ["const_new"] }
 take-static = "0.1"
 talc = { version = "5" }
+tar-no-std = { version = "0.4", features = ["alloc"] }
```
mkroening: Please feature-gate this.

src/mm/mod.rs
```diff
@@ -100,13 +146,24 @@ pub(crate) fn init() {
 		arch::mm::init();
 	}
 
+	Lazy::force(&HERMIT_IMAGE_START_AND_LEN);
+
 	let total_mem = physicalmem::total_memory_size();
 	let kernel_addr_range = KERNEL_ADDR_RANGE.clone();
+	let protect_addr_max = if let Some(hisnl) = *HERMIT_IMAGE_START_AND_LEN {
```
mkroening: Can't you destructure here and avoid the cryptic var name?


```suggestion
	let protect_addr_max = if let Some((image_start, image_len)) = *HERMIT_IMAGE_START_AND_LEN {
```

src/fs/mem.rs
```diff
@@ -464,6 +459,63 @@ impl MemDirectory {
...
 
+	pub fn try_from_image(image: &'static [u8]) -> io::Result<Self> {
+		let this = Self::new(AccessPermission::S_IRUSR);
+
+		let taref = tar_no_std::TarArchiveRef::new(image).map_err(|e| {
+			error!("[Hermit image] Tar file has invalid format: {e:?}");
+			Errno::Inval
+		})?;
+
+		for i in taref.entries() {
```
mkroening: `i` is not an index.


```suggestion
		for entry in taref.entries() {
```

src/fs/mem.rs
```diff
@@ -464,6 +459,63 @@ impl MemDirectory {
 		}
 	}
 
+	pub fn try_from_image(image: &'static [u8]) -> io::Result<Self> {
+		let this = Self::new(AccessPermission::S_IRUSR);
+
+		let taref = tar_no_std::TarArchiveRef::new(image).map_err(|e| {
```
mkroening: What does `taref` mean? Is it a short form of `tar_ref`?

src/fs/mod.rs
```diff
@@ -304,7 +304,13 @@ pub(crate) fn init() {
 	const VERSION: &str = env!("CARGO_PKG_VERSION");
 	const UTC_BUILT_TIME: &str = build_time::build_time_utc!();
 
-	let root_filesystem = Filesystem::new();
+	let mut root_filesystem = Filesystem::new();
+
+	// Handle optional Hermit Image specified in FDT.
+	if let Some(tar_image) = crate::mm::hermit_tar_image() {
+		root_filesystem.root =
+			MemDirectory::try_from_image(tar_image).expect("Unable to parse Hermit Image");
```
mkroening: Should we not print an error instead of panicking and continue?

src/mm/mod.rs
```diff
@@ -90,6 +90,52 @@ pub(crate) fn kernel_end_address() -> VirtAddr {
...
+/// (indicated via FDT).
+static HERMIT_IMAGE_START_AND_LEN: Lazy<Option<(VirtAddr, usize)>> = Lazy::new(|| {
+	let fdt = env::fdt()?;
+
+	// per FDT specification, /chosen always exists
+	let chosen = fdt.find_node("/chosen").unwrap();
+
+	let fdt::node::NodeProperty {
+		value: image_reg, ..
+	} = chosen.property("image_reg")?;
```
mkroening: Is there a spec on the design of this FDT node?

Why is it so much more complicated to parse than a Linux initrd (https://github.com/hermit-os/loader/blob/v0.5.6/src/arch/riscv64/mod.rs#L21-L31)?

Should we model it like a Linux initrd instead? We could do either straight up `linux,initrd-start` and `linux,initrd-end` as that is what we use with other VMMs or adapt Hermit-specific names just for Uhyve (`hermit,...` or `uhyve,...`).

src/mm/mod.rs
```diff
@@ -90,6 +90,52 @@ pub(crate) fn kernel_end_address() -> VirtAddr {
...
+			len.len(),
+		);
+		None
+	}
+});
+
+pub(crate) fn hermit_tar_image() -> Option<&'static [u8]> {
+	// technically, the following is UB, because the kernel might be contained within...
+	HERMIT_IMAGE_START_AND_LEN
+		.map(|(addr, len)| unsafe { core::slice::from_raw_parts(addr.as_ptr(), len) })
```
mkroening: Why does the FDT node include the kernel? Why not only put the initrd into the node?

src/mm/mod.rs
```diff
@@ -90,6 +90,52 @@ pub(crate) fn kernel_end_address() -> VirtAddr {
 	KERNEL_ADDR_RANGE.end
 }
 
+/// Physical and virtual address range of the Hermit image, in case it is present
+/// (indicated via FDT).
+static HERMIT_IMAGE_START_AND_LEN: Lazy<Option<(VirtAddr, usize)>> = Lazy::new(|| {
```
mkroening: Is having a lazy static for this worth it? Or would a function suffice?

src/mm/mod.rs
mkroening: Since this is not about protecting the memory through paging or similar, a different name might be more fitting. For example, `reserved_addr_range`. Reserved is also the term used by ACPI and thus UEFI and PVH.

fogti:
I could repeat a test similar to the one that is now in `hermit-entry`, is that okay?

mkroening (commented):
Thanks for the PR! :)

It would be great if you could add a high-level description of the changes to the PR description.

mkroening:
> I could repeat a test similar to the one that is now in `hermit-entry`, is that okay?

Yeah, sounds good! 👍

src/fs/mod.rs
fogti: It doesn't really make sense to boot a kernel that expects a very specific file system environment, and letting the kernel later run into errors when the image loading fails.

src/mm/mod.rs
fogti: I modeled it directly after the usually used `reg` property, afaik. See also https://docs.rs/fdt/latest/fdt/node/struct.FdtNode.html#method.reg ; https://github.com/devicetree-org/devicetree-specification/releases/download/v0.4/devicetree-specification-v0.4.pdf "2.3.6. `reg`"

src/mm/mod.rs
fogti: The kernel doesn't know if it is contained in it or not, it is simply possible that the ranges overlap (at least currently we don't outlaw that because it might provide memory usage savings in case of using the `hermit-loader`; `uhyve` always copies the kernel anyways).

src/mm/mod.rs
fogti: I would like to avoid parsing this information multiple times;

Normally, we need it twice: once during the memory setup (to ensure that memory is reserved for it, even tho the FDT reserved memory regions should take care of that), and once during filesystem setup.

src/fs/mem.rs
fogti: yes; I'll fix it.

src/mm/mod.rs
mkroening: I see. Doing it the way I described above would make sense, I think.

src/mm/mod.rs
mkroening: I don't understand. How can they ever overlap? Sure, the image could contain the kernel ELF disk image, but not the loaded image, right?

src/mm/mod.rs
mkroening: Can you measure any benefit here? I expect the parsing overhead to be minimal.

src/mm/mod.rs
fogti: For `uhyve`, the case is clear: https://github.com/fogti/uhyve/blob/278bfd24a8744adc762fd598bcd025495bcc1d9b/src/vm.rs#L365
For `loader`, I'm not sure. There is no final implementation yet, and idk how the loading (dyn-linking?) of the kernel works there.

src/mm/mod.rs
fogti: Probably; but I prefer an implementation that makes it harder to find out when stuff can panic and when not.

And in general, it might be possible to "thread" this information through the call-stack via argument passing, but `mm/mod.rs` and `fs/mod.rs` are quite far apart from each other in terms of invocation sites.

src/mm/mod.rs
mkroening: The loader works very similarly to Uhyve. It's just happening from inside the VM.

We get an ELF image and load the ELF image to memory by copying. The running executable is not running straight from the ELF image.

fogti: okay, then normally, no overlap should be possible (except by loader error).

src/mm/mod.rs
fogti: okay, hermit-specific naming would make sense.

fogti:
Okay, when I write a test, how should I embed that into the tree, given that the `tar` crate unfortunately requires `libstd` as a dependency (at least for now, until `std::io::{Error,ErrorKind}` are stable'd into `libcore`/`liballoc`?

fogti:
Okay, Hermit Image support is now contained in a proper release of `uhyve` (v0.9.1).

Test case: https://github.com/hermit-os/uhyve/pull/1135#issuecomment-3490790877

The test probably has to go into `hermit-rs`, but it is not uniquely obvious how `hermit-kernel/xtask` should be modified to support packing and invoking hermit images.
Proposal:
* Packing support is necessary because `xtask ci rs` does the build and the run parts, and the packing of an image has to be done in-between. There doesn't exist any specification format yet how to define what to pack into an image, tho there exists a canonical interpretation that takes a directory tree as the image root contents, and append a kernel into it at the location indicated by the hermit image config file `hermit.toml`. Thus the inputs of `--package $PACKAGE` and to-be-introduced `--image-root $IMAGE_ROOT` would suffice to create a compressed tarball from it.
* Invoking: afaik this makes the addition/forwarding of a CLI parameter `--hermit-image-mode (internal|external)`  necessary.

@mkroening would this approach be okay?

mkroening:
What about just putting a `root` directory next to the `Cargo.toml` for the hermit-rs test and then just building that when running that specific test application? That's also how the applications are currently tested (based on their name).

I'd like to avoid feature creep in xtask if possible. It is meant to be a script that helps with development, similar to Rust's `x`. It is not a goal to be a maximally general standalone Hermit application runner. If it is necessary, though, we can also add flags.

> * Invoking: afaik this makes the addition/forwarding of a CLI parameter `--hermit-image-mode (internal|external)`  necessary.

What would this do?

fogti:
> What would this do?

It switches between the Hermit image implementation that is external (i.e. implemented completely in Uhyve, and the kernel needs no adjustments to work with that) and the one that is internal (i.e. Uhyve only does the decompression, but the rest gets done in the Kernel, i.e. this PR, which should be theoretically faster because we can skip the hypercall overhead).

mkroening:
Ah, that makes sense. So that just results in slightly different Uhyve flags, right? I'll try adding support for passing arbitrary VMM arguments through xtask for https://github.com/hermit-os/kernel/pull/2416 anyway. That should be enough, no?

fogti:
@mkroening  How should I account for 7eb707f0c220e619c8f200a5a78e0df4dfe41ac0 ?

fogti:
Should I put the `xtask` implementation into a separate PR, given that `uhyve --hermit-image-mode external` already works, but `uhyve --hermit-image-mode internal` doesn't (classic `FrameAlloc` error)?

fogti:
I added some basic documentation for this in `lib.rs` and `xtask/README.md`.

fogti:
I'll rebase this branch once the other PRs this is blocking on are dealt with: #2624, #2633.

fogti:
How should I deal with the following?
```
 Documenting hermit-kernel v0.13.0 (/home/runner/work/kernel/kernel)
error: unresolved link to `hermit_entry::config::Config`
  --> src/lib.rs:35:35
   |
35 | //! conforming to the format of [`hermit_entry::config::Config`].
   |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `config` in module `hermit_entry`
   |
   = note: `-D rustdoc::broken-intra-doc-links` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(rustdoc::broken_intra_doc_links)]`

error: could not document `hermit-kernel`
```

mkroening:
> How should I deal with the following?

Oh no, what a pity. We cannot just build docs with dependencies because hashbrown is not `no_std` on `cfg(doc)`: https://github.com/rust-lang/hashbrown/blob/v0.17.1/src/lib.rs#L12

How about just linking to crates.io? https://docs.rs/hermit-entry/latest/hermit_entry/config/enum.Config.html

fogti:
I applied the suggested doc fix.

fogti:
How do I embed the test for this in CI?

fogti:
Oh right, CI is using the broken Uhyve version v0.9.1

@jounathaen Could I get another Uhyve release (which would then be `v0.10.0` due to breaking API changes https://github.com/hermit-os/uhyve/compare/v0.9.1...main) which has to include at least https://github.com/hermit-os/uhyve/pull/1447, but preferably just all the bugfix PRs and such that were recently merged.

mkroening:
You could install the git Uhyve in CI to ensure nothing else is missing. :)

fogti:
Looks good.

PR #2391, opened 2026-04-22
feat: add `uhyve` feature

mkroening:
Currently, Hermit unikernels can only be specialized to which device drivers to include, but not to which VMM the unikernel is run in.

This PR enables such specialization by feature-gating the Uhyve code. This means when Uhyve is targeted, the `uhyve` feature must be enabled. Currently, this only reduces code size and dependencies for unikernels that do not target Uhyve. In a follow-up PR, I will add the `loader` feature to enable savings for unikernels that do target Uhyve.

In the grander scheme of things, going this direction allows us to nicely specialize for direct PVH support in the future (https://github.com/hermit-os/kernel/pull/2366).

Depends on https://github.com/hermit-os/hermit-rs/pull/976.
I will pull out some of the preparatory refactorings into a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,9 +235,9 @@ jobs:
         if: matrix.arch == 'x86_64'
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} ${{ matrix.rs_flags }} --package rusty_demo --no-default-features --smp 4 firecracker --sudo
         if: matrix.arch == 'x86_64'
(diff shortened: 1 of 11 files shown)


jounathaen:
Can we make `uhyve` a default feature?

mkroening:
> Can we make `uhyve` a default feature?

Yes, I made it so. As of https://github.com/hermit-os/hermit-rs/pull/990, it is possible to do this atomically even without the previously necessary dance.

jounathaen (approved):

PR #2405, opened 2026-05-04
feat(aarch64/processor): print pretty CPU name

Gelbpunkt:
The MIDR_EL1 register contains an implementer ID and a part ID which we can map to a pretty name for the CPU core. We can do so by taking the lookup tables from util-linux, which has a well-maintained list of part IDs and their names.

Ideally this would be called *after* all secondary CPUs were initialized, but it currently runs before that is the case. Since heterogenous layouts are common on ARM, it would be nice to collect this information on all cores and print an aggregate, but that's for follow-up work.

I'm unsure whether we want to keep the lookup tables in the kernel itself or just move them to a separate crate, I think this is valuable information not just for Hermit, but also other kernels.

<hr>

What this looks like in KVM on my laptop:

```
[    0.000170][0][INFO  processor ] ========================== CPU INFORMATION ===========================
[    0.000589][0][INFO  processor ] Processor:               Qualcomm Oryon
[    0.000892][0][INFO  processor ] Counter frequency:       19200 KHz (from CNTFRQ_EL0)
[    0.001245][0][INFO  processor ] ======================================================================
```

and in QEMU:

```
[    0.000289][0][INFO  processor ] ========================== CPU INFORMATION ===========================
[    0.000432][0][INFO  processor ] Processor:               ARM Cortex-A72
[    0.000527][0][INFO  processor ] Counter frequency:       62500 KHz (from CNTFRQ_EL0)
[    0.000680][0][INFO  processor ] ======================================================================
```

This gets us a bit closer to the pretty printed CPU information on x86_64.

### DIFF
--- a/src/arch/aarch64/kernel/lscpu.rs
+++ b/src/arch/aarch64/kernel/lscpu.rs
@@ -1,0 +1,291 @@
+//! Helpers for mapping the implementer and part number fields in the MIDR_EL1
+//! register to a pretty, human-readable CPU core name.
+//!
+//! The lookup tables here are taken from util-linux:
+//! <https://github.com/util-linux/util-linux/blob/da322604a45f8094d439ea64be7d44cfea1f500a/sys-utils/lscpu-arm.c>
+//!
+//! util-linux is licensed under the GPL, but the implementer/model name tables
+//! are uncopyrightable and explicitly intended to be reused in other projects,
+//! see for example this commit:
+//! <https://github.com/util-linux/util-linux/commit/16724407c858f4d3e2d2497f758b678b644c0011>
+
+type PartsTable = &'static [(u16, &'static str)];
+
+static ARM_PARTS: PartsTable = &[
+	(0x810, "ARM810"),
+	(0x920, "ARM920"),
+	(0x922, "ARM922"),
+	(0x926, "ARM926"),
+	(0x940, "ARM940"),
+	(0x946, "ARM946"),
+	(0x966, "ARM966"),
+	(0xa20, "ARM1020"),
+	(0xa22, "ARM1022"),
+	(0xa26, "ARM1026"),
+	(0xb02, "ARM11-MPCore"),
+	(0xb36, "ARM1136"),
+	(0xb56, "ARM1156"),
+	(0xb76, "ARM1176"),
+	(0xc05, "Cortex-A5"),
+	(0xc07, "Cortex-A7"),
+	(0xc08, "Cortex-A8"),
+	(0xc09, "Cortex-A9"),
+	(0xc0d, "Cortex-A17"),
+	(0xc0f, "Cortex-A15"),
+	(0xc0e, "Cortex-A17"),
+	(0xc14, "Cortex-R4"),
+	(0xc15, "Cortex-R5"),
+	(0xc17, "Cortex-R7"),
+	(0xc18, "Cortex-R8"),
+	(0xc20, "Cortex-M0"),
+	(0xc21, "Cortex-M1"),
+	(0xc23, "Cortex-M3"),
+	(0xc24, "Cortex-M4"),
+	(0xc27, "Cortex-M7"),
+	(0xc60, "Cortex-M0+"),
+	(0xd01, "Cortex-A32"),
+	(0xd02, "Cortex-A34"),
+	(0xd03, "Cortex-A53"),
+	(0xd04, "Cortex-A35"),
+	(0xd05, "Cortex-A55"),
+	(0xd06, "Cortex-A65"),
+	(0xd07, "Cortex-A57"),
+	(0xd08, "Cortex-A72"),
+	(0xd09, "Cortex-A73"),
+	(0xd0a, "Cortex-A75"),
+	(0xd0b, "Cortex-A76"),
+	(0xd0c, "Neoverse-N1"),
+	(0xd0d, "Cortex-A77"),
+	(0xd0e, "Cortex-A76AE"),
+	(0xd13, "Cortex-R52"),
+	(0xd14, "Cortex-R82AE"),
+	(0xd15, "Cortex-R82"),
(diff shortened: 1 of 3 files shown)


mkroening (changes requested):
Thanks! Can you put this into a separate module? Something like `lscpu_arm`, like the original. Also, please add a clarifying module comment about licenses: The original code is GPL licensed, but we only use the data, not the code, which, e.g., Mozilla also does: https://github.com/mozilla-firefox/firefox/blob/FIREFOX_150_0_1_RELEASE/xpcom/base/nsSystemInfo.cpp#L850-L851

In the long term, it might make sense to factor this out into a separate crate, even, but it's fine to not do that for now.

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,284 @@
...
+//!
+//! The lookup tables here are taken from util-linux:
+//! <https://github.com/util-linux/util-linux/blob/da322604a45f8094d439ea64be7d44cfea1f500a/sys-utils/lscpu-arm.c>
+//!
+//! util-linux is licensed under the GPL, but the implementer/model name tables
+//! are uncopyrightable and explicitly intended to be reused in other projects,
+//! see for example this commit:
+//! <https://github.com/util-linux/util-linux/commit/16724407c858f4d3e2d2497f758b678b644c0011>
+
+const ARM_PARTS: &[(u64, &str)] = &[
```
mkroening: The original project uses `int` for the IDs. Would it make sense to use `u32` or even `u16` here? :)

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,284 @@
...
+		0x53 => ("Samsung", SAMSUNG_PARTS),
+		0x56 => ("Marvell", MARVELL_PARTS),
+		0x61 => ("Apple", APPLE_PARTS),
+		0x66 => ("Faraday", FARADAY_PARTS),
+		0x69 => ("Intel", INTEL_PARTS),
+		0x6d => ("Microsoft", MS_PARTS),
+		0x70 => ("Phytium", FT_PARTS),
+		0xc0 => ("Ampere", AMPERE_PARTS),
+		_ => return None,
+	};
```
mkroening: Why is this a `match` instead of another lookup table?

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,284 @@
...
+	(0x664, "FTC664"),
+	(0x862, "FTC862"),
+];
+
+const MS_PARTS: &[(u64, &str)] = &[(0xd49, "Azure-Cobalt-100")];
+
+pub fn pretty_implementer_and_part(
+	implementer_raw: u64,
+	part_raw: u64,
+) -> Option<(&'static str, Option<&'static str>)> {
```
mkroening, line 279: Related to the other types comment: The implementer field of MIDR_EL1 is 8 bits wide and part num is 12 bits. I think an `u8` and `u16` respectively could be nicer, no?

mkroening (changes requested):

src/arch/aarch64/kernel/lscpu.rs
Gelbpunkt: I agree, when I wrote this I had the same thought but didn't want to convince clippy that the truncation from the u64 returned by aarch64-cpu is fine. I'll change it :)

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,291 @@
...
+//!
+//! The lookup tables here are taken from util-linux:
+//! <https://github.com/util-linux/util-linux/blob/da322604a45f8094d439ea64be7d44cfea1f500a/sys-utils/lscpu-arm.c>
+//!
+//! util-linux is licensed under the GPL, but the implementer/model name tables
+//! are uncopyrightable and explicitly intended to be reused in other projects,
+//! see for example this commit:
+//! <https://github.com/util-linux/util-linux/commit/16724407c858f4d3e2d2497f758b678b644c0011>
+
+const ARM_PARTS: &[(u16, &str)] = &[
```
mkroening: I think `static` would be a better fit here.

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,291 @@
...
+	(0x661, "FTC661"),
+	(0x662, "FTC662"),
+	(0x663, "FTC663"),
+	(0x664, "FTC664"),
+	(0x862, "FTC862"),
+];
+
+const MS_PARTS: &[(u16, &str)] = &[(0xd49, "Azure-Cobalt-100")];
+
+const IMPLEMENTERS: &[(u8, &str, &[(u16, &str)])] = &[
```
mkroening: Clippy complains: https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity

You can decide whether to add a type definition or allow the lint here.

src/arch/aarch64/kernel/lscpu.rs
```diff
@@ -0,0 +1,291 @@
...
+		if *id == implementer_raw {
+			Some((*name, parts))
+		} else {
+			None
+		}
+	})?;
+
+	let part = parts
+		.iter()
+		.find_map(|(id, name)| if *id == part_raw { Some(*name) } else { None });
```
mkroening:
```suggestion
	let (implementer, parts) = IMPLEMENTERS
		.iter()
		.copied()
		.find_map(|(id, name, parts)| (id == implementer_raw).then_some((name, parts)))?;

	let part = parts
		.iter()
		.copied()
		.find_map(|(id, name)| (id == part_raw).then_some(name));
```

Though you can also leave it as is if you prefer.

mkroening (changes requested):

mkroening (approved):
Thanks! :)

PR #2413, opened 2026-05-06
feat(socket): update `MSG_*` constants

mkroening:
This updates our `MSG_*` constants from FreeBSD. While this changes `MSG_PEEK`, which is public through hermit-abi, libc, and Newlib, the constant has never been used in the kernel.


cagatay-y (approved):

mkroening:
This would create churn without immediate benefits.

PR #2415, opened 2026-05-07
feat(env): make `NO_COLOR` and `HERMIT_LOG_LEVEL_FILTER` respect the runtime environment

mkroening:
Depends on https://github.com/hermit-os/kernel/pull/2414.

PR #2417, opened 2026-05-07
docs: document environment variables

### LINKED ISSUE #2343: Environment variables are undocumented

mkroening:
Depends on a decision on https://github.com/hermit-os/kernel/pull/2415.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,40 @@
 )]
 #![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
 //!
+//! # Compile-time environment variables
+//!
+//! Compile-time environment variables are read from the build environment.
+//!
+//! - **`HERMIT_CAREFUL`** — Compiles the kernel with [`cargo careful`].
+//!
+//! [`cargo careful`]: https://github.com/RalfJung/cargo-careful
+//!
+//! # Run-time environment variables
+//!
+//! Run-time environment variables can be set via `env=KEY=VALUE` kernel args.
+//!
+//! ## Network environment variables
+//!
+//! - **`HERMIT_MTU`** — Sets the *maximum transmission unit* (MTU). Defaults to `1500`.
+//! - **`HERMIT_MRG_RXBUF_SIZE`** — Sets the receive buffer size. Useful for testing receive buffer merging. Defaults to unset.
+//! - **`HERMIT_IP`** — Sets the IPv4 address. Defaults to `10.0.5.3`.
+//! - **`HERMIT_GATEWAY`** — Sets the gateway IPv4 address. Defaults to `10.0.5.1`.
+//! - **`HERMIT_MASK`** — Sets the network mask. Defaults to `255.255.255.0`.
+//! - **`HERMIT_DNS1`** — Sets the first DNS server. Defaults to `9.9.9.9`.
+//! - **`HERMIT_DNS2`** — Sets the second DNS server. Defaults to `1.1.1.1`.
+//! - **`HERMIT_PCAP_PATH`** — Sets the packet capture file path. Defaults to `/root/`. See the `write-pcap-file` feature for details.
+//!
+//! ## Output environment variables
+//!
+//! - **`NO_COLOR`** — Prevents the addition of ANSI colors to the kernel output. Defaults to unset. For details, see [`NO_COLOR`].
+//! - **`HERMIT_LOG_LEVEL_FILTER`** — Sets the lowest log level to print. Defaults to `info`.
+//!
+//! [`NO_COLOR`]: https://no-color.org/
+//!
+//! ## Deprecated environment variables
+//!
+//! - **`UHYVE_MOUNT`** — Sets the Uhyve mount point. Defaults to `/root`.
+//!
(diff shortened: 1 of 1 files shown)


jounathaen (approved):

src/lib.rs
```diff
@@ -35,6 +35,40 @@
...
+//! 
+//! [`cargo careful`]: https://github.com/RalfJung/cargo-careful
+//! 
+//! # Run-time environment variables
+//! 
+//! Run-time environment variables can be set via `env=KEY=VALUE` kernel args.
+//! 
+//! ## Network environment variables
+//! 
+//! - **`HERMIT_MTU`** — Sets the *maximum transmission unit* (MTU). Defaults to `1514`.
```
Gelbpunkt: We should document that this definition of MTU includes the 14-byte ethernet header, contrary to e.g. Linux where the configured MTU is the IP MTU and excludes the ethernet header

See also the comment here: https://github.com/hermit-os/kernel/blob/a9e28f9198924dcaf39f5be35b29b04064bafd26/src/drivers/net/mod.rs#L50

Gelbpunkt: Alternatively it would probably be more intuitive to match Linux and make the user configure the IP MTU and default to 1500.

src/lib.rs
```diff
@@ -35,6 +35,40 @@
...
+//! ## Output environment variables
+//!
+//! - **`NO_COLOR`** — Prevents the addition of ANSI colors to the kernel output. Defaults to unset. For details, see [`NO_COLOR`].
+//! - **`HERMIT_LOG_LEVEL_FILTER`** — Sets the lowest log level to print. Defaults to `info`.
+//!
+//! [`NO_COLOR`]: https://no-color.org/
+//!
+//! ## Deprecated environment variables
+//!
+//! - **`UHYVE_MOUNT`** — Sets the Uhyve mount point. Defaults to `/root`.
```
cagatay-y, line 70: Is there a preferred alternative for the deprecated variable? Might be useful to mention if there is one.

src/lib.rs
```diff
@@ -35,6 +35,40 @@
...
+//! [`cargo careful`]: https://github.com/RalfJung/cargo-careful
+//!
+//! # Run-time environment variables
+//!
+//! Run-time environment variables can be set via `env=KEY=VALUE` kernel args.
+//!
+//! ## Network environment variables
+//!
+//! - **`HERMIT_MTU`** — Sets the *maximum transmission unit* (MTU). Defaults to `1500`.
+//! - **`HERMIT_MRG_RXBUF_SIZE`** — Sets the receive buffer size. Useful for testing receive buffer merging. Defaults to unset.
```
cagatay-y, line 53: We should probably clarify that the environment variable is only relevant for virtio-net devices and with a specific feature negotiated.

src/lib.rs
```diff
@@ -35,6 +35,40 @@
...
+//! - **`HERMIT_IP`** — Sets the IPv4 address. Defaults to `10.0.5.3`.
+//! - **`HERMIT_GATEWAY`** — Sets the gateway IPv4 address. Defaults to `10.0.5.1`.
```
cagatay-y, line 55: Since the DHCPv4 feature is enabled by default, it may be helpful to mention that the IP address set may be overwritten if addresses are successfully negotiated.

cagatay-y (commented):
Should this PR also change the README? `HERMIT_LOG_LEVEL_FILTER` is documented there too. Is it redundant or should they both stay? If both stay, should this one refer to the README for more information?

src/lib.rs
mkroening: This was done in https://github.com/hermit-os/kernel/pull/2421.

src/lib.rs
mkroening: I have added:

> Nowadays Uhyve supplies this to the kernel.

src/lib.rs
mkroening: I have added:

> Useful for testing receive buffer merging of virtio-net devices when the feature `VIRTIO_NET_F_MRG_RXBUF` is negotiated. Defaults to unset.

src/lib.rs
mkroening: I have added:

> Is only used when DHCP is not successful.

src/lib.rs
```diff
@@ -35,6 +35,40 @@
...
+//!
+//! - **`NO_COLOR`** — Prevents the addition of ANSI colors to the kernel output. Defaults to unset. For details, see [`NO_COLOR`].
+//! - **`HERMIT_LOG_LEVEL_FILTER`** — Sets the lowest log level to print. Defaults to `info`.
+//!
+//! [`NO_COLOR`]: https://no-color.org/
+//!
+//! ## Deprecated environment variables
+//!
+//! - **`UHYVE_MOUNT`** — Sets the Uhyve mount point. Defaults to `/root`. Nowadays Uhyve supplies this to the kernel.
+//!
```
jounathaen, line 71: I think we can also just silently drop this in this piece of new documentation.

jounathaen (approved):

src/lib.rs
mkroening: Since it still exists, I'd rather be complete with the docs. We can remove it once we remove it from the code.

PR #2422, opened 2026-05-08
feat(getsockopt): support `SO_SNDBUF` and `SO_RCVBUF` and more for iperf

cagatay-y:


mkroening (commented):
This is great, thanks! I tested it, and it works. I will take a look at the code later.

Two issues I noticed:

1. When Hermit is the server, it can only accept one client. After the first client is done, the next client cannot establish a connection. This does not show an error message, but might be related to the next issue.

2. When Hermit is the client, the test does not end cleanly:

   ```
   iperf3: error - unable to send control message - port may not be available, the other side may have stopped running, etc.: Function not implemented
   ```

These issues do not block this PR at all; I just wanted to document this.

cagatay-y:
I am not 100% sure if this is the same problem as the one you have with multiple connections on the server, but when I run the server with the `--one-off` option so that it shuts off after a test is completed, it gets stuck at https://github.com/esnet/iperf/blob/8a94f44cde986106abe8c8bf06d72d327570e34e/src/iperf_server_api.c#L488. I haven't had a chance to investigate further.

I need to check what the second issue is about.

mkroening (approved):
Thanks, this is great!

I split up the commits a bit and refactored some tiny things. I'll merge this tomorrow. :)

mkroening (approved):

PR #2427, opened 2026-05-11
fix(hermit-builtins): make math functions private

mkroening:
These are accessed through their symbol name, not through Rust.


mkroening:
Hmm. The syscalls are public for docs. Let's keep these public too, then, for consistency.

PR #2432, opened 2026-05-13
refactor(logging): make time formatting more explicit

cagatay-y:


mkroening (approved):
Thanks!

`Duration` is twice as large as the previous `u64`, but I think this is still fine. 👍

PR #2434, opened 2026-05-16
fix(vsock): restore listener after accept

### LINKED ISSUE #2433: vsock listener cannot accept a second connection
After a vsock server accepts and closes one connection, `accept()` returns `EBADF` (-9) on all subsequent calls. Any new connection attempt from the host is immediately reset (`ENODEV` on the client side).

## Expected behaviour

A vsock listener should accept multiple sequential connections, the same as TCP. This works correctly on Linux (see comparison below).

## Actual behaviour

- Connection 1: accepted, data exchanged, connection closed — works correctly.
- Connection 2: `accept()` returns `-9` (EBADF) immediately. The host client gets `ENODEV` (No such device, os error 19) when attempting to connect.

## Reproduction

The essential main logic in the vsock "server" compiled as a hermit-os application:

```rust
#[cfg(target_os = "hermit")]
use hermit as _;

/// Minimal vsock ping/pong server for hermit-os.
///
/// Listens on vsock port 1234. For each connection, reads one line and
/// writes "pong\n" back. Expected to handle multiple sequential connections.
///
/// Bug: after the first connection is closed, accept() returns EBADF (-9)
/// forever and any new connection attempt from the host gets RST immediately.
use hermit_abi::{
    accept, bind, listen, read, sa_family_t, sockaddr, sockaddr_vm, socket, socklen_t, write,
    AF_VSOCK, SOCK_STREAM, VMADDR_CID_ANY,
};
use std::mem::size_of;

const PORT: u32 = 1234;

fn main() {
    eprintln!("[pingpong] binding vsock port {PORT}");

    // Retry bind/listen until the vsock driver is ready (async PCI init).
    let listener_fd = 'bind: {
        for _ in 0..100 {
        let saddr = sockaddr_vm {
            svm_len: size_of::<sockaddr_vm>() as u8,
            svm_reserved1: 0,
            svm_family: AF_VSOCK as sa_family_t,
            svm_cid: VMADDR_CID_ANY,
            svm_port: PORT,
            svm_zero: [0; 4],
        };

        let fd = unsafe { socket(AF_VSOCK, SOCK_STREAM, 0) };
        if fd < 0 { std::thread::yield_now(); continue; }

        let r = unsafe { bind(fd, &saddr as *const _ as *const sockaddr, size_of::<sockaddr_vm>() as socklen_t) };
        if r < 0 { unsafe { hermit_abi::close(fd) }; std::thread::yield_now(); continue; }

        let r = unsafe { listen(fd, 128) };
        if r < 0 { unsafe { hermit_abi::close(fd) }; std::thread::yield_now(); continue; }

        eprintln!("[pingpong] listener fd={fd}");
        break 'bind fd;
        }
        eprintln!("[pingpong] failed to bind vsock port {PORT} after 100 attempts");
        std::process::exit(1);
    };

    let mut conn_count = 0u32;
    loop {
        eprintln!("[pingpong] waiting for connection #{}", conn_count + 1);

        let mut peer_addr = sockaddr_vm {
            svm_len: size_of::<sockaddr_vm>() as u8,
            svm_reserved1: 0,
            svm_family: AF_VSOCK as sa_family_t,
            svm_cid: 0,
            svm_port: 0,
            svm_zero: [0; 4],
        };
        let mut peer_len = size_of::<sockaddr_vm>() as socklen_t;

        let conn_fd = unsafe {
            accept(listener_fd, &mut peer_addr as *mut _ as *mut sockaddr, &mut peer_len)
        };

        if conn_fd < 0 {
            eprintln!("[pingpong] accept failed: {conn_fd} — listener fd={listener_fd} is broken after first connection");
            eprintln!("[pingpong] BUG: vsock listener cannot accept a second connection");
            std::process::exit(1);
        }

        conn_count += 1;
        eprintln!("[pingpong] accepted connection #{conn_count} fd={conn_fd} from cid={}", peer_addr.svm_cid);

        let mut buf = [0u8; 64];
        let n = unsafe { read(conn_fd, buf.as_mut_ptr(), buf.len()) };
        if n > 0 {
            let msg = std::str::from_utf8(&buf[..n as usize]).unwrap_or("<invalid utf8>");
            eprintln!("[pingpong] received ({n} bytes): {msg:?}");
        } else {
            eprintln!("[pingpong] read returned {n}");
        }

        let pong = b"pong\n";
        let w = unsafe { write(conn_fd, pong.as_ptr(), pong.len()) };
        eprintln!("[pingpong] wrote pong: {w} bytes");

        eprintln!("[pingpong] closing conn fd={conn_fd}");
        unsafe { hermit_abi::close(conn_fd) };
        eprintln!("[pingpong] conn fd={conn_fd} closed");
    }
}
```

Then a small client that I've compiled as a regular linux binary:
```
/// Minimal vsock ping tool for testing vsock-pingpong unikernel.
///
/// Usage: vsock-ping <cid> <port> [count]
use std::io::{Read, Write};
use vsock::VsockStream;

fn main() {
    let args: Vec<String> = std::env::args().collect();
    if args.len() < 3 {
        eprintln!("usage: vsock-ping <cid> <port> [count]");
        std::process::exit(1);
    }
    let cid: u32 = args[1].parse().expect("cid must be u32");
    let port: u32 = args[2].parse().expect("port must be u32");
    let count: usize = args.get(3).and_then(|s| s.parse().ok()).unwrap_or(2);

    let mut all_ok = true;
    for i in 1..=count {
        eprintln!("--- ping {i}/{count} ---");
        match do_ping(cid, port) {
            Ok(response) => {
                println!("response: {response:?}");
                eprintln!("--- ok ---");
            }
            Err(e) => {
                eprintln!("--- FAILED: {e} ---");
                all_ok = false;
            }
        }
        if i < count {
            std::thread::sleep(std::time::Duration::from_secs(1));
        }
    }

    if !all_ok {
        std::process::exit(1);
    }
}

fn do_ping(cid: u32, port: u32) -> Result<String, Box<dyn std::error::Error>> {
    eprintln!("connecting vsock cid={cid} port={port}...");
    let mut stream = VsockStream::connect_with_cid_port(cid, port)?;
    eprintln!("connected, sending ping...");
    stream.write_all(b"ping\n")?;
    eprintln!("sent, reading response...");
    let mut buf = [0u8; 64];
    let n = stream.read(&mut buf)?;
    Ok(String::from_utf8_lossy(&buf[..n]).into_owned())
}

```

### Steps

**Terminal 1** — build and boot the unikernel. I have used QEMU with `vhost-vsock-pci`:

```sh
cargo build --release;
qemu-system-x86_64 -enable-kvm -cpu host  -m 64M -kernel "$LOADER"  -initrd "target/x86_64-unknown-hermit/release/vsock-pingpong" -device "vhost-vsock-pci,guest-cid=99"  -nographic -serial stdio -monitor none  -no-reboot
```

**Terminal 2** — build and run the vsock-client that send two sequential pings once the VM prints `waiting for connection #1`:

```sh
cargo build;
target/debug/vsock-ping 99 1234 2
```

### Observed output (after boot output)

VM:
```[pingpong] binding vsock port 1234
[pingpong] listener fd=3
[pingpong] waiting for connection #1
[pingpong] accepted connection #1 fd=4 from cid=2
[pingpong] received (5 bytes): "ping\n"
[pingpong] wrote pong: 5 bytes
[pingpong] closing conn fd=4
[pingpong] conn fd=4 closed
[pingpong] waiting for connection #2
[pingpong] accept failed: -9 — listener fd=3 is broken after first connection
[pingpong] BUG: vsock listener cannot accept a second connection
Number of interrupts
[0][FPU]: 1
[0][virtio]: 3
exit status 1
```

Client output:
```
--- ping 1/2 ---
connecting vsock cid=99 port=1234...
connected, sending ping...
sent, reading response...
response: "pong\n"
--- ok ---
--- ping 2/2 ---
connecting vsock cid=99 port=1234...
--- FAILED: No such device (os error 19) ---
```

---

I did setup a similar ping server that I compiled against linux and it manages to handle the second accept:

```
[pingpong] listener fd=3
[pingpong] waiting for connection #1
[pingpong] accepted connection #1 fd=4 from cid=1
[pingpong] received (5 bytes): "ping\n"
[pingpong] wrote pong: 5 bytes
[pingpong] closing conn fd=4
[pingpong] conn fd=4 closed
[pingpong] waiting for connection #2
[pingpong] accepted connection #2 fd=4 from cid=1
[pingpong] received (5 bytes): "ping\n"
[pingpong] wrote pong: 5 bytes
[pingpong] closing conn fd=4
[pingpong] conn fd=4 closed
[pingpong] waiting for connection #3
```

archevel:
After accepting a connection, move the accepted socket to an ephemeral port and reset the listener entry to Listen state. Track the listen port separately in Socket so subsequent accept calls always find the listener entry regardless of how many connections have been accepted.

Depends on https://github.com/hermit-os/hermit-rs/pull/996.
Pre-Rebase: https://github.com/hermit-os/kernel/commit/b10475e1cd5fefe26db1c4a2683832ffeac07ac9

Fixes #2433

### DIFF
--- a/src/executor/vsock.rs
+++ b/src/executor/vsock.rs
@@ -136,6 +136,11 @@ async fn vsock_run() {
 					raw.state = VsockState::Connected;
 					raw.peer_buf_alloc = header.buf_alloc.to_ne();
 					raw.peer_fwd_cnt = header.fwd_cnt.to_ne();
+					// The blocking `connect()` future parks on `rx_waker` (see
+					// `Socket::connect`'s Connecting arm), so it MUST be woken
+					// here or the connect never returns. Wake `tx_waker` too:
+					// a freshly-connected socket is immediately writable.
+					raw.rx_waker.wake();
 					raw.tx_waker.wake();
 				}
 			} else if raw.remote_cid == header_cid {
--- a/src/executor/vsock.rs
+++ b/src/executor/vsock.rs
@@ -1,4 +1,4 @@
-use alloc::collections::{BTreeMap, btree_map};
+use alloc::collections::{BTreeMap, VecDeque, btree_map};
 use alloc::vec::Vec;
 use core::future;
 use core::task::Poll;
@@ -13,6 +13,7 @@ use crate::arch::kernel::mmio as hardware;
 use crate::drivers::pci as hardware;
 use crate::errno::Errno;
 use crate::executor::{WakerRegistration, spawn};
+use crate::fd::socket::SOMAXCONN;
 use crate::io;
 
 pub(crate) static VSOCK_MAP: InterruptTicketMutex<VsockMap> =
@@ -34,6 +35,30 @@ pub(crate) enum VsockState {
 
 pub(crate) const RAW_SOCKET_BUFFER_SIZE: usize = 256 * 1024;
 
+/// Default depth of a listener's pending-connection backlog, used when
+/// `accept()` is called without an explicit `listen()`. Mirrors the TCP
+/// socket's `DEFAULT_BACKLOG`.
+pub(crate) const DEFAULT_BACKLOG: usize = 128;
+
+/// A pending inbound connection request waiting to be `accept()`ed, captured
+/// from an `Op::Request` packet.
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct PendingRequest {
+	pub remote_cid: u32,
+	pub remote_port: u32,
+	pub peer_buf_alloc: u32,
+}
+
+/// Identifies an established connection by its local (listen) port and the
+/// remote endpoint `(remote_cid, remote_port)`. Multiple connections may share
+/// one local port, mirroring how TCP demultiplexes by the connection 4-tuple.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct ConnKey {
+	pub local_port: u32,
+	pub remote_cid: u32,
+	pub remote_port: u32,
+}
+
 #[derive(Debug)]
 pub(crate) struct RawSocket {
 	pub remote_cid: u32,
@@ -46,6 +71,12 @@ pub(crate) struct RawSocket {
 	pub rx_waker: WakerRegistration,
 	pub tx_waker: WakerRegistration,
 	pub buffer: Vec<u8>,
+	/// Inbound connection requests queued on a listener, awaiting `accept()`.
+	/// Only used by listener sockets; empty for connections.
+	pub pending: VecDeque<PendingRequest>,
+	/// Maximum depth of `pending` for a listener (set by `listen()`, clamped to
+	/// `SOMAXCONN`). Further requests beyond this are reset.
+	pub backlog: usize,
 }
 
 impl RawSocket {
@@ -61,6 +92,8 @@ impl RawSocket {
 			rx_waker: WakerRegistration::new(),
 			tx_waker: WakerRegistration::new(),
 			buffer: Vec::with_capacity(RAW_SOCKET_BUFFER_SIZE),
+			pending: VecDeque::new(),
+			backlog: DEFAULT_BACKLOG,
 		}
 	}
 }
@@ -75,6 +108,9 @@ async fn vsock_run() {
 		let mut driver_guard = driver.lock();
 		let mut hdr: Option<Hdr> = None;
 		let mut fwd_cnt: u32 = 0;
+		// Force the reply to be `Op::Rst` regardless of the incoming op (used
+		// for packets addressed to a socket that no longer exists).
+		let mut force_rst: bool = false;
 
 		driver_guard.process_packet(|header, data| {
 			let op = Op::try_from(header.op.to_ne()).unwrap();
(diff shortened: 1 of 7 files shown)


archevel:
As a reference I made this pull request as well with a regression test:
https://github.com/hermit-os/hermit-rs/pull/996

mkroening (changes requested):
Thanks for finding and fixing this bug! :)

Sorry for the late reply. I can reproduce the fix.

Regarding the implementation:
If I understand correctly, your changes now track two ports for listening sockets, one for the listener and one for the last connection, right? This does not support multiple open connections at the same time, right?

Perhaps doing something similar to TCP would be more useful: return a proper connection FD from accept instead of the current dummy `NullSocket`.

Regarding the test:
Could you adapt the existing vsock tests instead?

archevel:
No worries! I appreciate you taking the time to have look. I've kept working on this locally and had to do some additional changes. Those are out of scope for this I think, as they involved needing to patch libc and mio to handle epoll for hermit. I'm working on that in order to be able to use firecracker to run a hermit based vm and communicate over vsock. Anyway, that is a larger piece of work, but I think I actually ended up aligning this more with how the TCP is handled. I'll take a peek and adjust this PR.

archevel:
I've rebased the changes on main. I then changed to align it more with how the tcp socket works (handling multiple concurrent connections). Let me know if there are any other issues I should address/fix! The CI failed, so I'll take a peek and see if I understand why.

archevel:
I _think_ the CI fails because it isn't aligned with the changes in https://github.com/hermit-os/hermit-rs/pull/996

archevel:
~~@mkroening I realized that the code doesn't really handle multiple cids currently. E.g. If the host assigns two vsock cids, say 3 and 4. Currently listening on either cid on port 1234 will get connections for both I think. Might be a separate issue to solve.~~ 

Edit: I realized this wasn't a real issue since qemu can only assign a single cid to a guest with vhost-vsock. However, the code did not fail on binding on a non-assigned cid. I've now added so that a bind on e.g. cid 4 when cid 3 was assigned fails. Binding on VMADDR_CID_ANY works. Arguably there might be a case for allowing binding on CID 1 (loopback), but since hermit apps as far as I understand are essentially single processes there ought to be better ways to handle internal messages....

archevel:
@mkroening I'm not entierly familiar with the setup, but this PR won't pass the CI unless https://github.com/hermit-os/hermit-rs/pull/996 is merged. However, merging that will likely cause OTHER PRs to start failing before merging this PR... Is there anything I should amend to make this smoother?

archevel:
I ended up using claude to aid me with the accept backlog and try to align it with the tcp code. Not sure if that is acceptable? In the process I discovered that mio doesn't handle select in the same way for the native socket types (e.g. TcpStream wrapped in IoSource<T>) as it does for a raw file descriptor in a mio::unix::SourceFd. This causes an issue where the reader is never woken up even when data is arriving. I made a workaround in the app I am working on. Should I raise it as an issue against the kernel? Not sure exactly where the fix should go (might be in mio).

mkroening:
> needing to patch libc and mio to handle epoll for hermit

We do have patched versions of socket2 and tokio to make them work on Hermit ([hermit-rs/Cargo.toml#L35-L37](https://github.com/hermit-os/hermit-rs/blob/38548bda2f0a195e06501479440abeb7d7579d14/Cargo.toml#L35-L37)]. We do have some mio patches upstream already, though I need to rework our system call story in the winter.

> Arguably there might be a case for allowing binding on CID 1 (loopback), but since hermit apps as far as I understand are essentially single processes there ought to be better ways to handle internal messages....

Yeah, I think we can postpone that for now. :)

> I ended up using claude to aid me with the accept backlog and try to align it with the tcp code. Not sure if that is acceptable?

As long as you can stand behind the code as if you wrote it, it's fine. I need to understand it as well, after all. Please remove the Claude authorship from the commit (I can also do this before merging, if you prefer).

> In the process I discovered that mio doesn't handle select in the same way for the native socket types (e.g. TcpStream wrapped in IoSource) as it does for a raw file descriptor in a mio::unix::SourceFd. This causes an issue where the reader is never woken up even when data is arriving. I made a workaround in the app I am working on. Should I raise it as an issue against the kernel? Not sure exactly where the fix should go (might be in mio).

Yes, please go ahead and open an issue.

I will do another review today. Sorry for the delay again! Could you rebase on `main` instead of merging? Otherwise, the PR shows many unrelated commits from `main` in this PR's merge-base diff. Meanwhile, I will take a look at the direct diff ([`badd4c4..7ac357f`](https://github.com/hermit-os/kernel/compare/badd4c42cd28819662d9f407860219887854d17c..7ac357f0a50fcbce693b8c31a8adcfba5712e297)).

archevel:
> Sorry for the delay again! 

Absolutely, no worries! I have this working my machine and that's good enough for me :)

> Could you rebase on main instead of merging?

Done! I rebased on main and squashed the minor fixup commits into their parents. 

> Yes, please go ahead and open an issue.

Raised it here: https://github.com/hermit-os/kernel/issues/2533 
It is not a blocker for me since I used the workaround.

.github/workflows/ci.yml
```diff
@@ -289,6 +289,14 @@ jobs:
         working-directory: .
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} ${{ matrix.rs_flags }} --package vsock qemu ${{ matrix.qemu_flags }} --sudo --devices virtio-vsock-pci
         if: matrix.arch != 'riscv64'
+      # Run check if required server feature is present in hermit-rs for vsock example otherwise skips the test which would fail
```
archevel: I think this is worth calling out. I added this to gate the test code on if the example code in hermit-rs repo is updated. This could be removed once that's merged. Perhaps also removing the default example.

src/executor/vsock.rs
```diff
@@ -24,11 +24,39 @@ pub(crate) enum VsockState {
...
+pub(crate) struct PendingRequest {
+	pub remote_cid: u32,
+	pub remote_port: u32,
+	pub peer_buf_alloc: u32,
+}
+
+/// Identifies an established connection by its local (listen) port and the
+/// remote endpoint `(remote_cid, remote_port)`. Multiple connections may share
+/// one local port, mirroring how TCP demultiplexes by the connection 4-tuple.
+pub(crate) type ConnKey = (u32, u32, u32);
```
mkroening: Could you make this a struct instead?

mkroening (changes requested):
Thanks, this is looking great! Thanks again for the patience. I only have one more nit regarding the code, then I will merge it.

Regarding the CI: I would merge https://github.com/hermit-os/hermit-rs/pull/996 before merging this, so we don't need any conditionals in CI.

archevel:
> Thanks, this is looking great! Thanks again for the patience. I only have one more nit regarding the code, then I will merge it.
> 
> Regarding the CI: I would merge [hermit-os/hermit-rs#996](https://github.com/hermit-os/hermit-rs/pull/996) before merging this, so we don't need any conditionals in CI.

Awesome! I rebased the changes and changed to a struct instead of the tuple for the ConnKey. As mentioned this PR will fail CI until [hermit-os/hermit-rs#996](https://github.com/hermit-os/hermit-rs/pull/996) get's merged.

mkroening:
I'll take a last look tomorrow morning. :)

Meanwhile, would you mind telling me which coding assistant you used, so we can properly [attribute](https://docs.kernel.org/process/coding-assistants.html) it in the commit messages? I have done some rebasing locally and can add the attribution.

archevel:
I have used a mix of models from Anthropic. The model I most often use is Opus 4.7 (now 5, but started on 4.7 I believe). I did use Fable for aligning it with the TCP socket code.

mkroening (approved):
Thanks! Reviewing LLM code (and comments) requires extra caution, so this took a while, but I am happy to get this fix in. :)

PR #2435, opened 2026-05-18
perf(tcp): optimize active socket search

cagatay-y:
Increases iperf performance by ~15% in a non-scientific test.


mkroening (approved):
Thanks! :)

PR #2444, opened 2026-05-27
fix(idle-poll): wake the network task unconditionally

cagatay-y:
In order for the kernel to poll the NIC when idle as desired when the idle-poll feature is enabled, we need to wake the task unconditionally.

### DIFF
--- a/src/executor/network.rs
+++ b/src/executor/network.rs
@@ -246,19 +246,16 @@ async fn network_run() {
 
 		let now = now();
 
-		match nic.poll_common(now) {
-			PollResult::SocketStateChanged => {
-				// Progress was made
-				cx.waker().wake_by_ref();
-			}
-			PollResult::None => {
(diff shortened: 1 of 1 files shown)


mkroening (approved):
Thanks! :)

Is this fixing a failure that we could detect in CI or a slowdown only?

cagatay-y:
I noticed it when testing with broken interrupts on the MSI-X branch. We could consider disabling network interrupts when the `idle-poll` feature is enabled (that may make sense in general, aside from testing) but I am not sure if that would ensure that the network task timer is never enabled and does not mask the lack of idle-polling. I probably should have pinged @Gelbpunkt, as she has more experience with the network wakers.

Gelbpunkt (approved):
This seems to do what the `idle-poll` feature is documented to do, asides from the fact that we still have interrupts enabled from the NIC. I'm experimenting with that in #2463, so I wouldn't do it as part of this PR.

PR #2292, opened 2026-03-03
ci: try to reproduce spurious httpd timeout

mkroening:
This PR tries to reproduce the following spurious httpd timeout:

```console
$ cargo xtask ci rs --arch x86_64 --profile release  --package httpd --features ci,hermit/virtio-net qemu --accel --sudo --sudo --devices virtio-net-pci --tap
[...]
[    0.341513][0][INFO  hermit    ] Jumping into application
Now listening on port 9975
[CI] GET http://10.0.5.3:9975/
Error: timeout: global
Error: Process completed with exit code 1.
```

PR #2445, opened 2026-05-29
fix(tcp): only accept connections that may send

### LINKED ISSUE #906: Glitch in interrupt handling? Missed interrupt?
Test program Cargo.toml:
```
[package]
name = "testtcp"
version = "0.1.0"
edition = "2021"

[dependencies]

[target.'cfg(target_os = "hermit")'.dependencies.hermit]
path = "../rusty-hermit/hermit"
```

Test program src/main.rs:
```
#[cfg(target_os = "hermit")]
use hermit as _;

use std::net::TcpListener;

fn main() {
    let listener = TcpListener::bind("127.0.0.1:8080").unwrap();
    loop {
        listener.accept().unwrap();
    }
}
```

rusty-hermit is at f7b3c51cbb3cc4d2cd4c199370e53cb714c5e951
rusty-hermit/kernel is at 5dd731029f1d23c50dbd80c3c04dc7a95ed5bffc

Build and run with Hermit on Intel:
```
cargo build --target x86_64-unknown-hermit --release \
  -Z build-std=std,core,alloc,panic_abort \
  -Z build-std-features=compiler-builtins-mem

qemu-system-x86_64 \
    -cpu qemu64,apic,fsgsbase,fxsr,rdrand,rdtscp,xsave,xsaveopt \
    -smp 1 -m 512M \
    -device isa-debug-exit,iobase=0xf4,iosize=0x04 \
    -display none -serial stdio \
    -kernel ../rusty-loader/target/x86_64/release/hermit-loader \
    -initrd target/x86_64-unknown-hermit/release/testtcp \
    -netdev user,id=u1,hostfwd=tcp::8080-:8080 \
    -device virtio-net-pci,netdev=u1,disable-legacy=on
```

Test from another terminal:
```
while socat -u TCP:localhost:8080 CREATE:out ; do date ; done | uniq -c
```

What I see is initially about 120 connections accepted every second, but soon it drops to about 60, and every now and then something goes wrong, for example:
```
     72 Mon Sep 11 12:18:25 UTC 2023
    139 Mon Sep 11 12:18:26 UTC 2023
    131 Mon Sep 11 12:18:27 UTC 2023
    116 Mon Sep 11 12:18:28 UTC 2023
    102 Mon Sep 11 12:18:29 UTC 2023
     93 Mon Sep 11 12:18:30 UTC 2023
     86 Mon Sep 11 12:18:31 UTC 2023
     80 Mon Sep 11 12:18:32 UTC 2023
     76 Mon Sep 11 12:18:33 UTC 2023
     69 Mon Sep 11 12:18:34 UTC 2023
     41 Mon Sep 11 12:18:35 UTC 2023
     69 Mon Sep 11 12:18:53 UTC 2023   <-- 17 seconds with no connection here
    130 Mon Sep 11 12:18:54 UTC 2023
     67 Mon Sep 11 12:18:55 UTC 2023
```

Perhaps an interrupt is dropped and the program gets rescued by a timer?

I get very similar behaviour on Arm if I avoid #898 by moving the initialisation of `cursor` as described at 2023-09-11 11:38 +0100 in that issue.

jounathaen:
This PR prevents returning sockets from `accept()` that are in the `SYN-RECEIVED` state and have thus not been `ESTABLISHED` yet and might never be. This is causing https://github.com/hermit-os/kernel/issues/906, but has also been observed in other scenarios.

When we listen for connections, the following state transitions take place:

```
      TCP A                                                TCP B

  1.  CLOSED                                               LISTEN

  2.  SYN-SENT    --> <SEQ=100><CTL=SYN>               --> SYN-RECEIVED

  3.  ESTABLISHED <-- <SEQ=300><ACK=101><CTL=SYN,ACK>  <-- SYN-RECEIVED

  4.  ESTABLISHED --> <SEQ=101><ACK=301><CTL=ACK>       --> ESTABLISHED

  5.  ESTABLISHED --> <SEQ=101><ACK=301><CTL=ACK><DATA> --> ESTABLISHED

          Basic 3-Way Handshake for Connection Synchronization

                                Figure 7.
```

See [RFC 793](https://datatracker.ietf.org/doc/html/rfc793#autoid-19) for details.

If, for whatever reason, we don't get SYN,ACK in 3., we should not return the socket to the application, since we might never get it and thus the connection has not been `ESTABLISHED` yet. Currently we do, since we currently use [`Socket::is_active`](https://docs.rs/smoltcp/0.13.1/smoltcp/socket/tcp/struct.Socket.html#method.is_active) to decide whether it is ready to be returned. That method returns true if the socket is not `CLOSED`, `TIME-WAIT` or `LISTEN`.

This PR avoids returning such sockets to the application by using [`Socket::may_send`](https://docs.rs/smoltcp/0.13.1/smoltcp/socket/tcp/struct.Socket.html#method.may_send) instead, which is true if the socket is `ESTABLISHED` or `CLOSE-WAIT`.

`CLOSE-WAIT` explicitly makes sense and can be produced when the other side is opening a connection and closes it immediately. Still, data can be received that was sent intermittently, and data can be sent, since it was `ESTABLISHED` before.

Closes https://github.com/hermit-os/kernel/issues/906.

### DIFF
--- a/src/fd/socket/tcp.rs
+++ b/src/fd/socket/tcp.rs
@@ -317,7 +317,16 @@ impl ObjectInterface for Socket {
 			let socket_handle = self
 				.handle
 				.extract_if(.., |handle| {
-					nic.get_mut_socket::<tcp::Socket<'_>>(*handle).is_active()
+					let socket = nic.get_mut_socket::<tcp::Socket<'_>>(*handle);
+
+					// We are checking not only for ESTABLISHED, but also for CLOSE-WAIT.
+					//
+					// CLOSE-WAIT may occur here when the other side connects to us and
+					// immediately closes their transmit half (our receive half),
+					// potentially sending something in between. It is still useful for
+					// the application to be able to read any received data and to send a
+					// response.
+					socket.may_send()
 				})
 				.next();
 


src/fd/socket/tcp.rs
```diff
@@ -317,7 +317,10 @@ impl ObjectInterface for Socket {
 			let socket_handle = self
 				.handle
 				.extract_if(.., |handle| {
-					nic.get_mut_socket::<tcp::Socket<'_>>(*handle).is_active()
+					matches!(
+						nic.get_mut_socket::<tcp::Socket<'_>>(*handle).state(),
+						tcp::State::Established | tcp::State::CloseWait
```
mkroening: @stlankes suggested using [`Socket::may_send`](https://docs.rs/smoltcp/latest/smoltcp/socket/tcp/struct.Socket.html#method.may_send), which exactly matches the functionality.

mkroening: One thing I am wondering about now is the symmetry with [`Socket::may_recv`](https://docs.rs/smoltcp/latest/smoltcp/socket/tcp/struct.Socket.html#method.may_recv).

Should we check for the union of states (`may_send() || may_recv()`) or the intersection (`state() == State::Established`)? Or is there a specific reason we need only `may_send`?

mkroening:
The issue this fixes is different from https://github.com/hermit-os/kernel/pull/2292, unfortunately.

mkroening:
This does fix https://github.com/hermit-os/kernel/issues/906, though. 🥳

src/fd/socket/tcp.rs
mkroening: After some more testing around https://github.com/hermit-os/kernel/issues/906 as well as with a separate scenario, I believe that we should accept CLOSE-WAIT sockets. The other side might already have closed their send side, but the socket may still contain data to receive, and we can still answer.

mkroening:
I have adjusted the code according to @stlankes's suggestion and added a comment explaining the logic. I have also adjusted the PR description.

mkroening (approved):
This is great! Thanks so much for putting in the work for this! :)

PR #2447, opened 2026-05-31
refactor(virtio): make devices generic over transport

cagatay-y:
Instead of relying on importing one of the PCI and MMIO transport modules that implement identically named structs to select the transport used by VIRTIO device drivers, use generics. This is in preparation for removing the mutual exclusivity of PCI and MMIO-based devices.


cagatay-y:
I've marked the PR as draft because before we can allow MMIO and PCI drivers to coexist, we need to decide how we want to handle cases where there are multiple devices of a kind (e.g. NICs, console devices, filesystems, etc.). It shouldn't hurt to merge the changes before we make those decisions but it is also not particularly useful.

PR #2453, opened 2026-06-01
fix(physicalmem): don't claim low addresses

zyuiop:
## Context

While working on patches for AMD SEV-SNP, we use the DeviceMapper with an offset early at boot to allocate a (plaintext) page to communicate with the hypervisor (the GHCB).

We observed that _randomly_ an application compiled with hermit would crash or enter a spin-loop sometimes _around_ the introduction of the IDT. This bug was _random_ in that recompilation with some changes at unrelated places in the code could eliminate the bug. This made debugging "fun" (as in: adding a print makes the bug disappear fun).

This clearly _screams_ of "memory corruption" somewhere. But where, actually? I don't really know, but this patch is an attempt at fixing one possible source.

## What this does, and why?

This fix is actually two fixes in one.

- First, `map_frame_range` (in `physicalmem.rs`) will, when an offset is used, map all pages _also_ at a fixed offset. This will obviously require spawning new page tables, and it turns out that the physical free list may at this point contain `0x0` as a valid frame, which may get picked up. I don't know if it's bad-bad, but since the page table is identity mapped, this means that accessing the page table entry requires dereferencing virtual address 0, aka the null pointer, which certainly cannot be good.

- Second, it turns out that the SMP boot code has a fixed address of `0x8000`. This address is only the 8th frame in the free list, so you can be sure it is going to be used, either by the offset mapping, or later on when mapping the heap. However, when the SMP boot code is loaded, it will _overwrite_ that page table entry, and corrupt the page table. I guess this bug can happen if you use a heap intensive application, it will reach a point where it accesses a part of the heap which is corrupt and crash. Maybe. I'm not certain.

To avoid these two problems, we propose a quickfix: just ignore the first 2 MB of physical addresses.

## What else?

The `detect_from_fdt` method looks a bit "dangerous" to me, in that it detects memory and immediately uses it to map it (sometimes). In most cases, no mapping is created, because UEFI already identity mapped everything, but in the cases where it does use it immediately, it may use pages that should be reserved and would have left the free list later in the procedure. The fact that pages are mapped immediately _may_ cause issues. But not sure.

## UPDATE

I now believe that the memory corruption was more likely happening at the end of the stack, which happened to be the page table - see #2462. However, I am still not sure it's a good idea to have 0x0 used as a page table, and the possible corruption with SMP is still there.

### DIFF
--- a/src/mm/physicalmem.rs
+++ b/src/mm/physicalmem.rs
@@ -9,7 +9,7 @@ use memory_addresses::{PhysAddr, VirtAddr};
 
 #[cfg(target_arch = "x86_64")]
 use crate::arch::mm::paging::PageTableEntryFlagsExt;
-use crate::arch::mm::paging::{self, HugePageSize, PageSize, PageTableEntryFlags};
+use crate::arch::mm::paging::{self, HugePageSize, LargePageSize, PageSize, PageTableEntryFlags};
 use crate::env;
 use crate::mm::device_alloc::DeviceAlloc;
 use crate::mm::{PageRangeAllocator, PageRangeBox};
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
Good find! :)

I think it would be preferable to only exclude the null pointer page and allocate memory at the required memory for the boot code. Not sure if that is easily possible with the current boot flow, though. What do you think?

mkroening (approved):
Thanks and sorry for the wait! :)

I have opened an issue for my concern:
- https://github.com/hermit-os/kernel/issues/2661

We can take care of this later, though. I have rebased and inlined the constant. :)

PR #2454, opened 2026-06-01
feat: stub inode numbers and fcntl result

zyuiop:
In the process of porting some applications (C) for Hermit, we had to implement a few stub values. I have gated them behind a compilation flag.
In addition, we also fix a few problems/incompatibilities.

### DIFF
--- a/src/fs/mem.rs
+++ b/src/fs/mem.rs
@@ -8,6 +8,7 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
 use core::mem::MaybeUninit;
+use core::sync::atomic::{AtomicU64, Ordering};
 use core::{mem, ptr};
 
 use align_address::Align;
@@ -35,6 +36,8 @@ impl RomFileInner {
 		}
 	}
 }
+
+static VFS_INO_NUM: AtomicU64 = AtomicU64::new(10_000);
 
 pub struct RomFileInterface {
 	/// Position within the file
@@ -306,6 +309,7 @@ impl RomFile {
 			st_atim: t,
 			st_mtim: t,
 			st_ctim: t,
+			st_ino: VFS_INO_NUM.fetch_add(1, Ordering::AcqRel),
 			..Default::default()
 		};
 
@@ -361,6 +365,9 @@ impl RamFile {
 			st_atim: t,
 			st_mtim: t,
 			st_ctim: t,
+			st_nlink: 1,
+			st_blksize: 4096,
+			st_ino: VFS_INO_NUM.fetch_add(1, Ordering::AcqRel),
 			..Default::default()
 		};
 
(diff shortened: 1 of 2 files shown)


src/syscalls/mod.rs
```diff
@@ -717,6 +717,9 @@ pub extern "C" fn sys_fcntl(fd: i32, cmd: i32, arg: i32) -> i32 {
 				.map_or_else(|e| -i32::from(e), |()| 0)
 			},
 		)
+	} else if cfg!(feature = "syscall-fake-values") {
+		warn!("[syscall-fake-values] Unknown fcntl flag {cmd} {arg}, returning 0");
+		0
```
mkroening: What scenario needs this? Which command is provided?

src/syscalls/mod.rs
```diff
@@ -482,10 +482,6 @@ pub unsafe extern "C" fn sys_access(name: *const c_char, flags: i32) -> i32 {
 		return -i32::from(Errno::Inval);
 	};
 
-	if access_option.contains(AccessOption::F_OK) && access_option != AccessOption::F_OK {
-		return -i32::from(Errno::Inval);
-	}
```
mkroening, line 487: Good find! 👍

src/fs/mem.rs
```diff
@@ -361,6 +368,12 @@ impl RamFile {
 			st_atim: t,
 			st_mtim: t,
 			st_ctim: t,
+			#[cfg(feature = "syscall-fake-values")]
+			st_nlink: 1,
+			#[cfg(feature = "syscall-fake-values")]
+			st_blksize: 4096,
+			#[cfg(feature = "syscall-fake-values")]
+			st_ino: VFS_INO_NUM.fetch_add(1, Ordering::AcqRel),
```
mkroening: Which program needs these values? I wonder whether we should just fake them all the time.

mkroening (changes requested):
Thanks! :)

zyuiop:
Sadly I don't precisely remember which programs required these stubs :(

It has to be one of the following: Lighttpd, Valkey, or Axum. 

Sorry for not being more precise

mkroening:
No worries! I have cherry-picked the noncontroversial changes:

- https://github.com/hermit-os/kernel/pull/2482
- https://github.com/hermit-os/kernel/pull/2483
- https://github.com/hermit-os/kernel/pull/2484

Here, I'd also like to avoid the feature. I think it would be best to identify the programs that need the specific workarounds or argue that the stub values are always fine.

zyuiop:
I'll remove the feature flag, I only added it when making this PR public because I did not know if you would like having the stubs unconditionnally :)

zyuiop:
I'll update this branch to remove the three cherry picked commits, and remove the feature flag

mkroening:
Would it be easy for you to test and check which program needs the inodes and which program needs which specific fcntl operations? :)

If not, we can merge it as is, but I would prefer being explicit about these purpose of these kinds of hacks.

zyuiop:
I'll set this back to draft until I can find the program that required it, at which point I will submit it as example :)

PR #2455, opened 2026-06-01
feat: support non bool values in setsockopt

zyuiop:
Add non boolean values in setsockopt.

Initially, I used a similar structure for getsockopt but I rebased on the latest PRs and removed that to rely on `c_int` only, hoping we don't implement string sockopt values anytime soon. However, I have for now kept this flexibility in setsockopt. We may want to rewrite that to use `c_int` instead, for consistency?

This PR has been split from #2454 ~~but relies on its feature flag (not completely critical to be fair)~~ (edit: 08.2026 -- no longer the case)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -926,6 +926,7 @@ dependencies = [
  "virtio-spec",
  "volatile 0.6.1",
  "x86_64",
+ "zerocopy",
  "zeroize",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,12 @@ mman = []
 ## [hermit-c]: https://github.com/hermit-os/hermit-c
 newlib = []
 
+## Enable stubbing result values for some system calls.
+##
+## For some system calls related to system options in which Hermit does not implement all possible options,
+## return 0 (success) instead of an error when the option is unknown.
+syscall-fake-success = []
+
 #! ### Platform Features
 
 ## Enables [Uhyve] support.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,8 +82,8 @@ newlib = []
 ## Enable stubbing result values for some system calls.
 ##
 ## For some system calls related to system options in which Hermit does not implement all possible options,
-## return 0 (success) instead of an error when the option is unknown.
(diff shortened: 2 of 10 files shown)


mkroening (changes requested):
Thanks, this is great!

> However, I have for now kept this flexibility in setsockopt. We may want to rewrite that to use `c_int` instead, for consistency? 

I think properly representing the pointer and the length is great.
Do you think it would make sense to represent it as a byte slice instead? We could leverage `zerocopy` in that case. Not sure how well it would tie in with strings, though.

Regarding the feature, same question as in the other PR: which software wants this with which values? How feasible would it be to support that "properly"? Returning fakes with finer granularity is fine of course.

zyuiop:
The three values (NODELAY, SNDBUF, RCVBUF) are used by the `socket2` rust library, and I think the particular application here was the Axum server.

Representing as a slice would indeed be cool, also because it reduces the amount of unsafety required. Not sure we could leverage zerocopy as in practice even booleans are encoded as i32 (and I assume zerocopy uses one byte to represent them?)

mkroening:
> The three values (NODELAY, SNDBUF, RCVBUF) are used by the `socket2` rust library, and I think the particular application here was the Axum server.

Huh, that's odd. We have an axum server that is tested in CI that does not need this. But that does currently patch `socket2` and `tokio` (https://github.com/hermit-os/hermit-rs/blob/main/Cargo.toml#L35-L37), since I would like to upstream it only after cleaning up the libc a bit.

Anyway, I think adding this support for non-bool values is great, I just would like to avoid the feature and stub values at a finer granularity (for each cmd explicitly, for example).

mkroening:
Could you rebase and address my last comment? :)

Or does this still depend on the other PR semantically?

zyuiop:
If we go for a `&[u8]`, should we use `'static` as lifetime?

mkroening:
I think `'static` would be wrong here. I think the lifetime should be tied to the C variable arguments in the syscall.

zyuiop:
I've tried something :)

mkroening (approved):
Okay, let's get this merged, then.

PR #2456, opened 2026-06-01
fix(socket): use smoltcp's Socket::recv_slice

zyuiop:
Following https://github.com/hermit-os/kernel/pull/2481, this completes the revert of https://github.com/hermit-os/kernel/pull/1606.

In addition to reintroducing the use of `Socket::recv_slice` for UDP, this also introduces it for TCP, which copies from the ringbuffer twice since it may wrap around. This should be strict improvements.

### DIFF
--- a/src/fd/socket/tcp.rs
+++ b/src/fd/socket/tcp.rs
@@ -203,15 +203,7 @@ impl ObjectInterface for Socket {
 					| tcp::State::TimeWait => Poll::Ready(Err(Errno::Io)),
 					_ => {
 						if socket.can_recv() {
-							Poll::Ready(
-								socket
-									.recv(|data| {
-										let len = core::cmp::min(buf.len(), data.len());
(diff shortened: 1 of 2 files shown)


mkroening (changes requested):
Thanks, this is great! Looking into why this was not done before led me to discover a soundness bug I recently introduced. That one will be fixed in https://github.com/hermit-os/kernel/pull/2481.

Once that one is merged, this one is good to go. We should also do it for UDP, though, which is also prepared in that PR.

Regarding performance: `recv_slice` calls [`RingBuffer::dequeue_slice`](https://docs.rs/smoltcp/0.13.1/smoltcp/storage/struct.RingBuffer.html#method.dequeue_slice) under the hood, which copies twice since we are reading from a ring buffer. This is a real improvement, though, I think, since we have a tighter hot loop here.

mkroening (approved):
Thanks! :)

PR #2457, opened 2026-06-01
feat/PCIe: verify the base address of PCI devices

zyuiop:
This ensures that a malicious hypervisor will not provide PCI devices that map to system memory

### DIFF
--- a/src/arch/x86_64/kernel/pci.rs
+++ b/src/arch/x86_64/kernel/pci.rs
@@ -6,6 +6,7 @@ use x86_64::instructions::port::Port;
 
 use crate::drivers::pci::{PCI_DEVICES, PciDevice};
 
+const PCI_MAX_BUS_NUMBER: u8 = 32;
 const PCI_CONFIG_ADDRESS_ENABLE: u32 = 1 << 31;
 
 const CONFIG_ADDRESS: Port<u32> = Port::new(0xcf8);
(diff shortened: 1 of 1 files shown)


src/arch/x86_64/kernel/pci.rs
```diff
@@ -180,6 +184,39 @@ mod pcie {
...
+				.map(|m| m.reg().unwrap().next().unwrap());
+
+			for mem_region in all_memory_regions {
+				let region_start = u64::try_from(mem_region.starting_address.addr()).unwrap();
+				let region_end = region_start + u64::try_from(mem_region.size.expect("found a memory region with no declared size")).unwrap();
+
+				if
+					(pci_start >= region_start && pci_start <= region_end) // PCI region starts within the memory region
+					|| (pci_end >= region_start && pci_end <= region_end) // PCI region ends within the memory region
+					|| (pci_start <= region_start && pci_end >= region_end) // PCI region contains the memory region
```
mkroening: Can't this be simplified?


```suggestion
					pci_start > region_end && pci_end < region_start
```

At least that is how I did it in [`PageRange::overlaps`](https://docs.rs/free-list/0.3.3/free_list/struct.PageRange.html#method.overlaps).

mkroening (changes requested):
Thanks, this looks good to me. :)

Just one suggestion.

src/arch/x86_64/kernel/pci.rs
zyuiop: Isn't it more `pci_start < region_end && pci_end > region_start` ?

I ended up using range intersections, which I think make the intent clearer

mkroening: You are right; I mistyped. Would you be okay with that? I'd like to avoid new nightly features if possible. :)

zyuiop: Okay, I was supposing this may be the case :)
Pushing a new version

mkroening (approved):
Thanks! :)

I'm squashing and merging. 👍

mkroening (approved):

PR #2462, opened 2026-06-03
feat: implement stack protection

zyuiop:
Isolate stacks in a separate virtual address space, just before the heap.
This ensures that stack overflows cause a page fault instead of silently corrupting the heap (for heap based stacks) or the page table (for the boot stack).

This has been tested only for x86_64 with UEFI, and will definitely not work on other platforms.

TODO:

- [x] Test/Fix on other platforms
- [x] Ensure we do this for all stacks (maybe it's not the case for IDT stacks after the first one)

### DIFF
--- a/src/arch/aarch64/kernel/scheduler.rs
+++ b/src/arch/aarch64/kernel/scheduler.rs
@@ -6,16 +6,16 @@ use core::sync::atomic::Ordering;
 use aarch64_cpu::asm::barrier::{SY, isb};
 use aarch64_cpu::registers::*;
 use align_address::Align;
-use free_list::{PageLayout, PageRange};
-use memory_addresses::{PhysAddr, VirtAddr};
+use memory_addresses::VirtAddr;
 
 use crate::arch::aarch64::kernel::CURRENT_STACK_ADDRESS;
 use crate::arch::aarch64::kernel::core_local::core_scheduler;
-use crate::arch::aarch64::mm::paging::{BasePageSize, PageSize, PageTableEntryFlags};
-use crate::mm::{FrameAlloc, PageAlloc, PageRangeAllocator};
+use crate::arch::aarch64::mm::paging::{BasePageSize, PageSize};
 use crate::scheduler::PerCoreSchedulerExt;
 use crate::scheduler::task::{Task, TaskFrame};
 use crate::{DEFAULT_STACK_SIZE, KERNEL_STACK_SIZE};
+use crate::arch::kernel::CURRENT_STACK;
+use crate::mm::stack_alloc::{allocate_stack, StackAllocation};
 
 #[derive(Debug)]
 #[repr(C, packed)]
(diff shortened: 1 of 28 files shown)


zyuiop:
Okay I have spent some time trying to port to arm/riscV

zyuiop:
I extracted the first two commits to #2466 so they can be merged independently :)

zyuiop:
Note than on ARM/RISC, the boot stack is not moved as we do in x86_64, so it is not protected.

I have a branch for RISCV on my PC, I can push it if there is interest.

zyuiop:
I can reproduce the `Laplace iterations` problem locally on `main` too, so I don't think this is my fault

mkroening (changes requested):
Could you rebase and squash any fixup commits? I believe some are already merged, though.

zyuiop:
my local rustfmt disagrees with the CI, how??
image

zyuiop:
I had to reduce the task stack size, otherwise the multi-processor/multi-task uhyve task was failing (4MB available for stacks was not enough when each stack takes 1MB).

In practice, user tasks spawned by the hermit-rs still use the old value, so this fix just frees-up the necessary 512KB for the multiple guard pages.

src/arch/x86_64/kernel/mod.rs
```diff
@@ -56,6 +59,59 @@ pub fn get_processor_count() -> u32 {
...
+}
+
+#[unsafe(naked)]
+unsafe extern "sysv64" fn clone_relocate_stack_raw(new_stack_top: usize, old_stack_top: usize) {
+	// Strategy: clone everything up to RSP, then set the new stack pointer (coming from RAX) and
+	// return
+	// This is dangerous
+	naked_asm!(
+		"mov rdx, rsp",
+		"call {clone_relocate_stack}",
```
stlankes: Do we have not to align the stack? 

I would add `sub rsp, 8` before` call {clone_relocate_stack}`

src/arch/x86_64/kernel/mod.rs
```diff
@@ -56,6 +59,59 @@ pub fn get_processor_count() -> u32 {
...
+		// Set stack pointer to new returned stack pointer, and hope
+		"mov rsp, rax",
+		"ret",
+
+		clone_relocate_stack = sym clone_relocate_stack,
+	)
+}
+
+unsafe fn clone_relocate_stack(
```
stlankes: Why do you use Rust-Default-ABI and not SysV (like clone_relocate_stack_raw)

src/arch/riscv64/start/smp.rs
```diff
@@ -2,11 +2,19 @@ use core::arch::naked_asm;
...
 
 #[unsafe(naked)]
 pub unsafe extern "C" fn smp_start(hart_id: usize) -> ! {
 	naked_asm!(
+		// Load page table if set
+		// Required in order to be able to change the stack pointer just after
+		"ld t0, {satp_value}",
+		"beqz t0, 1f",
+		"csrrw t0, satp, t0",
+
```
stlankes:
```suggestion
sfence.vma
```

src/arch/riscv64/start/hermit_entry.rs
```diff
@@ -32,6 +33,13 @@ pub unsafe extern "C" fn _start(hart_id: usize, boot_info: Option<&'static RawBo
 	}
 
 	naked_asm!(
+		// Load page table if set
+		// Required in order to be able to change the stack pointer just after
+		"ld t0, {satp_value}",
+		"beqz t0, 1f",
+		"csrrw t0, satp, t0",
+
```
stlankes:
```suggestion
sfence.vma
```

stlankes:
@zyuiop Can you resolve the conflicts.

src/arch/aarch64/kernel/scheduler.rs
```diff
@@ -94,156 +93,52 @@ pub(crate) struct State {
...
-		TaskStacks::Boot(BootStack { stack })
-	}
-
-	pub fn get_user_stack_size(&self) -> usize {
-		match self {
-			TaskStacks::Boot(_) => 0,
-			TaskStacks::Common(stacks) => stacks.total_size - DEFAULT_STACK_SIZE,
-		}
-	}
+		let kernel_stack = if let Some(stack) = CURRENT_STACK.lock().as_ref() {
```
stlankes: Ist it not possible that `finish_processor_init()`overwrite this value with a pointer to the next boot stack?

PR #2464, opened 2026-06-05
fix(virtio): do not fail on device configuration change if possible

cagatay-y:
Some of the device configuration change interrupts are only relevant if configuration values are cached by the driver. However, we do not make use of any such configurations. Thus, only panic if the interrupt is sent because the device needs reset, which we currently cannot handle and must not ignore.


mkroening (approved):
Thanks! :)

I have split this up and made some adjustments. I will merge if you don't have any objections: https://github.com/hermit-os/kernel/compare/cc6eee4de8b26b72519309a188c0deae654d6e4a..c2c0912a3612134b7c0240ec684f6fbc73f48ced

PR #2465, opened 2026-06-05
feat: device alloc: appropriate entire pages, lazily

zyuiop:
This is a partial rewrite of #1815.

When device memory is mapped at an offset, map it on demand instead of all at once at boot. To avoid doing a ton of TLB changes when VirtIO allocates memory (read: frequently), we keep the allocated pages in a dedicated list, and reuse them before doing any new frame allocation from the physical free list.

I have considered using a dedicated Talc allocator instead, but this was a bit too complicated (would have required a bunch of VirtIO code paths to do address translation, which we don't want).

### DIFF
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -2,11 +2,21 @@ use core::alloc::{AllocError, Allocator, Layout};
 use core::ptr::{self, NonNull};
 
 use align_address::Align;
+#[cfg(target_arch = "x86_64")]
+use free_list::FreeList;
 use free_list::{PageLayout, PageRange};
+#[cfg(target_arch = "x86_64")]
+use hermit_sync::InterruptTicketMutex;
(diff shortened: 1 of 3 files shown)


mkroening (changes requested):
Please put context and motivation into the PR description. This is for SEV, for which you treat the C-bit as an offset, right? 🤔 

I am wondering if there are any alternatives to this.

zyuiop:
Yes, this is required any time we want to appropriate memory in a way that requires a somewhat expensive operation (e.g. remapping it with the correct C bit).

We have discussed as well with @jounathaen about using a similar method to map the heap in a lazy fashion. 

I'm fine with setting that back to draft, or closing. I was interested in upstreaming because I already had to rebase this changeset a couple of times already. There may be better ways to achieve this, though.

zyuiop:
It turns out SEV does not need this either

PR #2466, opened 2026-06-05
fix(paging): memory unmapping for RISC-V and ARM

zyuiop:
When testing #2462 on ARM/Risc, I encountered page faults when trying to map pages.

Turns out that the previous unmapping was wrongly recorded.

This PR fixes that.


mkroening (approved):
Thanks! :)

Issue #2402, opened 2026-05-03
SMP boot code mapping got broken on Hermit


cagatay-y:
It seems like https://github.com/hermit-os/loader/commit/0c51a7266fa9dd0907629f0ebcca2fd54a5d1a0b broke SMP on Hermit. When I start Hermit with
```bash
  cargo xtask ci rs --arch x86_64 --profile dev  --package thread_test --smp 4 qemu --accel --sudo
```
it fails at https://github.com/hermit-os/kernel/blob/dd4c82a4e19499d2a07645a16c43c79723d39e16/src/arch/x86_64/kernel/apic.rs#L767-L776. I do not know if it is the loader that should not map the way it does or the kernel should not assume that the mapping does not exist / is 4 KiB-sized.

mkroening:
This is a kernel issue, not a loader issue. The kernel should handle this gracefully.

PR #2467, opened 2026-06-05
x86_64: support nested pages splitting

zyuiop:
Allows unmapping/mapping a page with a size lower than the actual mapping

### DIFF
--- a/src/arch/x86_64/mm/paging.rs
+++ b/src/arch/x86_64/mm/paging.rs
@@ -8,15 +8,15 @@ pub use x86_64::structures::idt::InterruptStackFrame as ExceptionStackFrame;
 use x86_64::structures::idt::PageFaultErrorCode;
 pub use x86_64::structures::paging::PageTableFlags as PageTableEntryFlags;
 use x86_64::structures::paging::frame::PhysFrameRange;
-use x86_64::structures::paging::mapper::{MapToError, MappedFrame, TranslateResult, UnmapError};
+use x86_64::structures::paging::mapper::{MapToError, MappedFrame, MapperFlush, TranslateResult, UnmapError};
 use x86_64::structures::paging::page::PageRange;
 use x86_64::structures::paging::{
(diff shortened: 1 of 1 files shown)


zyuiop:
Also fixes #2402

mkroening (changes requested):
Could you add a motivation to the PR description? I assume this is for SEV communication pages that need 4k pages, right?

Would it be possible to avoid splitting pages and use a separate page allocator instead for these 4k pages? That way, we can avoid making this split implicit.

zyuiop:
Yes, we could use a dedicated allocator ; but this is also useful because it makes it possible to call unmap/map for a page regardless of its size. Currently, unmapping à 4KiB page does not work if that page is part of a bigger page. If I remember correctly, this is required for the stack protection PR.

mkroening:
Yeah, that's the exact implicitness that I mean. I would personally be more comfortable if page sizes were handled explicitly by the mapper and page splitting could be avoided. If there's no way around this, it is fine, though.

Why does the stack protection PR need this? Can we not use fresh 4k pages there too?

zyuiop:
Because as a safety measure I remove any identity mapping for the physically addresses used for the stacks - ensuring nothing accidentally writes there using an old allocation

mkroening:
I would just leave the identity mapping. Do you think that is such a bad idea?

zyuiop:
I think it's better to have only one mapping per address, this way if for any reason/bug the identity address is used, it will cause a page fault instead of silently corrupting a stack somewhere.

It may be a bit overkill, but when I wrote this patch I was chasing after silent memory corruptions (that I was causing), so that's why

mkroening:
Okay then. Once you have rebased and put these reasons into the PR description or commit message, I will merge.

mkroening (approved):

mkroening:
I will discuss this next week, though. I am not very convinced that we should unmap identity-mapped memory. Linux does not do so either.

zyuiop:
@mkroening so actually my stack protection patch does not remove identity mapping ; this is only something I have in another branch. So we can keep this patch on the side in case it becomes needed again.

PR #2468, opened 2026-06-06
smp: always wake up core when ready

zyuiop:
This is a partial revert of commit f0bb0a0e.

This fixes a very tricky race condition that caused some cores to not be woken up, causing tests to fail (and weird bugs in the scheduling also)

# Context

In some PRs, `cargo xtask ci rs --arch x86_64 --profile dev  --package rusty_demo --features fs --smp 4 qemu --accel --sudo --devices virtio-fs-pci` fails due to timeout. 

This bug is reproducible locally, and is even more frequent with higher core counts (e.g., try it with `cargo xtask ci rs --arch x86_64 --profile dev  --package rusty_demo --features fs --smp 32 qemu --accel --sudo --devices virtio-fs-pci` on main and you're almost certain to get the bug. You can similarly confirm that this patch solves the bug.

The symptom is that the "Laplace" computation will get stuck. You can confirm this is not just the loop taking a long time by adding a small log in `compute` in `laplace.rs`

```rust
	for i in 0..iterations {
		iteration(current, next, size_x, size_y);
		mem::swap(&mut current, &mut next);
		println!("iteration {i}")
	}
```

You will observe the counter go up, then suddenly stop and nothing happens. This is the bug.

I initially believed this was a futex problem (so I rewrote them partially, another PR is incoming), but it was actually something different.

# The Bug (i think)

The thing that tipped me was that the bug was absent with the `idle-loop` flag. This greatly reduce the area that needed to be searched, since this flag is only present in a couple of places.

The bug is caused by a race condition when a core B wants to wake up a core A, which is going into halt.

The relevant call points are these:

- `PerCoreScheduler::run`. This is the idle task. It uses a backoff, checking periodically the core inbox to see if there is work. If there is none and the backoff is exhausted, it goes to sleep, calling `enable_and_wait`.
- `wakeup_core`. This is called by another CPU, after adding a task to the inbox. It sends an interrupt to the other core, to wake it up.

f0bb0a0e introduced a boolean for each core that indicates if the core is halting. If this boolean is set to `false`, `wakeup_core` skips the interrupt and returns immediately.
This works most of the time because `PerCoreScheduler::run` will always pick up tasks when some are available.

However, there is a possible race condition. Imagine the two following cores:

```
Core 1                                     Core 2

                                           run:
                                              check_inbox // inbox is empty
                                              enable_and_wait // go to enable interrupts

                                           enable_and_wait:
custom_wakeup:
  add_task_to_inbox
  wakeup_core

wakeup_core:
  check_hlt_state // hlt state is currently false
  return          // do nothing

                                              set_hlt_state(true)
                                              interrupts_enable
                                              hlt
```

I hope you appreciate my artistic talents, but that's the gist.

# The fix

Simply a partial revert of f0bb0a0e.
I was considering changing the boolean into a state machine, with a "don't sleep please" variant that could be sent by `wakeup_core` if the halt state was false, to indicate to a core that was about to go to sleep that it should not. 
I'm not sure it was working well because I was getting weird values in my `compare_and_exchange` calls, but this may be a viable alternative to reduce the number of interrupts we send.

# Notes

Increasing the core count increases the likelihood of the race condition happening. Laplace appears to be a good test-case because all threads are waiting on each other, so if ONE thread does not wake up, the program remains stuck.

It is possible that this patch will also solve other issues related to the scheduler suddenly becoming unresponsive until you cause an interrupt, but I have not tested that yet.

Fixes #2317

### DIFF
--- a/src/arch/x86_64/kernel/apic.rs
+++ b/src/arch/x86_64/kernel/apic.rs
@@ -886,10 +886,7 @@ pub fn ipi_tlb_flush() {
 #[allow(unused_variables)]
 pub fn wakeup_core(core_id_to_wakeup: CoreId) {
 	#[cfg(all(feature = "smp", not(feature = "idle-poll")))]
-	if core_id_to_wakeup != core_id()
-		&& !processor::supports_mwait()
-		&& scheduler::take_core_hlt_state(core_id_to_wakeup)
-	{
+	if core_id_to_wakeup != core_id() && !processor::supports_mwait() {
 		without_interrupts(|| {
 			let apic_ids = CPU_LOCAL_APIC_IDS.lock();
 			let local_apic_id = apic_ids[core_id_to_wakeup as usize];
--- a/src/arch/x86_64/kernel/core_local.rs
+++ b/src/arch/x86_64/kernel/core_local.rs
@@ -1,8 +1,6 @@
 use alloc::boxed::Box;
 use core::arch::asm;
 use core::cell::Cell;
-#[cfg(feature = "smp")]
-use core::sync::atomic::AtomicBool;
 use core::sync::atomic::Ordering;
 use core::{mem, ptr};
 
@@ -34,8 +32,6 @@ pub(crate) struct CoreLocal {
 	irq_statistics: &'static IrqStatistics,
 	/// The core-local async executor.
 	ex: StaticLocalExecutor<RawSpinMutex, RawRwSpinLock>,
-	#[cfg(feature = "smp")]
-	pub hlt: AtomicBool,
 	/// Queues to handle incoming requests from the other cores
 	#[cfg(feature = "smp")]
 	pub scheduler_input: InterruptTicketMutex<SchedulerInput>,
@@ -62,8 +58,6 @@ impl CoreLocal {
 			kernel_stack: Cell::new(ptr::null_mut()),
 			irq_statistics,
 			ex: StaticLocalExecutor::new(),
-			#[cfg(feature = "smp")]
-			hlt: AtomicBool::new(false),
 			#[cfg(feature = "smp")]
 			scheduler_input: InterruptTicketMutex::new(SchedulerInput::new()),
 		};
--- a/src/arch/x86_64/kernel/interrupts.rs
+++ b/src/arch/x86_64/kernel/interrupts.rs
@@ -82,8 +82,6 @@ pub(crate) fn enable_and_wait() {
 			);
 		}
 	} else {
-		#[cfg(feature = "smp")]
-		crate::CoreLocal::get().hlt.store(true, Ordering::Relaxed);
 		enable_and_hlt();
 	}
 }
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -8,8 +8,6 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::cell::RefCell;
 use core::ptr;
-#[cfg(all(target_arch = "x86_64", feature = "smp"))]
-use core::sync::atomic::AtomicBool;
 use core::sync::atomic::{AtomicI32, AtomicU32, Ordering};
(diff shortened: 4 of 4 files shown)


zyuiop:
I will note that after running a couple of times

```
cargo xtask ci rs --arch x86_64 --profile release  --package rusty_demo --features fs --features hermit/idle-poll --smp 4 qemu --accel
```

and 

```
cargo xtask ci rs --arch x86_64 --profile release  --package rusty_demo --features fs --smp 4 qemu --accel
```

`idle-poll` seems to outperform the non idle-poll version by a factor around 2x. Maybe this flag should become the default again?

mkroening (changes requested):
Thank you so much for finding the cause of that issue! :)

This PR does indeed solve it, but the slowdown feels too large for me to merge immediately.

On `main`, Laplace takes around 31 ms with and without idle-poll for me. With this PR, `idle-poll` stays within that range, which makes sense, but without `idle-poll`, Laplace takes more than 700 ms. This is in line with your performance comment.

I don't think we should default to `idle-poll`, since that wastes many resources and should be opt-in for high-performance scenarios.

I'll take a look at https://github.com/hermit-os/kernel/pull/2470 later. Maybe we can land that directly.

zyuiop:
That's weird. Locally I observed similar times with and without the patch, and certainly not a x30.
How many cores do you use? It looks like Laplace is only using a single core at a time anyway.

For this specific bench, idle-poll was faster than main, and my patch was basically identical

mkroening:
Okay, I have rerun the tests. I run `cargo xtask ci rs --arch x86_64 --package rusty_demo --release --smp 4 qemu --accel` and sometimes add `--feature hermit/idle-poll`.

In an Intel-based VM (Hermit is nested), I get the aforementioned slowdown (no poll-idle). On a bare-metal AMD EPYC 9354P machine, I only get a slowdown from about 20ms to about 30ms for Laplace (no poll-idle).

Laplace is using multiple threads through Rayon's `par_chunks_mut`.

zyuiop:
Thanks for the details.

My remark on "single thread at a time" was due to looking deeper into the futexes usage by rayon, and noticing that basically all threads are always waiting for the lock, so a single thread is used at a time - which you can also observe because iirc the single core time is faster than multiple cores.

On my side, I'm using an `AMD Ryzen AI 9 HX PRO 370 w/ Radeon 890M` (my laptop). Using `cargo xtask ci` to run 20 times laplace, in each config, I get:

- `main`: 14.4 ms average, 10.970 median. #wakeups (per core): 4/23/25/26
- this branch: 32.45 ms average, 31.441 median. #wakeups (per core): 12813/12781/12215/11863
- #2470: 10.2 ms average, 8.512 median, #wakeups (per core): 11/34/31/28

So it looks like #2470 completely fixes the issue AND brings the missing performance.

EDIT:

Here are the results with 8 cores.

- `main`: only 15 successful attempts. 886ms average, 775.791ms median
- this branch: 3858ms average, 3690ms median
- #2470: 608ms average, 523ms median.

Although I must say that I observe quite a lot of variability on this benchmark, so maybe it's not the best case to verify multicore performance.

mkroening:
https://github.com/hermit-os/kernel/pull/2470 brings the performance back down somewhat, but not better than before (before 30ms, this PR 700ms, that PR 38ms).

Still, correctness first, this bug is annoying. Let's merge this and try to gain the performance back eventually. Thanks again for this! :)

> My remark on "single thread at a time" was due to looking deeper into the futexes usage by rayon, and noticing that basically all threads are always waiting for the lock, so a single thread is used at a time - which you can also observe because iirc the single core time is faster than multiple cores.

I did not look at the futexes in detail yet, but this should scale with more cores. It does so on Linux in an expected way. On Hermit, it is weird. I opened https://github.com/hermit-os/kernel/issues/2480 to discuss this.

mkroening (approved):

PR #2469, opened 2026-06-06
feat: bucketized futexes

zyuiop:
While tracking #2468, I initially suspected the futex lock to be an issue, so I applied the "Todo" and made a bucket list instead of the single lock.

I have based this off on #2468 so that we get performance results that make sense.

### DIFF
--- a/src/synch/futex.rs
+++ b/src/synch/futex.rs
@@ -103,12 +103,19 @@ impl<const N: usize> BucketList<N> {
 	}
 
 	pub fn lock_bucket(&self, address: usize) -> SpinMutexGuard<'_, TaskListBucket> {
+		if N == 1 {
+			return self.0[0].lock()
+		}
 		let bucket = Self::hash_key(address);
 		self.0[bucket].lock()
 	}
 }
 
+#[cfg(feature = "smp")]
 static PARKING_LOT: BucketList<64> = BucketList::new();
+
+#[cfg(not(feature = "smp"))]
+static PARKING_LOT: BucketList<1> = BucketList::new();
 
 bitflags! {
 	pub struct Flags: u32 {
--- a/src/synch/futex.rs
+++ b/src/synch/futex.rs
@@ -1,13 +1,15 @@
+use alloc::collections::LinkedList;
 use alloc::collections::linked_list::CursorMut;
+use core::sync::atomic::AtomicU32;
+use core::sync::atomic::Ordering::SeqCst;
+
+use hermit_sync::{SpinMutex, SpinMutexGuard};
+
 use crate::arch::kernel::core_local::core_scheduler;
 use crate::arch::kernel::processor::get_timer_ticks;
 use crate::errno::Errno;
-use crate::scheduler::task::{TaskHandle, TaskHandlePriorityQueue};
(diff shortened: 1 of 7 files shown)


mkroening (commented):
The code looks okay to me. Unless it is too much effort, could you make the bucket map implementation generic? Did you measure any speedup in any stress test with this?

CC: @joboet for the original implementation.

zyuiop:
I can indeed make it generic, it's just that AFAIK there is no other usecase in the code for now for such a map. So this may be a case of "premature abstraction" :D 

But if you insist, I'll gladly do it :)

src/synch/futex.rs
```diff
@@ -1,20 +1,123 @@
...
+
+struct BucketList<const N: usize>([Bucket; N]);
+
+impl<const N: usize> BucketList<N> {
+	pub const fn new() -> Self {
+		Self([const { InterruptSpinMutex::new(TaskListBucket(LinkedList::new())) }; N])
+	}
+
+	fn hash_key(v: usize) -> usize {
+		let v = (v >> 3).to_be_bytes();
```
joboet, line 102: If you're trying to remove the zero bits resulting from `AtomicU32`'s alignment, then you have to shift by 2, not 3. Since you're hashing anyway, I don't think this is necessary anyway.

src/synch/futex.rs
```diff
@@ -1,20 +1,123 @@
...
 
-// TODO: Replace with a concurrent hashmap.
-static PARKING_LOT: InterruptTicketMutex<HashMap<usize, TaskHandlePriorityQueue, RandomState>> =
-	InterruptTicketMutex::new(HashMap::with_hasher(RandomState::with_seeds(0, 0, 0, 0)));
+struct BucketElem(usize, TaskHandlePriorityQueue);
+
+type Bucket = InterruptSpinMutex<TaskListBucket>;
+
+#[repr(transparent)]
+struct TaskListBucket(LinkedList<BucketElem>);
```
joboet, line 19: Using a linked-list here is a bit unfortunate – if there are a lot of tasks, the $O(n)$ lookup will start to matter. You could try using [`hashbrown::HashTable`](https://docs.rs/hashbrown/0.17.1/hashbrown/struct.HashTable.html) as an inner map to avoid having to recompute the hash while keeping hashmap-like lookup performance. Or alternatively, use a `BTreeMap` – that will also reduce memory usage as addresses become unused.

src/synch/futex.rs
zyuiop: I was trying to remove the last 3 bits from the address, which for some reason I believed to always be 0

But you are correct that this is in any case not needed since we hash. It may be the remain of an attempt to _not_ use a hash function at all for better performance. 

If it produced a more or less uniformly distributed distribution, `(addr >> 3) % N` would likely be better as it is less intensive than computing a hash.

Let me know what you think :)

src/synch/futex.rs
zyuiop: I think there is the same reasoning here... Wanting to avoid doing two hashes of the key. But I can switch to a BTreeMap.
And also it was likely easier to think in terms of iterators on a linked list than on a map

PR #2470, opened 2026-06-06
feat(smp): re-add an anti-spurious-wakeup feature

zyuiop:
As a follow-up to #2468, try to re-implement the logic I removed previously, but this time avoiding any potential race condition (hopefully??).

### DIFF
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -31,6 +31,8 @@ use crate::fd::{Fd, RawFd};
 use crate::io;
 use crate::scheduler::task::*;
 
+#[cfg(all(target_arch = "x86_64", feature = "smp", not(feature = "idle-poll")))]
+pub mod sleep_state;
 pub mod task;
 pub mod timer_interrupts;
 
@@ -890,6 +892,8 @@ pub(crate) fn add_current_core() {
 			core_id.try_into().unwrap(),
 			&CoreLocal::get().scheduler_input,
 		);
+		#[cfg(all(target_arch = "x86_64", not(feature = "idle-poll")))]
+		sleep_state::install_for_core(core_id);
 	}
 }
 
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -8,6 +8,8 @@ use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::cell::RefCell;
 use core::ptr;
+#[cfg(all(target_arch = "x86_64", feature = "smp", not(feature = "idle-poll")))]
+use core::sync::atomic::AtomicU8;
 use core::sync::atomic::{AtomicI32, AtomicU32, Ordering};
 
 use ahash::RandomState;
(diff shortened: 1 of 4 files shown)


src/scheduler/mod.rs
```diff
@@ -44,6 +46,129 @@ static WAITING_TASKS: InterruptTicketMutex<BTreeMap<TaskId, VecDeque<TaskHandle>
...
+	const STATUS_ACTIVE: u8 = 0;
+
+	/// The core is currently halted and can be interrupted for new tasks
+	const STATUS_IDLE: u8 = 1;
+
+	/// Another core tried to wake up this core but it was already active.
+	const STATUS_DONT_SLEEP: u8 = 2;
+
+	pub fn new() -> Self {
+		Self(AtomicU8::new(CoreState::STATUS_ACTIVE))
```
mkroening:
```suggestion
		Self(AtomicU8::new(Self::STATUS_ACTIVE))
```

src/scheduler/mod.rs
```diff
@@ -44,6 +46,129 @@ static WAITING_TASKS: InterruptTicketMutex<BTreeMap<TaskId, VecDeque<TaskHandle>
...
+			//    * AFTER: **We cannot be here!** Indeed, the atomic operation in `go_to_sleep`
+			//      also sets the state to STATUS_IDLE, so `previous_state` MUST BE STATUS_IDLE.
+			false
+		}
+	}
+}
+
+#[inline]
+#[cfg(all(target_arch = "x86_64", feature = "smp", not(feature = "idle-poll")))]
+pub fn core_sleep() -> bool {
```
mkroening: Maybe this would be clearer regarding the return value?

```suggestion
pub fn try_sleep() -> bool {
```

src/scheduler/mod.rs
```diff
@@ -44,6 +46,129 @@ static WAITING_TASKS: InterruptTicketMutex<BTreeMap<TaskId, VecDeque<TaskHandle>
...
+	CORE_HLT_STATE.read()[usize::try_from(core_id()).unwrap()].go_to_sleep()
+}
+
+#[inline]
+#[cfg(all(target_arch = "x86_64", feature = "smp", not(feature = "idle-poll")))]
+pub fn core_wake_up(core_id: CoreId) -> bool {
```
mkroening: Maybe this would be clearer regarding the return value?

```suggestion
pub fn try_wake_up(core_id: CoreId) -> bool {
```

mkroening (changes requested):
Thanks for this PR! :)

This does fix the laplace performance issue on my nested Intel VM (Hermit VM started with 4 cores):

- 9160d285a: ~38ms
- 30cb3f91c: ~700ms
- c88e43476: ~38ms

Regarding the code:
Could you move this into a `sleep_state.rs` file and call it `SleepState`? That way, the gates should be less scattered, and the name might be clearer.

Please also don't use `#[inline(always)]`. For reference, see [When to #[inline] - Standard library developers Guide](https://std-dev-guide.rust-lang.org/policy/inline.html). `#[inline]` is okay if the function is very small.

Regarding the atomics: relaxed ordering should be fine here, no? We don't use atomics here to build a synchronization primitive to synchronize memory access and are only interested in the values to make decisions.

mkroening (changes requested):
Looks good, just one question from last time remains unanswered:

> Regarding the atomics: relaxed ordering should be fine here, no? We don't use atomics here to build a synchronization primitive to synchronize memory access and are only interested in the values to make decisions.

zyuiop:
Re-checking the [nomicon](https://doc.rust-lang.org/nomicon/atomics.html) section, it looks like you're indeed correct, and `Relaxed` should work. Not sure this will bring a significant perf' advantage, but we can try :)

mkroening (approved):

PR #2473, opened 2026-06-09
refactor(interrupts): construct handler mappings during device init

cagatay-y:
MSI-X will require support for more than one interrupt vector per driver and more information from the driver side to decide how to handle the interrupts correctly. Revise how the mappings are constructed to make the construction more flexible and prepare for that use case.


src/arch/aarch64/kernel/interrupts.rs
```diff
@@ -87,16 +83,17 @@ pub fn disable() {
 	dmb(ISH);
 }
 
-pub(crate) fn install_handlers() {
+pub(crate) fn install_handlers(
+	old_handlers: HashMap<InterruptLine, InterruptHandlerQueue, RandomState>,
```
mkroening: Why is this called `old_handlers` here but not for the other arches?

src/arch/x86_64/kernel/interrupts.rs
```diff
@@ -161,8 +157,24 @@ pub(crate) fn install() {
...
 
-pub(crate) fn install_handlers() {
-	IRQ_HANDLERS.set(get_interrupt_handlers()).unwrap();
+pub(crate) fn install_handlers(
+	#[cfg_attr(not(feature = "pci"), expect(unused_mut))] mut handlers: HashMap<
+		InterruptLine,
+		InterruptHandlerQueue,
+		RandomState,
+	>,
+) {
```
mkroening: Can you put the attribute on the function instead? That formats a lot more nicely.

src/arch/x86_64/kernel/interrupts.rs
```diff
@@ -161,8 +157,24 @@ pub(crate) fn install() {
...
+	#[cfg(feature = "pci")]
+	{
+		use crate::kernel::serial;
+		add_irq_name(serial::SERIAL_IRQ, "COM1");
+		handlers
+			.entry(serial::SERIAL_IRQ)
+			.or_default()
+			.push_back(serial::handle_interrupt);
+	}
```
mkroening: Would it make sense to move this to the serial driver (module) instead and let it register there similar to the other drivers?

src/arch/aarch64/kernel/mmio.rs
```diff
@@ -81,7 +84,11 @@ impl MmioDriver {
 	}
 }
 
-#[cfg(any(feature = "virtio-console", feature = "virtio-fs", feature = "virtio-vsock"))]
+#[cfg(any(
+	feature = "virtio-console",
+	feature = "virtio-fs",
+	feature = "virtio-vsock"
+))]
```
mkroening, line 89: I can reproduce this, but I don't understand why VS code formatting changes this while `cargo fmt` from cmdline does not. I need to look into this.

src/arch/riscv64/kernel/devicetree.rs
```diff
@@ -99,7 +102,9 @@ pub fn init() {
 
 /// Inits drivers based on the device tree
 /// This function should only be called once
-pub fn init_drivers() {
+pub fn init_drivers(
+	#[allow(unused)] handlers: &mut HashMap<InterruptLine, InterruptHandlerQueue, RandomState>,
+) {
```
mkroening: I understand that we finer attributes are nice, but could you put this on the function too and perhaps also put a `cfg` and `expect`?

src/drivers/virtio/transport/mmio.rs
```diff
@@ -4,9 +4,10 @@
 //! For details on the Rust definitions, see [`virtio::mmio`].
 //!
 //! [Virtio Over MMIO]: https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-1650002
-
```
mkroening, line 7: Please restore this newline.

src/drivers/virtio/transport/pci.rs
```diff
@@ -631,6 +634,11 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
 /// list of the given device through [map_caps].
 pub(crate) fn init_device(
 	device: &PciDevice<PciConfigRegion>,
+	#[allow(unused_variables)] handlers: &mut HashMap<
```
mkroening: Please move to function, expect and cfg, please :)

src/drivers/mmio.rs
mkroening, line 1: It's very nice to get rid of the centralized handlers! :)

src/drivers/pci.rs
```diff
@@ -479,7 +401,13 @@ pub(crate) fn get_filesystem_driver() -> Option<&'static InterruptTicketMutex<Vi
 		.find_map(|drv| drv.get_filesystem_driver())
 }
 
-pub(crate) fn init() {
+pub(crate) fn init(
+	#[cfg_attr(
+		not(any(feature = "virtio", feature = "rtl8139")),
+		expect(unused_variables)
+	)]
```
mkroening: Dito.

mkroening (changes requested):
Could you expand a bit on the overall design and how it is necessitated by MSI-X?

Why have `fn init_dev(&mut self, handlers: &mut Map, irq: Option<Irq>)` instead of `fn interrupts(&self) -> impl Iterator<Item = Irq>`, for example?

src/arch/aarch64/kernel/interrupts.rs
cagatay-y: I didn't want to rename the locally declared `handlers` to keep the diff minimal and I cannot shadow shadow the argument because I need to have both accessible. I considered solving the shadowing issue by replacing the loop with a map and collect but hash_brown does not support `FromIterator`. If you think it makes more sense, I can rename the locally declared variable and rename the parameter to `handlers`.

mkroening: It's fine the way it is, then. 👍

cagatay-y:
> Why have `fn init_dev(&mut self, handlers: &mut Map, irq: Option<Irq>)` instead of `fn interrupts(&self) -> impl Iterator<Item = Irq>`, for example?

MSI-X can have different interrupt vectors for device configuration interrupts and different queues. An implementation that uses an iterator based approach would require the caller to have a lot of knowledge about the functioning of the device, for example which number is for which kind of queue, which queues we are interested in the interrupts of, etc. An iterator that returns vector-handler pairs could overcome this issue but giving the drivers access to the handler mapping allows the drivers to determine which spots are free and thus avoid vector collisions, reducing the latency for handling.

src/arch/x86_64/kernel/interrupts.rs
cagatay-y: `UartDevice` initialization is behind a lazy cell, so there is not an easy way we can pass `handlers` to the initializer without its restructuring.

mkroening: We could still move this code to a freestanding function in `serial.rs`, right? That way, we don't mix serial-specific code into this file.

mkroening (approved):
Thanks! :)

PR #2481, opened 2026-06-16
fix(soundness): initialize user-provided buffers with random data before writing into them

mkroening:
https://github.com/hermit-os/kernel/pull/1606 changed our read methods to take `&mut [MaybeUninit<u8>]` instead of `&mut [u8]`. This was done to avoid UB when applications pass uninitialized buffers to `read` and similar functions, either through C or through Rust's unstable `Read::read_buf`.

https://github.com/hermit-os/kernel/commit/445dac2d046572244a176345c4db9988004b51ed changed this back as part of https://github.com/hermit-os/kernel/pull/1891 to allow integration with the ecosystem's `Read::read` functions, that take a `&mut [u8]`. The soundness implications were not noticed.

This PR harmonizes Hermit's internal API by making `recv` also take a `&mut [u8]` instead of a `&mut [MaybeUninit<u8>]`.

Then, it makes the APIs sound when uninitialized buffers are passed by introducing a `fn init_buf(buf: &mut [MaybeUninit<u8>]) -> &mut [u8]` that is currently an assembler story no-op. This function might need to do actual work if Hermit ever gains support for actual uninitialized memory that might return different values when read without being written to. This is not the case yet, so the no-op is fine.

For details, see https://github.com/rust-lang/unsafe-code-guidelines/issues/527#issuecomment-4717256088.

### DIFF
--- a/src/init_buf.rs
+++ b/src/init_buf.rs
@@ -1,0 +1,45 @@
+use core::mem::MaybeUninit;
+
+/// Initializes a buffer.
+///
+/// This function initializes the provided buffer with arbitrary data. This is
+/// useful for passing user-allocated maybe-uninitialized memory to
+/// `Read::read` or similar functions.
+///
+/// No guarantee is being made about the values that the buffer is initialized
+/// with.
+///
+/// # Current implementation
+///
+/// This currently performs no actual instructions to initialize the buffer.
+/// Instead, it provides an assembly story that initializes the buffer with
+/// arbitrary values. This might change in the future, if Hermit gains support
+/// for memory allocation techniques. Those memory allocation techniques could
+/// make multiple reads from memory that has not been written to return
+/// different values. In that case, appropriate steps have to be taken here.
+///
+/// An example of such a technique on Linux is `MADV_FREE`, which would require
+/// writing at least once per page.
+///
+/// For details on assembly stories, see [How to use storytelling to fit inline
+/// assembly into Rust].
+///
+/// [How to use storytelling to fit inline assembly into Rust]: https://www.ralfj.de/blog/2026/03/13/inline-asm.html
+#[inline]
+pub fn init_buf(buf: &mut [MaybeUninit<u8>]) -> &mut [u8] {
+	// SAFETY: The story of this assembly block is that it writes arbitrary
+	// data into the buffer, initializing it. On Hermit, there is currently no
+	// way for never-written-to allocated memory contents to change. Thus,
+	// doing nothing is currently sufficient for writing arbitrary data.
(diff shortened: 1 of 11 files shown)


src/init_buf.rs
```diff
@@ -0,0 +1,46 @@
+use core::mem::MaybeUninit;
+use core::slice;
+
+/// Initializes a buffer.
+///
+/// This function initializes the provided buffer with random data. This is
```
CAD97:
```suggestion
/// This function initializes the provided buffer with arbitrary data. This is
```

I would suggest describing the resulting state as "arbitrary" instead of "random" here. The story that the assembly implements is that it fills the buffer with (*very* poor) random values, but because the values are not all that random, it would be better to describe the result to users of `init_buf` as being arbitrary, i.e. that no guarantees are provided as to what values are written, just that *some* value is written.

src/init_buf.rs
```diff
@@ -0,0 +1,46 @@
...
+/// [How to use storytelling to fit inline assembly into Rust]: https://www.ralfj.de/blog/2026/03/13/inline-asm.html
+#[inline]
+pub fn init_buf(buf: &mut [MaybeUninit<u8>]) -> &mut [u8] {
+	let ptr = buf.as_mut_ptr().cast::<u8>();
+	let len = buf.len();
+
+	// The story of this assembly block is that it writes random data into the
+	// buffer, initializing it.
+	//
+	// SAFETY: This assembly block contains no instructions.
```
CAD97: Technically, the safety for an assembly block is that it correctly implements the chosen story. (That is, as far as the Rust AM can ever possibly tell, it has picked random values and written them to all bytes in the provided array.) In combination with the documentation defined above, you *have* successfully claimed that, so this is a purely syntactical complaint that your SAFETY comment is implicitly relying on the reader to understand why performing no instructions is sufficient to write (horrific quality) random data.

src/init_buf.rs
mkroening: Thanks! :)

src/init_buf.rs
mkroening: Thanks! :)

mkroening:
@CAD97, thanks for your feedback!

I have tried to incorporate it: [`8286394..6335688`](https://github.com/hermit-os/kernel/compare/8286394d28ea74d26ad59300b0ad715edd13c666..63356888d66793738ce1fafe6a01bff613b22fd7)

CAD97 (approved):

mkroening:
Rebased on `main` to resolve conflicts.

PR #2488, opened 2026-06-18
fix(scheduler): drop the global TASKS entry on exit

stlankes:
The global TASKS BTreeMap got a new TaskHandle on every fork/spawn but was never cleaned up. This made join()'s
TASKS.contains_key(&id) always return true for any tid that ever existed, and slowly leaked TaskHandles. exit() already removes the WAITING_TASKS queue — remove the TASKS entry symmetrically.

### DIFF
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -194,6 +194,8 @@ impl PerCoreSchedulerExt for &mut PerCoreScheduler {
 					self.custom_wakeup(task);
 				}
 			}
+
+			TASKS.lock().remove(&current_id);
 		});
 
 		self.reschedule();


mkroening (approved):
Great! I tested with this program:

```rust
for i in 0usize.. {
    let thread = std::thread::spawn(move || {
        if i.is_multiple_of(0x1000) {
            println!("{i:#x}");
        }
    });
    thread.join().unwrap();
}
```

I ran on x86-64 with 32 MiB of RAM.

Before this PR, I get the following before `0x2_3000` threads:

```
[0][PANIC] panicked at /rustc/14210df0e27ccd7d9e6a05b8085cbd438e4bbc65/library/alloc/src/alloc.rs:573:9:
memory allocation of 384 bytes failed
```

With this PR, I don't get a crash even after `0x100_0000` threads (more than 100 times as many). I think we can be quite confident that we no longer have a memory leak on thread exit with this PR.

PR #2492, opened 2026-06-20
scheduler: add optional preemptive scheduling

stlankes:
The scheduler is tickless (a TSC-deadline one-shot timer), so a CPU-bound task that never blocks or yields keeps its core indefinitely. Add an optional `preemptive` feature that round-robins between ready tasks via a time slice.

A new `Source::Preemption` timer slot is armed for `now + 10ms` whenever the scheduler switches to a task while others are ready, and whenever `custom_wakeup` makes a task ready (so a freshly woken task preempts a non-yielding runner soon). When the slice expires, the timer handler's existing `reschedule()` switches to the next ready task. With no contention nothing is armed, so the system stays tickless.

### DIFF
--- a/src/scheduler/mod.rs
+++ b/src/scheduler/mod.rs
@@ -20,6 +20,8 @@ use timer_interrupts::TimerList;
 
 use crate::arch::kernel;
 use crate::arch::kernel::core_local::*;
+#[cfg(feature = "preemptive")]
+use crate::arch::kernel::processor;
 use crate::arch::kernel::scheduler::TaskStacks;
 #[cfg(target_arch = "riscv64")]
 use crate::arch::kernel::switch::switch_to_task;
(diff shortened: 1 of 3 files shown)


src/scheduler/mod.rs
```diff
@@ -398,6 +405,14 @@ impl PerCoreScheduler {
 		without_interrupts(|| {
 			let task = self.blocked_tasks.custom_wakeup(task);
 			self.ready_queue.push(task);
+			// The woken task is now ready. Arm the preemption timer so the running
+			// task — which may be CPU-bound and never yield — is preempted soon and
+			// the newcomer gets to run.
+			#[cfg(feature = "preemptive")]
+			{
+				let deadline = processor::get_timer_ticks() + PREEMPTION_SLICE_US;
+				timer_interrupts::create_timer_abs(timer_interrupts::Source::Preemption, deadline);
```
Gelbpunkt, line 414:
```suggestion
				timer_interrupts::create_timer(timer_interrupts::Source::Preemption, PREEMPTION_SLICE_US);
```

same for all the other calls. The function `create_timer` needs its `#[cfg(feature = "net")]` replaced with `#[cfg(any(feature = "net", feature = "preemptive"))]`

mkroening (changes requested):
Please address @Gelbpunkt's comment. Then, this is fine with me. 👍

stlankes:
Why is feature `net` required?

mkroening (approved):

PR #2493, opened 2026-06-22
ci: upgrade to Ubuntu 26.04

mkroening:


cagatay-y:
We should be able to uncomment these lines once this PR is merged
https://github.com/hermit-os/kernel/blob/5d52e43b0eb5a186d3f9bad0540b7347b8f9471d/.github/workflows/ci.yml#L295-L301

mkroening:
Yeah, that's one of the reasons I wanted to try this out. ^^

I am unsure whether we should land this now, though, since it is still a public preview only. 🤔

Gelbpunkt:
Ah, I was about to submit a PR for this and just realized that the same branch already existed on the repo and that's why the diff looked so weird.

This gets us to QEMU 10.2.1 up from 8.2.2, unfortunately not enough to enable semihosting on aarch64_be (which is fixed in QEMU 11+). I think we should land this now if the image works for us.

cagatay-y:
> unfortunately not enough to enable semihosting on aarch64_be (which is fixed in QEMU 11+).

I didn't look into the details but this may be relevant for the still too old QEMU version issue:
https://ubuntu.com/blog/ubuntu-virtualization-hwe-stack-confidential-computing

mkroening:
> I didn't look into the details but this may be relevant for the still too old QEMU version issue:
> https://ubuntu.com/blog/ubuntu-virtualization-hwe-stack-confidential-computing

Yeah, I am also looking forward to that. Right now, it does not bump the QEMU version, though: [qemu-system-x86-hwe](https://packages.ubuntu.com/resolute/qemu-system-x86-hwe)

I think we can expect it in October.

mkroening:
CI seems to be a bit slower with this. Let's hope it is due to external factors.

PR #2508, opened 2026-06-25
refactor(fd/socket/udp): Flatten deeply nested if chains by reordering

fogti:
I found this code segment very hard to reach, so I switched the `if` and `else` parts including conditions around in a few places.

I haven't measured the performance impact yet.

### DIFF
--- a/src/fd/socket/udp.rs
+++ b/src/fd/socket/udp.rs
@@ -157,25 +157,23 @@ impl ObjectInterface for Socket {
 	async fn recvfrom(&self, buf: &mut [u8]) -> io::Result<(usize, Endpoint)> {
 		future::poll_fn(|cx| {
 			self.with(|socket| {
-				if socket.is_open() {
-					if socket.can_recv() {
-						match socket.recv_slice(buf) {
-							Ok((len, meta)) => {
-								if self.remote_endpoint.is_none_or(|ep| ep == meta.endpoint) {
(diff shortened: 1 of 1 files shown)


mkroening:
Good idea. 👍 Do you still want to pursue this? :)

fogti:
Yes. Do we have any benchmarks one could mod for this? afaik the existing benchmarks (e.g. in uhyve) only use TCP…

mkroening:
This is only a trivial refactor, right? I don't expect any performance changes here. Or am I missing any changes that you expect to be difficult to optimize?

mkroening (approved):
Thanks! :)

I prefer early returns, so I adjusted the code accordingly. :)

src/fd/socket/udp.rs
```diff
@@ -54,21 +54,21 @@ impl Socket {
...
+				}
+
+				if !socket.can_send() {
+					socket.register_recv_waker(cx.waker());
+					return Poll::Pending;
 				}
+
+				Poll::Ready(
+					socket
+						.send_slice(buffer, *meta)
```
fogti, line 68: https://doc.rust-lang.org/std/result/enum.Result.html#method.map_or_else might be a good idea to use here...

fogti: okay, nope, not a good idea.

PR #2517, opened 2026-07-01
refactor(riscv): use volatile ref for interrupt controller accesses

cagatay-y:


Cargo.toml
```diff
@@ -348,7 +347,7 @@ take-static = "0.1"
 talc = { version = "5" }
 thiserror = { version = "2", default-features = false }
 time = { version = "0.3", default-features = false }
-volatile = "0.6"
+volatile = { version = "0.6", features = ["derive"] }
```
mkroening: Could you add this dependency with the required feature to `target.'cfg(target_arch = "riscv64")'.dependencies` instead? That way we can avoid the feature if it is not needed.

src/arch/riscv64/kernel/interrupts.rs
```diff
@@ -1,22 +1,112 @@
...
+	_reserved2: [u32; (CONTEXT_BASED_REGISTERS - 0x1f_2000) / size_of::<u32>()],
+	context_based_registers: [ContextBasedRegisters; NUMBER_OF_CONTEXTS],
+}
+
+const _: () = if offset_of!(Plic, interrupt_pending_bits) != INTERRUPT_PENDING_BITS_OFFSET
+	|| offset_of!(Plic, interrupt_enable_bits) != INTERRUPT_ENABLE_BITS_OFFSET
+	|| offset_of!(Plic, context_based_registers) != CONTEXT_BASED_REGISTERS
+{
+	panic!();
+};
```
mkroening: Could you split these into three separate items and use `assert!`?

```rust
const _: () = assert!(offset_of!(Plic, interrupt_pending_bits) == INTERRUPT_PENDING_BITS_OFFSET);
```

src/arch/riscv64/kernel/interrupts.rs
```diff
@@ -1,22 +1,112 @@
...
+
+impl Plic {
+	fn interrupt_priority<'a>(
+		ptr: VolatilePtr<'a, Self>,
+		source: NonZeroU16,
+	) -> VolatilePtr<'a, u32> {
```
mkroening: Could you make these methods taking `self` instead, by implementing an extension trait?

src/arch/riscv64/kernel/interrupts.rs
```diff
@@ -1,22 +1,112 @@
...
+				.update(|mut word| {
+					word.set_bit(source % 32, value);
+					word
+				});
+		};
+	}
+}
+
+static PLIC: SpinMutex<core::cell::OnceCell<VolatileRef<'static, Plic>>> =
+	SpinMutex::new(core::cell::OnceCell::new());
```
mkroening:
```suggestion
static PLIC: SpinMutex<OnceCell<VolatileRef<'static, Plic>>> =
	SpinMutex::new(OnceCell::new());
```

mkroening (changes requested):
Thanks! :)

src/arch/riscv64/kernel/interrupts.rs
cagatay-y: There is also `hermit_sync::OnceCell` in use. I thought it does not make sense to use that one for `PLIC`, since `hermit_sync::OnceCell` has a mutex and `PLIC` is already a mutex. `hermit_sync::OnceCell` makes sense for `PLIC_CONTEXT`, though, since it is not in a mutex itself. Is my reasoning correct? If it is, maybe I can make both use a qualifier so that the difference is more visible and the usage is more consistent.

mkroening: Ah, that's right. I overlooked the fact that this is a different `OnceCell`. Why do we need that, anyway? I'd suggest just using a `SpinMutex<VolatileRef>`.

cagatay-y: We discover the address of the PLIC after parsing the device tree. Should we use a dummy value instead?

mkroening (approved):
Thanks! :)

PR #2518, opened 2026-07-02
feat(x86_64): require Uhyve-reported CPU count

mkroening:


jounathaen (approved):

fogti:
What about the following remaining usage?
```
src/scheduler/mod.rs
28:use crate::arch::kernel::{get_processor_count, interrupts};
310:			if id == get_processor_count() {
913:		CORE_COUNTER.fetch_add(1, Ordering::SeqCst) % get_processor_count()
```

mkroening:
Those should be fine, I think. `get_processor_count()` does not have the best name. It returns the number of CPUs that are online. `get_possible_cpus()` returns the total number of CPUs, some of which may still be offline. The function is used to bring them online.

The old code that was removed in this PR had a race condition, which is why this transition was initiated in https://github.com/hermit-os/kernel/pull/509.

PR #2521, opened 2026-07-02
fix(virtio-net): discard packets whose checksums we cannot verify

cagatay-y:
Verifying the checksum of fragmented packets would require reassembling them, which we do not want to do in the kernel because of its complexity. IP fragmentation is "considered fragile" anyways (see RFC 8900).

### DIFF
--- a/src/drivers/net/virtio/mod.rs
+++ b/src/drivers/net/virtio/mod.rs
@@ -354,6 +354,13 @@ impl RxToken<'_> {
 				    return false;
 				}
 
+				if ip_packet.more_frags() || ip_packet.frag_offset() != 0 {
+					// The packet is part of a fragmented IP packet (RFC 791 page 8). We could check its checksum only
+					// if we reassembled it, so we discard it instead.
+					warn!("packet is fragmented, discarding!");
+					return false;
+				}
+
 				Self::is_ip_packet_passable(
 					ip_packet.next_header(),
 					ip_packet.payload(),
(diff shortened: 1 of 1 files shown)


mkroening (approved):
Thanks! :)

What would happen at the moment when such a packet is found? Will it fail at a later stage or corrupt data?

Should we open an issue to support Packetization Layer Path MTU Discovery (PLPMTUD), which is recommended by RFC 8900?

cagatay-y:
I forgot to mention that I triggered the issue by running `iperf -c <ip-address> --udp --length 3000` (or any other length larger than the MTU), which already causes iperf3 to emit `warning: UDP block size 3000 exceeds TCP MSS 1460, may result in fragmentation / drops`.

> What would happen at the moment when such a packet is found? Will it fail at a later stage or corrupt data?

The UDP header length field (or rather the bytes at that position for the later fragments) didn't match the payload size of the IP packet we received, so when we tried to construct a `UDPPacket` from the payload buffer, smoltcp failed at `UdpPacket::new_checked`.
 
> Should we open an issue to support Packetization Layer Path MTU Discovery (PLPMTUD), which is recommended by RFC 8900?

I don't know anything about PLPMTUD, so I can't really comment but I don't think it is necessary. The way I've discovered the error, the client would already send packets smaller than the MTU limit if I didn't force it to send more, so we do not have an issue of MTU discovery.

mkroening:
All right. 👍 Thanks for the context! :)

PR #2522, opened 2026-07-02
feat(virtio-net): declare support for TCP receive coalescing offload

cagatay-y:


mkroening (approved):
Thanks a lot! I will run a test on the BlueFields later before merging. :)

mkroening:
Yeah, this seems stable. 👍 

From host to guest via vhost-net I get 5.1 Gbit/s with this patch, up from 4.4 Gbit/s before.

No change on the BlueField. At least via DPDK, the card does not advertise guest TSO.

cagatay-y:
> No change on the BlueField. At least via DPDK, the card does not advertise guest TSO.

:/

The DPDK documentation [seems to suggest](https://doc.dpdk.org/guides/nics/mlx5.html#mlx5-lro) that it is supported, but requires some configuration that may be too much work for a quick test.

PR #2523, opened 2026-07-03
feat: Bochs VBE linear framebuffer support

GloriousAlpaca:
I'm currently working on porting Doom to Hermit and needed a feature to directly write pixel data into the Bochs Graphics Adaptor as Hermit currently only has support for VGA in textmode. 

The feature currently adds standard BGA preparation [(Link to OSDev)](https://wiki.osdev.org/Bochs_VBE_Extensions#Programming_the_BGA) with a hardcoded resolution (640x400), bpp (32) and a systemcall (sys_get_framebuffer) to receive the address of the framebuffer.
The systemcall falls back to 0 if there was a failure in initializing the framebuffer (e.g. if pci is missing, or the bga feature is not enabled).

I have currently only tested this feature with c programs on a mac using Qemu. The feature is also only available on the x86_64 architecture.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -216,6 +216,10 @@ virtio-vsock = ["virtio"]
 ## This is only useful on PCs (x86-64).
 vga = []
 
+## Enables the _Bochs Graphics Adapter_ (BGA) driver.
+##
+## This provides a linear framebuffer for pixel graphics and is primarily useful in QEMU/Bochs environments (x86-64).
+bga = ["pci"]
 #! ### Performance Features
 
 ## Disables putting the CPU to sleep.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -219,7 +219,9 @@ vga = []
 ## Enables the _Bochs Graphics Adapter_ (BGA) driver.
 ##
 ## This provides a linear framebuffer for pixel graphics and is primarily useful in QEMU/Bochs environments (x86-64).
+## It does not enable any kernel rendering, applications need to implement their own rendering on top of the framebuffer.
 bga = ["pci"]
+
 #! ### Performance Features
 
 ## Disables putting the CPU to sleep.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -219,7 +219,7 @@ vga = []
 ## Enables the _Bochs Graphics Adapter_ (BGA) driver.
 ##
 ## This provides a linear framebuffer for pixel graphics and is primarily useful in QEMU/Bochs environments (x86-64).
-## It does not enable any kernel rendering, applications need to implement their own rendering on top of the framebuffer.
+## Any rendering has to be done by the application; the kernel itself does not output anything on the framebuffer.
 bga = ["pci"]
 
 #! ### Performance Features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -239,6 +239,11 @@ vga = []
 ## It allows receiving scancodes from the PS/2 keyboard that can be used to port applications.
 ## This is only useful on PCs (x86-64).
 pc-keyboard = []
+## Enables the _Bochs Graphics Adapter_ (BGA) driver.
+##
+## This provides a linear framebuffer for pixel graphics and is primarily useful in QEMU/Bochs environments (x86-64).
+## Any rendering has to be done by the application; the kernel itself does not output anything on the framebuffer.
+bga = ["pci"]
 
 #! ### Performance Features
 
--- a/src/arch/x86_64/kernel/bga.rs
+++ b/src/arch/x86_64/kernel/bga.rs
@@ -39,7 +39,7 @@ pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
 		let bga_version = data_port.read();
 
 		if bga_version != VBE_DISPI_ID5 {
-			error!("Unsupported BGA version: {:#06x}", bga_version);
+			error!("Unsupported BGA version: {bga_version:#06x}");
 			return;
 		}
 
--- a/src/arch/x86_64/kernel/bga.rs
+++ b/src/arch/x86_64/kernel/bga.rs
@@ -1,0 +1,91 @@
+use core::sync::atomic::{AtomicU64, Ordering};
+
+use memory_addresses::{PhysAddr, VirtAddr};
+use pci_types::{Bar, CommandRegister};
+use x86_64::instructions::port::Port;
+
+use crate::arch::x86_64::mm::paging;
+use crate::arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags, PageTableEntryFlagsExt};
+use crate::drivers::pci::PciDevice;
+use crate::kernel::pci::PciConfigRegion;
+
+static FRAMEBUFFER_PHYS: AtomicU64 = AtomicU64::new(0);
+
+const VBE_DISPI_IOPORT_INDEX: u16 = 0x01ce;
+const VBE_DISPI_IOPORT_DATA: u16 = 0x01cf;
+
+const VBE_DISPI_INDEX_ID: u16 = 0;
+const VBE_DISPI_INDEX_XRES: u16 = 1;
+const VBE_DISPI_INDEX_YRES: u16 = 2;
+const VBE_DISPI_INDEX_BPP: u16 = 3;
+const VBE_DISPI_INDEX_ENABLE: u16 = 4;
+
+const VBE_DISPI_DISABLED: u16 = 0x00;
+const VBE_DISPI_ENABLED: u16 = 0x01;
+const VBE_DISPI_LFB_ENABLED: u16 = 0x40;
+const VBE_DISPI_ID5: u16 = 0xb0c5;
+
+pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
+	//To Do: Detect Resolution automatically
+	let width: u16 = 640;
+	let height: u16 = 400;
+	let bpp: u16 = 32;
+
+	unsafe {
+		let mut index_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_INDEX);
+		let mut data_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_DATA);
+
+		index_port.write(VBE_DISPI_INDEX_ID);
+		let bga_version = data_port.read();
+
+		if bga_version != VBE_DISPI_ID5 {
+			error!("Unsupported BGA version: {:#06x}", bga_version);
+			return;
+		}
+
+		index_port.write(VBE_DISPI_INDEX_ENABLE);
+		data_port.write(VBE_DISPI_DISABLED);
+
+		index_port.write(VBE_DISPI_INDEX_XRES);
+		data_port.write(width);
+
+		index_port.write(VBE_DISPI_INDEX_YRES);
+		data_port.write(height);
+
+		index_port.write(VBE_DISPI_INDEX_BPP);
+		data_port.write(bpp);
+
+		index_port.write(VBE_DISPI_INDEX_ENABLE);
+		data_port.write(VBE_DISPI_ENABLED | VBE_DISPI_LFB_ENABLED);
+	}
+
+	adapter.set_command(CommandRegister::MEMORY_ENABLE);
+
+	let (phys_addr, size) = match adapter.get_bar(0) {
+		Some(Bar::Memory32 { address, size, .. }) => (u64::from(address), size as usize),
+		Some(Bar::Memory64 { address, size, .. }) => (address, size as usize),
+		_ => return,
+	};
+
+	FRAMEBUFFER_PHYS.store(phys_addr, Ordering::Release);
+
+	assert!(
+		size % 4096 == 0,
+		"Framebuffer size must be a multiple of 4096 bytes"
+	);
+	let page_count = size / 4096;
+
+	let mut flags = PageTableEntryFlags::empty();
+	flags.device().writable().execute_disable();
+	flags.insert(PageTableEntryFlags::USER_ACCESSIBLE);
+	paging::map::<BasePageSize>(
+		VirtAddr::new(phys_addr),
+		PhysAddr::new(phys_addr),
+		page_count,
+		flags,
+	);
+}
+
+pub fn get_framebuffer_address() -> u64 {
+	FRAMEBUFFER_PHYS.load(Ordering::Acquire)
+}
--- a/src/arch/x86_64/kernel/bga.rs
+++ b/src/arch/x86_64/kernel/bga.rs
@@ -1,63 +1,131 @@
-use core::sync::atomic::{AtomicU64, Ordering};
+//! This module contains the implementation of the Bochs Graphics Adapter (BGA) driver.
+//!
+//! The driver uses the Bochs VBE Extensions, which use two I/O ports to communicate with the
+//! emulated VGA card instead of relying on a 16-bit VBE BIOS. The driver initializes the BGA
+//! device, sets the desired resolution and bits per pixel (BPP), and maps the framebuffer
+//! into the virtual address space. It also provides a function to retrieve the physical
+//! address of the framebuffer.
 
+use hermit_sync::OnceCell;
 use memory_addresses::{PhysAddr, VirtAddr};
 use pci_types::{Bar, CommandRegister};
-use x86_64::instructions::port::Port;
+use x86_64::instructions::port::{Port, PortWriteOnly};
 
-use crate::arch::x86_64::mm::paging;
-use crate::arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags, PageTableEntryFlagsExt};
+use crate::arch::x86_64::mm::paging::{
+	self, BasePageSize, PageTableEntryFlags, PageTableEntryFlagsExt,
+};
 use crate::drivers::pci::PciDevice;
 use crate::kernel::pci::PciConfigRegion;
 
-static FRAMEBUFFER_PHYS: AtomicU64 = AtomicU64::new(0);
+pub struct BgaInfo {
+	pub framebuffer: *mut u8,
+	pub width: u16,
+	pub height: u16,
+	pub bpp: u16,
+}
+
+static BGA_INFO: OnceCell<BgaInfo> = OnceCell::new();
+
+unsafe impl Send for BgaInfo {}
+unsafe impl Sync for BgaInfo {}
 
 const VBE_DISPI_IOPORT_INDEX: u16 = 0x01ce;
 const VBE_DISPI_IOPORT_DATA: u16 = 0x01cf;
 
-const VBE_DISPI_INDEX_ID: u16 = 0;
-const VBE_DISPI_INDEX_XRES: u16 = 1;
-const VBE_DISPI_INDEX_YRES: u16 = 2;
-const VBE_DISPI_INDEX_BPP: u16 = 3;
-const VBE_DISPI_INDEX_ENABLE: u16 = 4;
+pub struct VbeDispiIndex;
+
+#[allow(dead_code)]
+impl VbeDispiIndex {
+	#[doc(alias = "VBE_DISPI_INDEX_ID")]
+	pub const ID: u16 = 0;
+	#[doc(alias = "VBE_DISPI_INDEX_XRES")]
+	pub const XRES: u16 = 1;
+	#[doc(alias = "VBE_DISPI_INDEX_YRES")]
+	pub const YRES: u16 = 2;
+	#[doc(alias = "VBE_DISPI_INDEX_BPP")]
+	pub const BPP: u16 = 3;
+	#[doc(alias = "VBE_DISPI_INDEX_ENABLE")]
+	pub const ENABLE: u16 = 4;
+	#[doc(alias = "VBE_DISPI_INDEX_BANK")]
+	pub const BANK: u16 = 5;
+	#[doc(alias = "VBE_DISPI_INDEX_VIRT_WIDTH")]
+	pub const VIRT_WIDTH: u16 = 6;
+	#[doc(alias = "VBE_DISPI_INDEX_VIRT_HEIGHT")]
+	pub const VIRT_HEIGHT: u16 = 7;
+	#[doc(alias = "VBE_DISPI_INDEX_X_OFFSET")]
+	pub const X_OFFSET: u16 = 8;
+	#[doc(alias = "VBE_DISPI_INDEX_Y_OFFSET")]
+	pub const Y_OFFSET: u16 = 9;
+}
 
 const VBE_DISPI_DISABLED: u16 = 0x00;
 const VBE_DISPI_ENABLED: u16 = 0x01;
 const VBE_DISPI_LFB_ENABLED: u16 = 0x40;
-const VBE_DISPI_ID5: u16 = 0xb0c5;
+
+#[allow(dead_code)]
+const VBE_DISPI_NOCLEARMEM: u16 = 0x80;
+
+pub struct VbeDispiId;
+
+#[allow(dead_code)]
+impl VbeDispiId {
+	#[doc(alias = "VBE_DISPI_ID0")]
+	pub const ID0: u16 = 0xb0c0;
+	#[doc(alias = "VBE_DISPI_ID1")]
+	pub const ID1: u16 = 0xb0c1;
+	#[doc(alias = "VBE_DISPI_ID2")]
+	pub const ID2: u16 = 0xb0c2;
+	#[doc(alias = "VBE_DISPI_ID3")]
+	pub const ID3: u16 = 0xb0c3;
+	#[doc(alias = "VBE_DISPI_ID4")]
+	pub const ID4: u16 = 0xb0c4;
+	#[doc(alias = "VBE_DISPI_ID5")]
+	pub const ID5: u16 = 0xb0c5;
+}
+
+struct BgaRegisters;
+
+impl BgaRegisters {
+	pub fn read(index: u16) -> u16 {
+		let mut index_port: PortWriteOnly<u16> = PortWriteOnly::new(VBE_DISPI_IOPORT_INDEX);
+		let mut data_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_DATA);
+		unsafe {
+			index_port.write(index);
+			data_port.read()
+		}
+	}
+
+	pub fn write(index: u16, value: u16) {
+		let mut index_port: PortWriteOnly<u16> = PortWriteOnly::new(VBE_DISPI_IOPORT_INDEX);
+		let mut data_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_DATA);
+		unsafe {
+			index_port.write(index);
+			data_port.write(value);
+		}
+	}
+}
 
 pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
-	//To Do: Detect Resolution automatically
+	//To Do: Add support for different resolutions and BPP values
 	let width: u16 = 640;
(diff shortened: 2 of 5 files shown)


src/syscalls/system.rs
```diff
@@ -6,3 +6,18 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
+pub extern "C" fn sys_get_framebuffer() -> u64 {
+	crate::kernel::bga::get_framebuffer_address()
+}
+
+#[cfg(not(all(target_arch = "x86_64", feature = "bga")))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub extern "C" fn sys_get_framebuffer() -> u64 {
+	0
+}
```
mkroening: We usually just don't create the symbol if the feature is not enabled until we find an existing application that is okay with a stub. Does this apply in your case? If not, please remove this.

src/syscalls/system.rs
```diff
@@ -6,3 +6,18 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
 	BasePageSize::SIZE.try_into().unwrap()
 }
+
+/// Returns the address of the framebuffer, if available. Returns 0 if no framebuffer is available.
+#[cfg(all(target_arch = "x86_64", feature = "bga"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub extern "C" fn sys_get_framebuffer() -> u64 {
+	crate::kernel::bga::get_framebuffer_address()
+}
```
mkroening: The design of system calls needs special scrutiny and discussion. 

First, if the user is meant to write into the frame buffer through the return value of this system call, the return type must be a pointer type, not just an integer address.

Second, I am not that deep into graphics, but is the size and dimensions not something that is essential when drawing into a framebuffer? Is that currently just hardcoded? My gut feeling would be to propose something as follows:

```rust
pub extern "C" fn sys_framebuffer(framebuffer: *mut Framebuffer) -> c_int;

#[repr(C)]
pub struct Framebuffer {
    height: u32,
    width: u32,
    size: u64,
}
```

This is just an initial thought, though. It would be useful to look into related existing APIs such as [libdrm](https://gitlab.freedesktop.org/mesa/libdrm/-/tree/libdrm-2.4.134/man) and UEFI's [Graphics Output Protocol](https://uefi.org/specs/UEFI/2.11/12_Protocols_Console_Support.html#graphics-output-protocol) ([`uefi::proto::console::gop::GraphicsOutput`](https://docs.rs/uefi/0.38.0/uefi/proto/console/gop/struct.GraphicsOutput.html)).

Another consideration might be multi-monitor support or more properties that a framebuffer has. Of course, the design of this PR does not need to be the last API ever on Hermit. We can replace it if necessary. Still, it would be nice to at least discuss some degree of future-proofing.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
+use core::sync::atomic::{AtomicU64, Ordering};
+
+use memory_addresses::{PhysAddr, VirtAddr};
+use pci_types::{Bar, CommandRegister};
+use x86_64::instructions::port::Port;
+
+use crate::arch::x86_64::mm::paging;
+use crate::arch::x86_64::mm::paging::{BasePageSize, PageTableEntryFlags, PageTableEntryFlagsExt};
```
mkroening: Nit:

```suggestion
use crate::arch::x86_64::mm::paging::{self, BasePageSize, PageTableEntryFlags, PageTableEntryFlagsExt};
```

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::drivers::pci::PciDevice;
+use crate::kernel::pci::PciConfigRegion;
+
+static FRAMEBUFFER_PHYS: AtomicU64 = AtomicU64::new(0);
```
mkroening: Could you make this an atomic pointer to preserve provenance?


```suggestion
static FRAMEBUFFER: AtomicPtr<u8> = AtomicPtr::new(ptr::null_mut());
```

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+const VBE_DISPI_ID5: u16 = 0xb0c5;
+
+pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
+	//To Do: Detect Resolution automatically
+	let width: u16 = 640;
+	let height: u16 = 400;
+	let bpp: u16 = 32;
+
+	unsafe {
+		let mut index_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_INDEX);
```
mkroening: This is not meant to be read, no?

You could make this a write-only port, then.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+static FRAMEBUFFER_PHYS: AtomicU64 = AtomicU64::new(0);
+
+const VBE_DISPI_IOPORT_INDEX: u16 = 0x01ce;
+const VBE_DISPI_IOPORT_DATA: u16 = 0x01cf;
+
+const VBE_DISPI_INDEX_ID: u16 = 0;
+const VBE_DISPI_INDEX_XRES: u16 = 1;
+const VBE_DISPI_INDEX_YRES: u16 = 2;
+const VBE_DISPI_INDEX_BPP: u16 = 3;
+const VBE_DISPI_INDEX_ENABLE: u16 = 4;
```
mkroening: This looks like a candidate for enums. Do you think an enum would make sense here?

Could you add the remaining indexes? If necessary, use `#[allow(dead_code)]` on the struct or `#[expect(dead_code)]` on the enum variants. You could still keep it searchable by adding `#[doc(alias = "ORIGINAL_NAME")]` to the variants.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+
+const VBE_DISPI_DISABLED: u16 = 0x00;
+const VBE_DISPI_ENABLED: u16 = 0x01;
+const VBE_DISPI_LFB_ENABLED: u16 = 0x40;
```
mkroening, line 60: The [vgabios docs](https://github.com/qemu/vgabios/blob/19ea12c230ded95928ecaef0db47a82231c2e485/vbe_display_api.txt#L119) seem to call this differently. Where is your name from? 🤔 

```suggestion
const VBE_DISPI_VBE_ENABLED: u16 = 0x40;
```

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+const VBE_DISPI_ID5: u16 = 0xb0c5;
```
mkroening: Could you add the missing IDs as well as `VBE_DISPI_NOCLEARMEM` for completeness?

Do you think the IDs would be a candidate for grouping in an enum?

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+
+pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
+	//To Do: Detect Resolution automatically
```
mkroening: Do you want to do this in this PR or in another, or is this meant for other people in the future?

src/arch/x86_64/kernel/bga.rs
mkroening: If we need to store more info such as dimensions and size, a `OnceCell` might make sense, instead.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+		index_port.write(VBE_DISPI_INDEX_ID);
+		let bga_version = data_port.read();
+
+		if bga_version != VBE_DISPI_ID5 {
+			error!("Unsupported BGA version: {bga_version:#06x}");
+			return;
+		}
+
+		index_port.write(VBE_DISPI_INDEX_ENABLE);
+		data_port.write(VBE_DISPI_DISABLED);
```
mkroening: We should have some abstractions here. What about something like this?

```rust
struct BgaRegisters;

impl BgaRegisters {
    pub fn read(index: u16) -> u16;
    pub fn write(index: u16, value: u16);
}
```

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+	let (phys_addr, size) = match adapter.get_bar(0) {
+		Some(Bar::Memory32 { address, size, .. }) => (u64::from(address), size as usize),
+		Some(Bar::Memory64 { address, size, .. }) => (address, size as usize),
+		_ => return,
+	};
+
+	FRAMEBUFFER_PHYS.store(phys_addr, Ordering::Release);
+
+	assert!(
+		size % 4096 == 0,
```
mkroening: I am wondering why [`clippy::manual_is_multiple_of`](https://rust-lang.github.io/rust-clippy/stable/index.html#manual_is_multiple_of) is not triggering here.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+		"Framebuffer size must be a multiple of 4096 bytes"
+	);
+	let page_count = size / 4096;
+
+	let mut flags = PageTableEntryFlags::empty();
+	flags.device().writable().execute_disable();
+	flags.insert(PageTableEntryFlags::USER_ACCESSIBLE);
```
mkroening: Why do we need this? The application is still running in kernel mode, no?

mkroening (dismissed):
Thanks for the PR! :)

This is great! First I was wondering why we should use BGA instead of UEFI's GOP, which would also work on real machines, but I guess BGA is just easier at the moment. Since it is not that invasive, this is fine with me.

I am a bit confused regarding naming; maybe you could clear that up in a doc comment of the module or even the feature documentation. If I understand correctly, BGA is the Bochs Graphics Adapter, which we talk to via the VESA BIOS Extensions (VBE) API. But instead of talking to the BIOS via interrupts, we can just talk to the VBE API through the special BGA ports, right? In that case, the module names, feature names, and constant names make sense to me now.

Cargo.toml
```diff
@@ -216,6 +216,10 @@ virtio-vsock = ["virtio"]
 ## This is only useful on PCs (x86-64).
 vga = []
 
+## Enables the _Bochs Graphics Adapter_ (BGA) driver.
+##
+## This provides a linear framebuffer for pixel graphics and is primarily useful in QEMU/Bochs environments (x86-64).
+bga = ["pci"]
```
mkroening, line 246: This should also mention that this does not enable any kernel rendering on the device and that applications must be ported to use this framebuffer.

mkroening:
Ah, please also ensure that your PR does not include merge commits. Please rebase instead.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,91 @@
...
+const VBE_DISPI_LFB_ENABLED: u16 = 0x40;
+const VBE_DISPI_ID5: u16 = 0xb0c5;
+
+pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
+	//To Do: Detect Resolution automatically
+	let width: u16 = 640;
+	let height: u16 = 400;
+	let bpp: u16 = 32;
+
+	unsafe {
```
mkroening: Please try to have as few unsafe operations per block as possible. Ideally, only one.

src/syscalls/system.rs
GloriousAlpaca: Initially the driver threw no real errors, but I guess I was a bit overzealous while implementing fail states. I will remove the stub!

src/syscalls/system.rs
GloriousAlpaca: Yes this sounds good, I'd maybe setup the struct in a similar way to UEFI [Under EFI_GRAPHICS_OUTPUT_MODE_INFORMATION](https://uefi.org/specs/UEFI/2.11/12_Protocols_Console_Support.html#efi-graphics-output-protocol)

```rust
#[repr(C)]
pub struct FramebufferInfo {
pub address: *mut u8,
pub width: u32,
pub height: u32,
pub bpp: u32,
}
```

I'd then change the syscall to take a pointer with which to fill the data and return either 0 or an error code (e.g. if no hardware is present):
```rust
sys_get_framebuffer(info: *mut FramebufferInfo, struct_size: usize) -> i32
```
(Struct Size so we can expand the Syscall later, while legacy code still works as expected)

I don't think BGA has multi-monitor support itself, the driver would probably have to look different for this and it would probably make more sense to do virtio-gpu instead. Virtio could then just expand the same struct downwards (for example with pitch or bitmasks).

src/arch/x86_64/kernel/bga.rs
GloriousAlpaca: I actually took that name out of the docs to enable the linear framebuffer [On OSDev](https://wiki.osdev.org/Bochs_VBE_Extensions#Using_a_linear_frame_buffer_(LFB)):

_When using a linear framebuffer, the BGA exposes all of the video memory in a single linearly addressable section of memory. The address of the framebuffer is not fixed, and must be read from the first [PCI](https://wiki.osdev.org/PCI) base address register (BAR 0 of device 0x1234:0x1111). To enable the linear framebuffer, use the VBE_DISPI_LFB_ENABLED flag (0x40) when enabling the BGA in conjunction with the VBE_DISPI_ENABLED (0x01) flag._

I might be misunderstanding something though.

src/arch/x86_64/kernel/bga.rs
GloriousAlpaca: An enum seems a bit inconvenient for this as we have to cast the elements to a u16 everytime we use them with the ports.
What would you think of a struct in this form:
```rust
pub struct VbeDispiIndex;

impl VbeDispiIndex {
	#[doc(alias = "VBE_DISPI_INDEX_ID")]
	pub const ID: u16 = 0;
	#[doc(alias = "VBE_DISPI_INDEX_XRES")]
	pub const XRES: u16 = 1;
	#[doc(alias = "VBE_DISPI_INDEX_YRES")]
	pub const YRES: u16 = 2;
	#[doc(alias = "VBE_DISPI_INDEX_BPP")]
	pub const BPP: u16 = 3;
	#[doc(alias = "VBE_DISPI_INDEX_ENABLE")]
	pub const ENABLE: u16 = 4;
	#[doc(alias = "VBE_DISPI_INDEX_BANK")]
	pub const BANK: u16 = 5;
	#[doc(alias = "VBE_DISPI_INDEX_VIRT_WIDTH")]
	pub const VIRT_WIDTH: u16 = 6;
	#[doc(alias = "VBE_DISPI_INDEX_VIRT_HEIGHT")]
	pub const VIRT_HEIGHT: u16 = 7;
	#[doc(alias = "VBE_DISPI_INDEX_X_OFFSET")]
	pub const X_OFFSET: u16 = 8;
	#[doc(alias = "VBE_DISPI_INDEX_Y_OFFSET")]
	pub const Y_OFFSET: u16 = 9;
}
```

GloriousAlpaca:
Thanks a lot for the feedback! I have pushed a commit, where I've addressed the feedback you've given. I have actually implemented the systemcall a bit differently than I've written before. Please do have a look if this is a good way to design it. I have also used structs instead of enums to consolidate the constants. Let me know if this is alright with you.

src/syscalls/system.rs
```diff
@@ -1,8 +1,46 @@
...
 pub extern "C" fn sys_getpagesize() -> i32 {
 	BasePageSize::SIZE.try_into().unwrap()
 }
+
+/// Returns the framebuffer information for the BGA device, if it has been initialized. Returns 0
+/// on success, or -1 if the BGA device has not been initialized.
+#[cfg(all(target_arch = "x86_64", feature = "bga"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_framebuffer(info: *mut FramebufferInfo) -> c_int {
```
jounathaen:
```suggestion
pub unsafe extern "C" fn sys_get_framebuffer_info(info: *mut FramebufferInfo) -> c_int {
```

src/syscalls/system.rs
```diff
@@ -1,8 +1,46 @@
...
+	match bga_info {
+		Some(bga_info) => {
+			let info_c = FramebufferInfo {
+				framebuffer: bga_info.framebuffer,
+				width: u32::from(bga_info.width),
+				height: u32::from(bga_info.height),
+				bpp: u32::from(bga_info.bpp),
+			};
+			unsafe {
+				info.write(info_c);
```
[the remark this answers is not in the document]
jounathaen: That's AI Slop.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+	}
+}
+
+pub fn init_device(adapter: &PciDevice<PciConfigRegion>) {
+	//To Do: Add support for different resolutions and BPP values
+	let width: u16 = 640;
+	let height: u16 = 400;
+	let bpp: u16 = 32;
+
+	let bga_version = BgaRegisters::read(VbeDispiIndex::ID);
```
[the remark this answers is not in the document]
jounathaen: AI Slop

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+	BgaRegisters::write(VbeDispiIndex::YRES, height);
+	BgaRegisters::write(VbeDispiIndex::BPP, bpp);
+	BgaRegisters::write(
+		VbeDispiIndex::ENABLE,
+		VBE_DISPI_ENABLED | VBE_DISPI_LFB_ENABLED,
+	);
+
+	adapter.set_command(CommandRegister::MEMORY_ENABLE);
+
+	let (phys_addr, size) = match adapter.get_bar(0) {
```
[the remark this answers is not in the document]
jounathaen: Lol.
AI Slop

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+		Some(Bar::Memory32 { address, size, .. }) => (u64::from(address), size as usize),
+		Some(Bar::Memory64 { address, size, .. }) => (address, size as usize),
+		_ => return,
+	};
+
+	BGA_INFO
```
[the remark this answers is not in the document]
jounathaen: This is actually a kind of valid point, but the solution here is to `unwrap()` the set, to capture cases where the initialization is executed twice.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+		.set(BgaInfo {
+			framebuffer: core::ptr::with_exposed_provenance_mut(phys_addr as usize),
```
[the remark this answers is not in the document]
jounathaen: AI Slop

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+			width,
+			height,
+			bpp,
+		})
+		.ok();
+
+	assert!(
```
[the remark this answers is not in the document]
jounathaen: AI Slop

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+		size.is_multiple_of(4096),
+		"Framebuffer size must be a multiple of 4096 bytes"
+	);
+	let page_count = size / 4096;
+
+	let mut flags = PageTableEntryFlags::empty();
```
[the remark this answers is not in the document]
jounathaen: AI Slop

src/syscalls/system.rs
```diff
@@ -1,8 +1,46 @@
...
+use core::ffi::c_int;
+
 use crate::arch::mm::paging::{BasePageSize, PageSize};
+#[repr(C)]
+pub struct FramebufferInfo {
+	pub framebuffer: *mut u8,
+	pub width: u32,
+	pub height: u32,
+	pub bpp: u32,
+}
```
jounathaen: This should be config gated for `bga` and I'd suggest moving it down, close to the actual framebuffer fn.

src/arch/x86_64/kernel/bga.rs
jounathaen: I find the LFB term also in the vgabios library: https://github.com/qemu/vgabios/blob/19ea12c230ded95928ecaef0db47a82231c2e485/vbe.h#L310

Looks to me like the `VBE_ENABLED` variant in that library is either a mistake or a relic from the older VBE mode.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+	pub framebuffer: *mut u8,
+	pub width: u16,
+	pub height: u16,
+	pub bpp: u16,
+}
+
+static BGA_INFO: OnceCell<BgaInfo> = OnceCell::new();
+
+unsafe impl Send for BgaInfo {}
+unsafe impl Sync for BgaInfo {}
```
jounathaen: Do we really need these impls? I guess `Send` is ok, since the framebuffer is not thread bound, but the struct isn't `Sync` in my view.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+	pub const BANK: u16 = 5;
+	#[doc(alias = "VBE_DISPI_INDEX_VIRT_WIDTH")]
+	pub const VIRT_WIDTH: u16 = 6;
+	#[doc(alias = "VBE_DISPI_INDEX_VIRT_HEIGHT")]
+	pub const VIRT_HEIGHT: u16 = 7;
+	#[doc(alias = "VBE_DISPI_INDEX_X_OFFSET")]
+	pub const X_OFFSET: u16 = 8;
+	#[doc(alias = "VBE_DISPI_INDEX_Y_OFFSET")]
+	pub const Y_OFFSET: u16 = 9;
+}
```
jounathaen: I think, an enum might be a good alternative here.

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,165 @@
...
+	#[doc(alias = "VBE_DISPI_ID4")]
+	pub const ID4: u16 = 0xb0c4;
+	#[doc(alias = "VBE_DISPI_ID5")]
+	pub const ID5: u16 = 0xb0c5;
+}
+
+struct BgaRegisters;
+
+impl BgaRegisters {
+	pub fn read(index: u16) -> u16 {
```
jounathaen: When you use an enum for the index, as suggested above, you can make the interface here more robust by only allowing the enum's values. This automatically makes the safety requirements of the port I/O write/read trivial to fulfill.

Same for the `write` fn.

Cargo.toml
jounathaen: I find "enable any kernel rendering" a little bit too unspecific. Maybe something like:

"Any rendering has to be done by the application; the kernel itself does not output anything on the framebuffer."

src/arch/x86_64/kernel/bga.rs
```diff
@@ -1,64 +1,132 @@
...
+	pub const ID1: u16 = 0xb0c1;
+	#[doc(alias = "VBE_DISPI_ID2")]
+	pub const ID2: u16 = 0xb0c2;
+	#[doc(alias = "VBE_DISPI_ID3")]
+	pub const ID3: u16 = 0xb0c3;
+	#[doc(alias = "VBE_DISPI_ID4")]
+	pub const ID4: u16 = 0xb0c4;
+	#[doc(alias = "VBE_DISPI_ID5")]
+	pub const ID5: u16 = 0xb0c5;
+}
```
sarahspberrypi: This could also be a candidate for an enum.

src/arch/x86_64/kernel/bga.rs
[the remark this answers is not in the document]
sarahspberrypi: Actually, quite the opposite is true, QEMU uses/offers the latest version. Bad bot!

src/arch/x86_64/kernel/bga.rs
GloriousAlpaca: I've gotten rid of this unclean way to circumvent a compiler error and have just changed the type of the address to usize. That way we can just get the pointer with exposed provenance in the systemcall.

jounathaen: We have a "PhysAddr" type (or something called similar), which would be an even better fit.

src/syscalls/system.rs
```diff
@@ -6,3 +9,50 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
+
+/// Returns the framebuffer information for the BGA device, if it has been initialized. Returns 0
+/// on success, or -1 if the BGA device has not been initialized.
+#[cfg(all(target_arch = "x86_64", feature = "bga"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_get_bga_info(info: *mut FramebufferInfo) -> c_int {
```
jounathaen:
```suggestion
/// Returns the framebuffer information for the video output device.
/// Returns 0 on success, or -1 if the BGA device has not yet been initialized.
#[cfg(all(target_arch = "x86_64", feature = "bga"))]
#[hermit_macro::system]
#[unsafe(no_mangle)]
pub unsafe extern "C" fn sys_get_framebuffer_info(info: *mut FramebufferInfo) -> c_int {
```

src/syscalls/system.rs
```diff
@@ -6,3 +9,50 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
+			0
+		}
+		None => -1,
+	}
+}
+
+#[cfg(all(target_arch = "x86_64", feature = "bga"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_set_bga_resolution(width: u16, height: u16, bpp: u16) -> c_int {
```
jounathaen:
```suggestion
pub unsafe extern "C" fn sys_set_resolution(width: u16, height: u16, bpp: u16) -> c_int {
```

src/arch/x86_64/kernel/mod.rs
```diff
@@ -13,6 +13,8 @@ use crate::arch::kernel::core_local::*;
 #[cfg(feature = "acpi")]
 pub mod acpi;
 pub mod apic;
+#[cfg(feature = "bga")]
```
jounathaen:
```suggestion
#[cfg(all(target_arch = "x86_64", feature = "bga"))]
```

src/arch/x86_64/kernel/bga.rs
```diff
@@ -0,0 +1,170 @@
...
+
+	pub fn write(index: VbeDispiIndex, value: u16) {
+		let mut index_port: PortWriteOnly<u16> = PortWriteOnly::new(VBE_DISPI_IOPORT_INDEX);
+		let mut data_port: Port<u16> = Port::new(VBE_DISPI_IOPORT_DATA);
+		unsafe {
+			index_port.write(index as u16);
+			data_port.write(value);
+		}
+	}
+}
```
jounathaen, line 109: This is not thread safe, as two parallel invocations could cause a race.

Instead: Make these functions a member function of `BgaRegisters` that takes a `&mut` reference. The struct itself should be put into a `OnceCell` with a Mutex.

Then you can add a corresponding safety comment as well 😉

GloriousAlpaca: I've implemented the feedback. (I have amended these changes to the main commit)

PR #2524, opened 2026-07-03
fix(aarch64/interrupt): match against the correct exception class

cagatay-y:
A mix-up seems to have happened in 341a1bf when replacing hardcoded literals with the enum variants.


Gelbpunkt (approved):
Thanks for fixing this, I'm still not sure how this came to be, I might've been staring at bits for too long when writing this :)

PR #2525, opened 2026-07-03
build(deps): patch trapframe dependency

mkroening:
This PR patches the trapframe dependency to include https://github.com/rcore-os/trapframe-rs/pull/18, which should fix the linker issues I saw in https://github.com/hermit-os/kernel/pull/2425 (CI only, very fragile).

In the long term, we should probably migrate away from this dependency (https://github.com/hermit-os/kernel/issues/2526).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1912,8 +1912,7 @@ checksum = "8d2d250f87fb3fb6f225c907cf54381509f47b40b74b1d1f12d2dccbc915bdfe"
 [[package]]
 name = "trapframe"
 version = "0.10.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


sbutz:
I can reproduce the issue locally. The proposed patch in trapframe-rs fixes the issue.

mkroening:
> I can reproduce the issue locally. The proposed patch in trapframe-rs fixes the issue.

Great! Since no PRs are currently blocked on this, let's wait a bit for a release of https://github.com/rcore-os/trapframe-rs/pull/19. If we get one soon, we can avoid the git patch.

mkroening:
trapframe-rs was upgraded properly in https://github.com/hermit-os/kernel/pull/2536.

PR #2528, opened 2026-07-06
feat(aarch64): Seed entropy using ARM RNG

sbutz:
Implement missing entropy seeding on ARM.
[ArmRng Implentation](https://github.com/rust-embedded/aarch64-cpu/blob/ee779cbf9e082f026037623e5ba372d4da053ca6/packages/aarch64-cpu/src/asm/random.rs#L18)
[ARM Doc](https://developer.arm.com/documentation/111107/2026-03/AArch64-Registers/RNDR--Random-Number)

For RISC-V we could leverage the optional Zkr extension.
[RISC-V Doc](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#zkr)
Changes follow in separate PR.

### DIFF
--- a/src/arch/aarch64/kernel/processor.rs
+++ b/src/arch/aarch64/kernel/processor.rs
@@ -1,6 +1,7 @@
 use core::arch::asm;
 use core::{fmt, mem};
 
+use aarch64_cpu::asm::random::ArmRng;
 use aarch64_cpu::registers::*;
 use hermit_sync::{Lazy, OnceCell, without_interrupts};
 
@@ -197,7 +198,13 @@ impl fmt::Display for CpuFrequency {
 }
 
 pub fn seed_entropy() -> Option<[u8; 32]> {
-	None
+	let rng = ArmRng::new()?;
+	let mut buf = [0u8; 32];
+	for word in buf.chunks_mut(8) {
+		let value = rng.rndr()?.to_ne_bytes();
+		word.copy_from_slice(&value);
+	}
+	Some(buf)
 }
 
 /// The halt function stops the processor until the next interrupt arrives


src/arch/aarch64/kernel/processor.rs
```diff
@@ -196,8 +197,16 @@ impl fmt::Display for CpuFrequency {
 	}
 }
 
+static RNG: Lazy<Option<ArmRng>> = Lazy::new(ArmRng::new);
+
 pub fn seed_entropy() -> Option<[u8; 32]> {
-	None
+	let rng = RNG.as_ref()?;
```
mkroening: Since this carries no data and I suspect that the register access to check its availability is cheap, we should not need the static.

```suggestion
	let rng = ArmRng::new()?;
```

src/arch/aarch64/kernel/processor.rs
```diff
@@ -196,8 +197,16 @@ impl fmt::Display for CpuFrequency {
...
+	let mut buf = [0u8; 32];
+	for word in buf.chunks_mut(8) {
+		let value = rng.rndr().or(None).unwrap().to_ne_bytes();
```
mkroening: The `or(None)` does nothing.

```suggestion
		let value = rng.rndr().unwrap().to_ne_bytes();
```

mkroening: I am unsure about the unwrap. Would it not be better to return `None` via `?` instead?

mkroening (changes requested):
Thanks, this is great! :)

mkroening:
I have opened https://github.com/hermit-os/kernel/pull/2544 to enable the CPU feature in CI.

src/arch/aarch64/kernel/processor.rs
sbutz: That was my intention. Still new to rust.

sbutz: Fixed

src/arch/aarch64/kernel/processor.rs
sbutz: Done

mkroening (approved):
Thanks, this is great! Also, thanks for your patience.

I have rebased to have CI run this with `-cpu max,lpa2=off` (https://github.com/hermit-os/kernel/pull/2544).

Edit: I'll rebase once more after https://github.com/hermit-os/kernel/pull/2545 is merged.

PR #2531, opened 2026-07-06
experiment: run more CI jobs on ARM

mkroening:


mkroening:
This does not make too much sense due to some prebuilt binaries only being available for x86. 🤔

PR #2532, opened 2026-07-06
feat: add PS/2 keyboard interrupt driver

GloriousAlpaca:
This feature adds support for the PS2 legacy keyboard in Qemu x86_64.
Currently it does:
1. A minimized hardware initialization (e.g. flush the output buffer) to use the keyboard, according to the standards for the Intel 8042 chip: [Link to OSDev](https://wiki.osdev.org/I8042_PS/2_Controller#Initialising_the_PS/2_Controller)
2. Install an interrupt handler, which is done in the install_handlers function in interrupt.rs, as it is a hardware interrupt but does not belong in the pci interrupt handlers list (legacy device).
3. Add a systemcall to pop the last scancode out of the keyboard input ringbuffer (which currently has a length of 256). The Ringbuffer also uses Atomics, to prevent races and deadlocks.
The systemcall returns 0 if the keyboard feature is disabled.

I have only tested this feature with C programs on a Mac using Qemu.


Cargo.toml
```diff
@@ -42,6 +42,7 @@ default = [
 	"dhcpv4",
 	"fsgsbase",
 	"kernel-stack",
+    "keyboard",
```
mkroening: I don't think this should be a default feature.

Cargo.toml
```diff
@@ -216,6 +217,11 @@ virtio-vsock = ["virtio"]
 ## This is only useful on PCs (x86-64).
 vga = []
 
+## Enables the PS/2 keyboard driver.
+##
+## This is only useful on PCs (x86-64).
+keyboard = []
```
mkroening: Could you rename this to `pc-keyboard`? Also, please make sure that the docs don't imply that this is needed for keyboard support in general. Serial-based keyboard support is separate from this. Similar to the framebuffer, please also put a disclaimer that this does not make the kernel use the driver and instead exposes an API for applications that need to be ported.

src/syscalls/system.rs
```diff
@@ -6,3 +6,17 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
+pub extern "C" fn sys_read_keyboard() -> u8 {
+	crate::kernel::keyboard::pop_scancode().unwrap_or(0)
+}
+
+#[cfg(not(all(target_arch = "x86_64", feature = "keyboard")))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub extern "C" fn sys_read_keyboard() -> u8 {
+	0
+}
```
mkroening: Similar to the framebuffer PR, please either merge the function definitions or (preferrably) just remove the symbol when the feature is disabled.

src/arch/x86_64/kernel/interrupts.rs
```diff
@@ -157,7 +157,14 @@ pub(crate) fn install() {
 	IRQ_NAMES.lock().insert(7, "FPU");
 }
 
-pub(crate) fn install_handlers(handlers: InterruptHandlerMap) {
+pub(crate) fn install_handlers(#[allow(unused_mut)] mut handlers: InterruptHandlerMap) {
```
mkroening: Please move the allow to the function level instead of having it inline like this.

src/arch/x86_64/kernel/interrupts.rs
```diff
@@ -157,7 +157,14 @@ pub(crate) fn install() {
...
+	#[cfg(feature = "keyboard")]
+	{
+		use crate::arch::kernel::keyboard::get_keyboard_handler;
```
mkroening:
```suggestion
		use crate::arch::kernel::keyboard::get_keyboard_handler;

```

src/arch/x86_64/kernel/pc_keyboard.rs
mkroening, line 1: Would the [pc-keyboard](https://docs.rs/pc-keyboard) crate help here in any way? I'd like to avoid reimplementing logic if the ecosystem already has a well-established crate for (parts of) this.

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+const PS2_CMD_WRITE_CNFG: u8 = 0x60;
+const PS2_CMD_DISABLE_KEYBOARD: u8 = 0xad;
+const PS2_CMD_DISABLE_MOUSE: u8 = 0xa7;
+const PS2_CMD_ENABLE_KEYBOARD: u8 = 0xae;
+const PS2_CNFG_ENABLE_KEYBOARD_INTERRUPT: u8 = 0x01;
+const PS2_BUFFER_FULL: u8 = 0x01;
+
+static KEYBOARD_BUFFER: [AtomicU8; BUFFER_SIZE] = [ATOMIC_ZERO; BUFFER_SIZE];
+static WRITE_INDEX: AtomicUsize = AtomicUsize::new(0);
+static READ_INDEX: AtomicUsize = AtomicUsize::new(0);
```
mkroening: While atomics prevent data races and deadlocks here, I am pretty sure we still have undesirable race conditions or possible data loss like this.

What about having a proper struct behind a mutex for this? That way, we can also use `VecDeque` instead of having to reimplement one ourselves. You can take a look at the serial input for reference:

https://github.com/hermit-os/kernel/blob/274120e1a84cdad6f0efa5ea0c541bd1bc64ae96/src/arch/x86_64/kernel/serial.rs#L15-L21

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub(crate) fn get_keyboard_handler() -> (u8, fn()) {
+	unsafe {
```
mkroening: Please try to have as few unsafe operations per block as possible. Ideally, only one.

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+		cmd_port.write(PS2_CMD_READ_CNFG);
+		let mut config = data_port.read();
+
+		config |= PS2_CNFG_ENABLE_KEYBOARD_INTERRUPT;
+
+		cmd_port.write(PS2_CMD_WRITE_CNFG);
+		data_port.write(config);
+		cmd_port.write(PS2_CMD_ENABLE_KEYBOARD);
+	}
+	fn keyboard_handler() {
```
mkroening: Please move the handler to the top level outside of this function.

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+pub(crate) fn get_keyboard_handler() -> (u8, fn()) {
+	unsafe {
+		let mut cmd_port = Port::<u8>::new(PS2_CMD_PORT);
+		let mut data_port = Port::<u8>::new(PS2_DATA_PORT);
+		cmd_port.write(PS2_CMD_DISABLE_KEYBOARD);
+		cmd_port.write(PS2_CMD_DISABLE_MOUSE);
+
+		while (cmd_port.read() & PS2_BUFFER_FULL) != 0 {
+			let _ = data_port.read();
+		}
```
mkroening: Why do we discard the device buffer instead of filling our read buffer?

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+const PS2_BUFFER_FULL: u8 = 0x01;
+
+static KEYBOARD_BUFFER: [AtomicU8; BUFFER_SIZE] = [ATOMIC_ZERO; BUFFER_SIZE];
+static WRITE_INDEX: AtomicUsize = AtomicUsize::new(0);
+static READ_INDEX: AtomicUsize = AtomicUsize::new(0);
+
+pub(crate) fn get_keyboard_handler() -> (u8, fn()) {
+	unsafe {
+		let mut cmd_port = Port::<u8>::new(PS2_CMD_PORT);
+		let mut data_port = Port::<u8>::new(PS2_DATA_PORT);
```
mkroening: Do you think an abstraction similar to the proposal for BGA would make sense?

```rust
struct Ps2;

impl Ps2 {
    fn read_data() -> u8;
    fn write_data(data: u8);
    fn status() -> u8;
    fn command(command: u8);
}
```

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+
+	interrupts::add_irq_name(1, "PS/2 Keyboard");
+
+	(1, keyboard_handler)
+}
+
+/// Pops a scancode from the keyboard buffer, returning None if the buffer is empty.
+pub fn pop_scancode() -> Option<u8> {
```
mkroening: A scancode can never be zero, right? Returning `Option<NonZero<u8>>` would be preferable in that case.

src/syscalls/system.rs
```diff
@@ -6,3 +6,17 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
 pub extern "C" fn sys_getpagesize() -> i32 {
 	BasePageSize::SIZE.try_into().unwrap()
 }
+
+#[cfg(all(target_arch = "x86_64", feature = "keyboard"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub extern "C" fn sys_read_keyboard() -> u8 {
+	crate::kernel::keyboard::pop_scancode().unwrap_or(0)
+}
```
mkroening: I am not too sure about this API. Is the application supposed to busy loop on this and retrieve one event at a time?

What about doing something similar to Linux's event device (evdev) interface ([Linux docs](https://www.kernel.org/doc/html/latest/input/input.html))? Reading from `/dev/input/event0` would then fill a user buffer with input events and blocks if no events are there unless opened with `O_NONBLOCK`.

mkroening (dismissed):
Thanks for the PR! :)

I was wondering why implement drivers for legacy devices instead of USB keyboards ([xhci](https://docs.rs/xhci), [usb-oxide](https://github.com/RJR777/hermit-localLLM/blob/main/usb-oxide/src/hid.rs), [embassy-usb](https://github.com/embassy-rs/embassy/tree/main/embassy-usb)). I guess it is because of simplicity.

It would be great to discuss the high-level application-facing API, since that is the hardest to change once merged.

Cargo.toml
GloriousAlpaca: Yes, this accidentally slipped through sorry!

src/arch/x86_64/kernel/pc_keyboard.rs
GloriousAlpaca: I took a look at the crate and the following stood out to me:

There are three basic steps to handling keyboard input. Your application may bypass some of these.

- Ps2Decoder - converts 11-bit PS/2 words into bytes, removing the start/stop bits and checking the parity bits. Only needed if you talk to the PS/2 keyboard over GPIO pins and not required if you talk to the i8042 PC keyboard controller.
- ScancodeSet - converts from Scancode Set 1 (i8042 PC keyboard controller) or Scancode Set 2 (raw PS/2 keyboard output) into a symbolic KeyCode and an up/down KeyState.
- EventDecoder - converts symbolic KeyCode and KeyState into a Unicode characters (where possible) according to the currently selected KeyboardLayout.

We actually don't need the first step because we are using the i8042 keyboard controller. 
The other two steps look promising, but I actually would not put them into the kernel driver itself, but rather into the application.
I mainly implemented this driver to use with the doom port I'm currently working on, where I have to translate the keys anyways, which makes it irrelevant if it's scancodes or keycodes. If we pre-translate in the kernel we would also have to handle different keyboard layouts, which I feel would overcomplicate this pretty simple driver, especially because Qemu might emulate a different layout than the host. It also seems to me like something the application itself should handle instead of the driver. What do you think?

src/arch/x86_64/kernel/pc_keyboard.rs
GloriousAlpaca: At the start there might be garbage data in the output buffer, that can get stuck if we don't clear it.

https://wiki.osdev.org/I8042_PS/2_Controller#Step_4:_Flush_The_Output_Buffer

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,85 @@
...
+const PS2_CMD_WRITE_CNFG: u8 = 0x60;
+const PS2_CMD_DISABLE_KEYBOARD: u8 = 0xad;
+const PS2_CMD_DISABLE_MOUSE: u8 = 0xa7;
+const PS2_CMD_ENABLE_KEYBOARD: u8 = 0xae;
+const PS2_CNFG_ENABLE_KEYBOARD_INTERRUPT: u8 = 0x01;
+const PS2_BUFFER_FULL: u8 = 0x01;
+
+const BUFFER_SIZE: usize = 256;
+struct Ps2;
+
```
jounathaen:
```suggestion

struct Ps2;
```

Small nitpick

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Ps2 {
+	pub fn read_status() -> u8 {
+		let mut status_port = Port::<u8>::new(PS2_CMD_PORT);
+		unsafe { status_port.read() }
```
jounathaen:
```suggestion
		unsafe { Port::<u8>::new(PS2_CMD_PORT).read() }
```

src/syscalls/system.rs
GloriousAlpaca: Since we don't use a filesystem like Linux I would propose something like this:
`pub extern "C" fn sys_read_keyboard(buffer: *mut u8 ,size: usize, nonblock: bool) -> isize`
(Should I also rename the systemcall to sys_read_pc_keyboard?)
This would be pretty flexible, we can throw standard errorcodes or return the buffersize in bytes like this.
We can then use a semaphore to block the thread until there are scancodes in the vecdeque.

src/arch/x86_64/kernel/pc_keyboard.rs
jounathaen: I agree with @GloriousAlpaca. This is not something for the kernel.

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::kernel::interrupts;
+
+const PS2_DATA_PORT: u16 = 0x60;
+const PS2_CMD_PORT: u16 = 0x64;
+const PS2_CMD_READ_CNFG: u8 = 0x20;
+const PS2_CMD_WRITE_CNFG: u8 = 0x60;
+const PS2_CMD_DISABLE_KEYBOARD: u8 = 0xad;
+const PS2_CMD_DISABLE_MOUSE: u8 = 0xa7;
+const PS2_CMD_ENABLE_KEYBOARD: u8 = 0xae;
```
jounathaen: Also here, this looks like an enum for me in conjunction with the `read_cmd`/`write_cmd` functions.

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,85 @@
...
+	Lazy::new(|| InterruptTicketMutex::new(VecDeque::with_capacity(BUFFER_SIZE)));
+
+fn keyboard_handler() {
+	let scancode = Ps2::read_data();
+	let mut buffer = KEYBOARD_BUFFER.lock();
+
+	if buffer.len() >= BUFFER_SIZE {
+		buffer.pop_front();
+	}
+	buffer.push_back(scancode);
```
jounathaen: I wonder if retaining the keys in the queue is how this is handled best. I'm thinking that maybe adding a timestamp to each key event and discarding it after x seconds is a correct approach. But maybe I'm prematurely optimizing this. It would be interesting to know how other systems are handling this.

Cargo.toml
```diff
@@ -216,6 +216,15 @@ virtio-vsock = ["virtio"]
 ## This is only useful on PCs (x86-64).
 vga = []
 
+## Enables the PS/2 keyboard driver.
+## 
+## This feature initializes the PS/2 keyboard controller and installs a keyboard interrupt handler. 
+## It also provides a system call to receive the last scancode from the internal keyboard buffer. 
+## Note that this is not a complete keyboard driver and not needed for general keyboard support.
```
jounathaen:
```suggestion
## Note that this is not a complete keyboard driver and not needed for serial input/output.
```

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,129 @@
...
+const PS2_CNFG_ENABLE_KEYBOARD_INTERRUPT: u8 = 0x01;
+const PS2_BUFFER_FULL: u8 = 0x01;
+
+const BUFFER_SIZE: usize = 256;
+static KEYBOARD_SEMAPHORE: Semaphore = Semaphore::new(0);
+
+struct Ps2;
+impl Ps2 {
+	pub fn read_status() -> u8 {
+		unsafe { Port::<u8>::new(PS2_CMD_PORT).read() }
```
jounathaen, line 35: Please add a brief safety comment:
```suggestion
		// SAFETY: Correct port access without safety related side-effects.
		unsafe { Port::<u8>::new(PS2_CMD_PORT).read() }
```

src/syscalls/system.rs
```diff
@@ -6,3 +6,22 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
 pub extern "C" fn sys_getpagesize() -> i32 {
 	BasePageSize::SIZE.try_into().unwrap()
 }
+
+#[cfg(all(target_arch = "x86_64", feature = "pc-keyboard"))]
+#[hermit_macro::system]
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn sys_read_keyboard(buffer: *mut u8, size: usize, nonblock: bool) -> isize {
```
jounathaen, line 18: Please add a doc comment

src/syscalls/system.rs
```diff
@@ -6,3 +6,22 @@ use crate::arch::mm::paging::{BasePageSize, PageSize};
...
+	if buffer.is_null() {
+		return -(crate::errno::Errno::Fault as isize);
+	}
+	if size == 0 {
+		return 0;
+	}
+	let buffer_slice: &mut [u8] = unsafe { core::slice::from_raw_parts_mut(buffer, size) };
```
jounathaen, line 27:
```suggestion
    // SAFETY: We have to trust the user input, because we are a unikernel and if the user wants to crash the program they are free to do so.
	let buffer_slice: &mut [u8] = unsafe { core::slice::from_raw_parts_mut(buffer, size) };
```

src/syscalls/system.rs
jounathaen, line 18: Please add a doc comment

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,129 @@
...
+	EnableKeyboard = 0xae,
+	#[allow(dead_code)]
+	EnableMouse = 0xa8,
+	TestFirstPort = 0xab,
+}
+
+const PS2_CNFG_ENABLE_KEYBOARD_INTERRUPT: u8 = 0x01;
+const PS2_BUFFER_FULL: u8 = 0x01;
+
+const BUFFER_SIZE: usize = 256;
```
jounathaen:
```suggestion
const MAX_INP_BUFFER_SIZE: usize = 256;
```

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,129 @@
...
+			// Pop the oldest scancode if the buffer is full.
+			if buffer.len() >= BUFFER_SIZE {
+				buffer.pop_front();
+				sem = false;
+			}
+			buffer.push_back(valid_scancode);
+		}
+		if sem {
+			KEYBOARD_SEMAPHORE.release();
+		}
```
jounathaen: Logic is fine, but you could rewrite it to be a bit simpler by returning in the line that is currently `sem = false` (don't forget the `push_back`).

src/syscalls/system.rs
jounathaen, line 18: Please add a doc comment

src/syscalls/system.rs
jounathaen, line 27:
```suggestion
    // SAFETY: We have to trust the user input, because we are a unikernel and if the user wants to crash the program they are free to do so.
	let buffer_slice: &mut [u8] = unsafe { core::slice::from_raw_parts_mut(buffer, size) };
```

src/arch/x86_64/kernel/pc_keyboard.rs
```diff
@@ -0,0 +1,129 @@
...
+		unsafe { Port::<u8>::new(PS2_DATA_PORT).read() }
+	}
+
+	pub fn write_data(data: u8) {
+		unsafe { Port::<u8>::new(PS2_DATA_PORT).write(data) }
+	}
+}
+
+static KEYBOARD_BUFFER: Lazy<InterruptTicketMutex<VecDeque<NonZeroU8>>> =
+	Lazy::new(|| InterruptTicketMutex::new(VecDeque::with_capacity(BUFFER_SIZE)));
```
jounathaen:
```suggestion
  	Lazy::new(|| InterruptTicketMutex::new(VecDeque::with_capacity(32)));
```

jounathaen (approved):

PR #2416, opened 2026-05-07
feat(env): remove comptime env vars

mkroening:
The kernel is currently compiling the value of the following environment variables into the image:

- `NO_COLOR`
- `HERMIT_LOG_LEVEL_FILTER`
- `HERMIT_MTU`
- `HERMIT_MRG_RXBUF_SIZE`
- `HERMIT_IP`
- `HERMIT_GATEWAY`
- `HERMIT_MASK`
- `HERMIT_DNS1`
- `HERMIT_DNS2`
- `UHYVE_MOUNT`

The following variables are only set via the runtime environment:

- `HERMIT_PCAP_PATH`

The following variables affect the compilation but are not visible to the kernel:

- `HERMIT_CAREFUL` (via xtask)
- `HERMIT_LOG_STATIC_MAX_LEVEL` (future, via xtask)
- `HERMIT_MANIFEST_DIR` (wrapper crate only)

This PR proposes making all kernel-visible environment variables runtime-only, since there is no inherent benefit of compiling them into the image. Instead, they should be set via kernel args (e.g., `env=NO_COLOR=1`). This avoids recompilation and forces users to not compile them into the image. This also reduces maintenance of the wrapper crate by reducing the rebuild-inducing env vars.

What are your opinions on this? I am also fine with going into the other direction, if that turns out to be the consensus.

Depends on:
- https://github.com/hermit-os/kernel/pull/2415
- https://github.com/hermit-os/kernel/pull/2638

PR #2540, opened 2026-07-09
feat(xtask): add support for custom VMM and Hermit args

mkroening:
This PR adds support for passing custom arguments to VMM and Hermit when using xtask. This allows you to set runtime environment variables via Hermit arguments, for example, which is necessary for https://github.com/hermit-os/kernel/pull/2416 to be useful.

Instead of

```bash
HERMIT_LOG_LEVEL_FILTER=trace cargo xtask ci rs --arch x86_64 --package rusty_demo qemu
```

You would now do

```bash
cargo xtask ci rs --arch x86_64 --package rusty_demo qemu -- -- env=HERMIT_LOG_LEVEL_FILTER=trace
```

### DIFF
--- a/xtask/src/ci/firecracker.rs
+++ b/xtask/src/ci/firecracker.rs
@@ -5,22 +5,32 @@ use anyhow::Result;
 use clap::Args;
 use xshell::cmd;
 
+use crate::ci;
+
 /// Run image on Firecracker.
 #[derive(Args)]
 pub struct Firecracker {
 	/// Run Firecracker using `sudo`.
 	#[arg(long)]
 	sudo: bool,
+
+	/// Arguments to pass to Firecracker and Hermit, separated by another `--`.
+	#[arg(last = true)]
+	firecracker_and_hermit_args: Vec<String>,
 }
 
 impl Firecracker {
 	pub fn run(self, image: &Path, smp: usize) -> Result<()> {
 		let sh = crate::sh()?;
 
+		let (firecracker_args, hermit_args) = ci::split_args(&self.firecracker_and_hermit_args);
+		let quoted_hermit_args = shlex::try_join(hermit_args.iter().map(AsRef::as_ref))?;
+
 		let config = format!(
 			include_str!("firecracker_vm_config.json"),
 			kernel_image_path = "hermit-loader-x86_64-fc",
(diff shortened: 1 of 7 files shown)


xtask/src/ci/firecracker.rs
```diff
@@ -5,22 +5,31 @@ use anyhow::Result;
...
 	pub fn run(self, image: &Path, smp: usize) -> Result<()> {
 		let sh = crate::sh()?;
 
+		let (firecracker_args, hermit_args) = ci::split_args(&self.firecracker_and_hermit_args);
+
 		let config = format!(
 			include_str!("firecracker_vm_config.json"),
 			kernel_image_path = "hermit-loader-x86_64-fc",
 			initrd_path = image.display(),
+			boot_args = hermit_args.join(" "),
```
fogti: Shouldn't these undergo general shell argument escaping (e.g. handling of quotes, spaces in arguments, etc.)?

xtask/src/ci/uhyve.rs
```diff
@@ -5,26 +5,37 @@ use anyhow::Result;
...
+
 		let uhyve = env::var("UHYVE").unwrap_or_else(|_| "uhyve".to_owned());
 		let program = if self.sudo { "sudo" } else { uhyve.as_str() };
 		let arg = self.sudo.then_some(uhyve.as_str());
 		let smp_arg = format!("--cpu-count={smp}");
 
-		cmd!(sh, "{program} {arg...} {smp_arg} {image}")
-			.env("RUST_LOG", "debug")
-			.run()?;
+		cmd!(
```
fogti, line 31: imo here `{uhyve_and_hermit_args...}` should be passed through as-is, instead of splitting the arguments and then concatenating them again.

xtask/src/ci/firecracker.rs
mkroening: Good catch. I added quoting using `shlex`. 👍

xtask/src/ci/uhyve.rs
mkroening: I also considered it but was undecided. I have changed it to your preference. 👍

fogti (approved):

PR #2545, opened 2026-07-10
ci(benchmarks): adapt to new benchmark paths

mkroening:
For details, see https://github.com/hermit-os/hermit-rs/pull/1019.


jounathaen:
Looks to me as if this test was successful, right?

PR #2547, opened 2026-07-10
feat: Add virtio-rng driver

### LINKED ISSUE #1752: Tracking issue for a virtio-rng driver
It would be nice to have a [virtio-rng](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-3050004) driver in Hermit.

sbutz:
In virtualized deployments seed DRNG from entropy provided by host.

Tbh this driver is of limited use, but was good exercise to get started with virtio.

|Arch|mmio|pci|
|---|---|---|
|x86_64|✅|✅|
|aarch64|✅|✅|
|riscv64|✅|❌ - pci will be added with https://github.com/hermit-os/kernel/pull/2550|

Blocked by integration of https://github.com/rust-osdev/virtio-spec-rs/pull/24 and new release of virtio-spec-rs.

### DIFF
--- a/src/drivers/rng.rs
+++ b/src/drivers/rng.rs
@@ -1,0 +1,198 @@
+//! A virtio-rng driver.
+//!
+//! For details on the device, see [Entropy Device].
+//!
+//! [Entropy Device]: https://docs.oasis-open.org/virtio/virtio/v1.4/cs01/virtio-v1.4-cs01.html#x1-4130004
+//!
+
+use alloc::vec::Vec;
+
+use embedded_io::{ErrorType, Read};
+use pci_types::InterruptLine;
+use smallvec::SmallVec;
+use volatile::VolatileRef;
+use volatile::access::ReadOnly;
+
+use crate::config::{RNG_PACKET_SIZE, VIRTIO_MAX_QUEUE_SIZE};
+use crate::drivers::error::DriverError;
+#[cfg(not(feature = "pci"))]
+use crate::drivers::mmio::get_rng_driver;
+#[cfg(feature = "pci")]
+use crate::drivers::pci::get_rng_driver;
+use crate::drivers::virtio::error::VirtioRngError;
+use crate::drivers::virtio::transport::UniCapsColl;
+use crate::drivers::virtio::virtqueue::split::SplitVq;
+use crate::drivers::virtio::virtqueue::{
+	AvailBufferToken, BufferElem, BufferType, UsedBufferToken, VirtQueue, Virtq,
+};
+use crate::drivers::{Driver, InterruptHandlerMap};
+use crate::errno::Errno;
+use crate::mm::device_alloc::DeviceAlloc;
+
+pub fn seed_entropy() -> Option<[u8; 32]> {
+	get_rng_driver().and_then(|drv| {
+		let mut buf = [0u8; 32];
+		if drv.lock().read(&mut buf).ok()? == buf.len() {
+			Some(buf)
+		} else {
+			None
+		}
+	})
+}
+
+fn fill_queue(vq: &mut VirtQueue, num_packets: u16, packet_size: u32) {
+	for _ in 0..num_packets {
+		let buff_tkn = match AvailBufferToken::new(SmallVec::new(), {
+			let mut vec = SmallVec::new();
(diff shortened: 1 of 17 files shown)


sbutz:
@mkroening does the new feature `virtio-entropy` require a new kernel version and a bump in hermit-rs?
The ci tests are added in the last commit. It can be moved to separate PR which can be merged after the next new kernel release and bump in hermit-rs.

mkroening:
> @mkroening does the new feature `virtio-entropy` require a new kernel version and a bump in hermit-rs? The ci tests are added in the last commit. It can be moved to separate PR which can be merged after the next new kernel release and bump in hermit-rs.

All it should need is the new feature being listed in the hermit-rs wrapper crate's Cargo.toml. A new kernel release or kernel bump in hermit-rs should not be necessary for the kernel CI to pick up the feature. Kernel CI always uses the latest hermit-rs commit. So please open a PR in hermit-rs, adding the feature. That uncontroversial one-line change can be merged immediately to unblock your further work here.

Sorry for the long wait. I have not forgotten this PR and will review it in the next days. :)

Edit: Thanks for keeping this PR up to date and rebasing on https://github.com/hermit-os/kernel/pull/2527, which I had prioritized for review in preparation for this PR.

sbutz:
Adding feature flag in hermit-rs: https://github.com/hermit-os/hermit-rs/pull/1049

src/entropy.rs
```diff
@@ -36,12 +40,20 @@ pub fn read(buf: &mut [u8], _flags: Flags) -> io::Result<usize> {
...
-				return Err(Errno::Nosys);
+			let seed = {
+				#[cfg(feature = "virtio-rng")]
+				{
+					virtio_seed_entropy()
+				}
+				#[cfg(not(feature = "virtio-rng"))]
+				{
+					processor_seed_entropy()
+				}
```
mkroening: We could support both at the same time:

```rust
fn seed_entropy() -> Option<[u8; 32]> {
    #[cfg(feature = "virtio-rng")]
    if let Some(entropy) = virtio_seed_entropy() {
        return Some(entropy);
    }

    processor_seed_entropy()
}
```

Would you say it makes sense to fall back to the processor?

src/drivers/rng.rs
```diff
@@ -0,0 +1,207 @@
+//! A virtio-rng driver.
+//!
+//! For details on the device, see [Entropy Device].
+//! For details on the Rust definitions, see [`virtio::entropy`].
+//!
+//! [Entropy Device]: https://docs.oasis-open.org/virtio/virtio/v1.4/virtio-v1.4.pdf#page=180.53
```
mkroening: What do you think about using the HTML link instead? :)

```suggestion
//! [Entropy Device]: https://docs.oasis-open.org/virtio/virtio/v1.4/cs01/virtio-v1.4-cs01.html#x1-4130004
```

mkroening (changes requested):
Sorry again for the wait! :)

This looks great and can be merged nearly as is. Sorry for the amount of code duplication you had to do in the arch-specific modules. We are trying to improve the situation slowly.

My main comment is regarding the virtio-spec types. As mentioned in https://github.com/rust-osdev/virtio-spec-rs/pull/24#pullrequestreview-4907565402, I don't see too much value in defining empty types that might not be necessary in other kernel architectures. In my local test, your PR should still work fine when using `type DeviceFeatures = virtio::F` and `type Config = ()`. What do you think?

Other than that, please squash the CI commit into the addition of the driver.

src/drivers/rng.rs
sbutz: Done

src/entropy.rs
sbutz: Yes.
Any users who has strong requirements on the entropy used, will anyhow configure the hypervisor to trap when the guest tries to execute the respective instructions and inject entropy from a suitable source.

mkroening (approved):
Thanks a lot! :)

mkroening:
Removed some allows that should not be needed anymore: https://github.com/hermit-os/kernel/compare/e8bc06f9e5a2e2a7132c453b47bfab91d58dde0b..9c773d3acf4731b186761a045dcc7908075ce74c

PR #2550, opened 2026-07-13
feat(riscv64): Add PCI support

sbutz:
Follow-up Tasks:
- PCI Handling is very similar across architectures. Reuse code across architectures where applicable.
- Upgrade crate fdt to version 0.2 once released

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,12 +177,12 @@ jobs:
             qemu_flags: --accel --sudo
             rs_flags: --features hermit/loader
           - arch: aarch64
-            packages: qemu-system-arm-hwe ipxe-qemu
+            packages: qemu-system-arm-hwe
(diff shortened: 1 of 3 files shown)


stlankes (approved):

mkroening (changes requested):
Thanks! Excellent work! :)

Could you enable the RISC-V PCI tests on CI that are currently disabled for RISC-V?

mkroening:
I think CI needs ixpe-qemu on all architectures now. I have added it.

mkroening:
RTL8139 does not work, but that is not relevant for this PR. I have excluded them for RISC-V again and opened https://github.com/hermit-os/kernel/issues/2636.

mkroening:
mio seems broken too, though I am not sure if that is because of PCI, since we don't test it on MMIO.

mkroening (approved):

PR #2487, opened 2026-06-18
fix(aarch64): SMP and timers on macOS HVF

stlankes:
This PR makes Hermit successfully run with long-running SMP jobs on Apple Silicon with macOS HVF through QEMU.

To reproduce, run:

```bash
cargo xtask ci rs --arch aarch64 --package rusty_demo --no-default-features --features hermit/loader --release --smp 2 qemu --accel
```

Closes https://github.com/hermit-os/kernel/issues/1686.

PR #2555, opened 2026-07-15
parse the timer interrupt only once

stlankes:
init() (boot core) and init_cpu() (application processors) both parsed the "arm,armv8-timer" node of the device tree to find the timer interrupt, and each picked its own triplet out of the four the node lists. The timer is a PPI, so every core enables it in its own redistributor — if the two parsers ever disagree about which interrupt the kernel uses, the application processors enable an interrupt that never fires and tasks sleeping on them are not woken up again.

Parse the node once in init(), cache the interrupt and its trigger mode, and let both callers enable it from there. No functional change: both sites selected the Non-secure Phys timer before.


stlankes:
PR includes #2558 to pass the CI tests.

stlankes:
replaced by #2487

PR #2556, opened 2026-07-16
 refactor(virtio-console): remove VirtioUART wrapper

cagatay-y:
Rather than accessing the virtio-console driver through the global driver vector, place it into the using struct in a way similar to what was done in b735657.

### DIFF
--- a/src/console/mod.rs
+++ b/src/console/mod.rs
@@ -1,6 +1,8 @@
 #[cfg(feature = "uhyve")]
 mod uhyve;
 
+#[cfg(feature = "virtio-console")]
+use alloc::boxed::Box;
 use core::{fmt, mem};
 
 use embedded_io::{ErrorType, Read, ReadReady, Write};
@@ -9,7 +11,7 @@ use hermit_sync::{InterruptTicketMutex, Lazy};
 
 use crate::arch::kernel::serial::SerialDevice;
 #[cfg(feature = "virtio-console")]
-use crate::drivers::console::VirtioUART;
+use crate::drivers::console::VirtioConsoleDriver;
 use crate::errno::Errno;
 use crate::executor::WakerRegistration;
 
@@ -20,7 +22,7 @@ pub(crate) enum IoDevice {
 	Uhyve(uhyve::UhyveSerial),
 	Uart(SerialDevice),
 	#[cfg(feature = "virtio-console")]
-	Virtio(VirtioUART),
+	Virtio(Box<VirtioConsoleDriver>),
 }
 
 impl ErrorType for IoDevice {
@@ -77,7 +79,7 @@ impl Write for IoDevice {
 }
 
 pub(crate) struct Console {
-	device: IoDevice,
+	pub device: IoDevice,
 	buffer: Vec<u8, SERIAL_BUFFER_SIZE>,
 }
 
@@ -88,11 +90,12 @@ impl Console {
 			buffer: Vec::new(),
 		}
 	}
+}
 
-	#[cfg(feature = "virtio-console")]
-	pub fn replace_device(&mut self, device: IoDevice) {
-		self.device = device;
-	}
+#[cfg(feature = "virtio-console")]
+pub(crate) fn switch_to_virtio(device: VirtioConsoleDriver) {
+	info!("Switch to virtio console");
+	CONSOLE.lock().device = IoDevice::Virtio(Box::new(device));
 }
 
 impl ErrorType for Console {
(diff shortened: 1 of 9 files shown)


src/init_cell.rs
```diff
@@ -1,9 +1,5 @@
 #![cfg_attr(
-	not(any(
-		feature = "virtio-vsock",
-		feature = "virtio-fs",
-		feature = "virtio-console"
-	)),
+	not(any(feature = "virtio-vsock", feature = "virtio-fs",)),
```
mkroening:
```suggestion
	not(any(feature = "virtio-vsock", feature = "virtio-fs")),
```

Please also search for other occurrences of `,)` introduced by this PR.

src/drivers/pci.rs
```diff
@@ -331,37 +323,28 @@ pub(crate) fn print_information() {
...
-		match self {
-			Self::VirtioFs(drv) => Some(drv),
-			#[allow(unreachable_patterns)]
-			_ => None,
+		#[cfg_attr(not(feature = "virtio-vsock"), expect(irrefutable_let_patterns))]
+		if let Self::VirtioFs(drv) = self {
+			Some(drv)
+		} else {
+			None
 		}
```
mkroening: You change a lot of these from matches to if-lets. This is not related to the point of the PR, though, is it?

In that case, could you please remove these changes from this PR and propose them separately, if you think they are worthwhile? Why change this anyway? I find the former more readable, personally.

src/drivers/console.rs
```diff
@@ -340,9 +299,9 @@ impl super::virtio::VirtioDriver for VirtioConsoleDriver {
 				InterruptCapability::IsrStatus(_) => {
 					let irq = irq.unwrap();
 					handlers.entry(irq).or_default().push_back(|| {
-						if let Some(driver) = get_console_driver() {
-							driver.lock().handle_interrupt();
-						};
+						if let IoDevice::Virtio(ref mut driver) = CONSOLE.lock().device {
+							driver.handle_interrupt();
+						}
```
mkroening: I think we should avoid `ref` if possible.

```suggestion
						if let IoDevice::Virtio(driver) = &mut CONSOLE.lock().device {
							driver.handle_interrupt();
						}
```

src/console/mod.rs
```diff
@@ -9,18 +9,19 @@ use hermit_sync::{InterruptTicketMutex, Lazy};
...
 
+#[allow(clippy::large_enum_variant)]
 pub(crate) enum IoDevice {
 	#[cfg(feature = "uhyve")]
 	Uhyve(uhyve::UhyveSerial),
 	Uart(SerialDevice),
 	#[cfg(feature = "virtio-console")]
-	Virtio(VirtioUART),
+	Virtio(VirtioConsoleDriver),
 }
```
mkroening: How about boxing to avoid the large enum variant?

mkroening (changes requested):
Great effort! Thanks a lot! :)

src/drivers/pci.rs
cagatay-y: I changed it to avoid the `clippy::match-wildcard-for-single-variants` lint, which is now an issue because only two variants are left, but we can also just allow it. While testing the lint I also noticed that this is not an issue for enums that are marked with `#[non_exhaustive]`, so we can also go with that. Non-exhaustiveness also seems to allow us to remove the existing `#[allow(unreachable_patterns)]`es.

mkroening: Ah, sounds great! Could you do that in a preceding commit or PR? :)

cagatay-y: I guess I ran the wrong clippy command when testing it. `non_exhaustive` does not help. I will just allow `clippy::match-wildcard-for-single-variants`

mkroening: I see. Then I guess to swap two lints for one makes sense. Could you still do it in a preceding commit or PR? :)

mkroening: I think there may also be new conflicts due to https://github.com/hermit-os/kernel/pull/2547.

cagatay-y: I figured out the reason I was confused: `non_exhaustive` does not help with `unreachable_patterns` but it does help with `clippy::match-wildcard-for-single-variants`. We can keep the lint count constant and not change the matching structure by adding `#[non-exhaustive]` to MMIO enums (PCI already has it).

Sorry for the confusion :)

mkroening: Yeah, that sounds good. Do you want to do it, or should I? :)

cagatay-y: I've pushed a commit for it.

mkroening: Ah sorry. I misunderstood. Squashing into one commit would reduce the noise. 😅 

And CI is not happy. 👀

mkroening (approved):
Thanks a lot! :)

PR #2557, opened 2026-07-17
refactor(aarch64/lscpu): Make implementer IDs more obvious

Gelbpunkt:
It was just pointed out to me that these are actually mostly just ASCII values of the first letters of the manufacturer names. Where it makes sense, we can use this more obvious representation of the implementer IDs.


mkroening (approved):
Thanks! :)

PR #2558, opened 2026-07-19
xtask: wait for the stdin echo instead of a fixed delay

stlankes:
test_stdin fed each message to the guest with a one-second pause and then terminated QEMU immediately, reading the echoed output only afterwards. On a slow target the last echo had not been written yet when QEMU was killed, so the collected output ended mid-line and the assertion failed intermittently. This showed up on [aarch64_be](https://github.com/hermit-os/kernel/actions/runs/29453848137/job/87482443036?pr=2555), which is emulated without hardware acceleration.

Collect the guest's stdout on a background thread and wait until every message has been echoed (bounded by a 60 s timeout) before terminating QEMU. The check no longer depends on the guest keeping pace with a fixed schedule.

### DIFF
--- a/xtask/src/ci/qemu.rs
+++ b/xtask/src/ci/qemu.rs
@@ -3,7 +3,8 @@ use std::net::{IpAddr, Ipv4Addr, SocketAddr, TcpStream, UdpSocket};
 use std::path::Path;
 use std::process::{Child, Command, ExitStatus, Stdio};
 use std::str::from_utf8;
-use std::time::Duration;
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
 use std::{env, fs, io, thread};
 
 use anyhow::{Context, Result, bail, ensure};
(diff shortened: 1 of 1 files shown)


mkroening (changes requested):
This alone does not seem to fix the issue: https://github.com/hermit-os/kernel/actions/runs/29735468922/job/88329640752?pr=2293

stlankes:
If remember correctly, this issue was fixed by [2f10654](https://github.com/hermit-os/kernel/pull/2555/commits/2f10654572d860865cb001696671e9d5f8c74733)

mkroening:
> If remember correctly, this issue was fixed by [2f10654](https://github.com/hermit-os/kernel/pull/2555/commits/2f10654572d860865cb001696671e9d5f8c74733)

I don't think that fixes the issue. I think you were lucky. The issue is spurious, and I can still reproduce it with all of https://github.com/hermit-os/kernel/pull/2555 merged: https://github.com/hermit-os/kernel/actions/runs/29830730765/job/88634571115?pr=2293

PR #2572, opened 2026-07-22
fix(virtio/pci): map BARs only once

cagatay-y:
It is possible for multiple capabilities to be accessed through the same base address, so only create a new mapping if one does not already exist for the BAR slot.

### DIFF
--- a/src/drivers/net/rtl8139.rs
+++ b/src/drivers/net/rtl8139.rs
@@ -9,7 +9,7 @@ use core::mem::ManuallyDrop;
 use core::ptr::NonNull;
 
 use endian_num::{le16, le32, le64};
-use pci_types::{Bar, CommandRegister, MAX_BARS};
+use pci_types::CommandRegister;
 use smoltcp::phy::DeviceCapabilities;
 use thiserror::Error;
 use volatile::access::{NoAccess, ReadOnly, ReadWrite};
@@ -746,19 +746,15 @@ pub(crate) fn init_device(
 	handlers: &mut InterruptHandlerMap,
 ) -> Result<RTL8139Driver, DriverError> {
 	let irq = device.get_irq().unwrap();
-	let mut regs = None;
-
-	for i in 0..MAX_BARS {
-		let Some(Bar::Memory32 { .. }) = device.get_bar(i.try_into().unwrap()) else {
-			continue;
-		};
-
-		let (addr, _size) = device.memory_map_bar(i.try_into().unwrap(), true).unwrap();
-
-		regs = Some(unsafe { VolatileRef::new(NonNull::new(addr.as_mut_ptr()).unwrap()) });
-	}
-
-	let mut regs = regs.ok_or(DriverError::InitRTL8139DevFail(RTL8139Error::Unknown))?;
+	let mut regs = device
+		.memory_map_bars(true)
+		.into_iter()
+		.find_map(|bar| {
+			bar.map(|(addr, _size)| unsafe {
+				VolatileRef::new(NonNull::new(addr.as_mut_ptr()).unwrap())
+			})
+		})
+		.ok_or(DriverError::InitRTL8139DevFail(RTL8139Error::Unknown))?;
 
 	debug!("Found RTL8139 at IO {regs:?} (irq {irq})");
 
--- a/src/drivers/virtio/transport/pci.rs
+++ b/src/drivers/virtio/transport/pci.rs
@@ -581,6 +581,8 @@ pub(crate) fn map_caps(
 	#[cfg(target_arch = "x86_64")]
 	let mut msix_table = None;
 
+	let bar_mappings = device.memory_map_bars(true);
+
 	// Reads all PCI capabilities, starting at the capabilities list pointer from the
 	// PCI device.
 	//
@@ -594,7 +596,7 @@ pub(crate) fn map_caps(
 					continue;
 				}
 				let slot = cap.bar;
-				let Some((addr, size)) = device.memory_map_bar(slot, true) else {
+				let Some((addr, size)) = bar_mappings[usize::from(slot)] else {
 					continue;
 				};
 				let Some(pci_cap) = PciCap::new(
@@ -661,9 +663,12 @@ pub(crate) fn map_caps(
 			#[cfg(target_arch = "x86_64")]
 			PciCapability::MsiX(mut msix_capability) => {
 				msix_capability.set_enabled(true, device.access());
-				let (base_addr, _) = device
-					.memory_map_bar(msix_capability.table_bar(), true)
-					.unwrap();
+
+				// the capability should provide a valid BAR ID and "[t]he BAR [...] must map
+				// Memory Space" (PCIe spec. 6.0 sec. 7.7.2) for the MSI-X capability
+				let (base_addr, _) =
+					bar_mappings[usize::from(msix_capability.table_bar())].unwrap();
+
 				let table_ptr = NonNull::slice_from_raw_parts(
 					NonNull::with_exposed_provenance(
 						core::num::NonZero::new(
--- a/src/drivers/pci.rs
+++ b/src/drivers/pci.rs
@@ -111,53 +111,65 @@ impl<T: ConfigRegionAccess> PciDevice<T> {
 		}
 	}
 
-	/// Memory maps pci bar with specified index to identical location in virtual memory.
+	/// Memory maps pci BARs to identical location in virtual memory.
 	/// no_cache determines if we set the `Cache Disable` flag in the page-table-entry.
-	/// Returns (virtual-pointer, size) if successful, else None (if bar non-existent or IOSpace)
-	pub fn memory_map_bar(&self, index: u8, no_cache: bool) -> Option<(VirtAddr, usize)> {
-		let (address, size, prefetchable, _width) = match self.get_bar(index) {
-			Some(Bar::Io { .. }) => {
-				warn!("Cannot map IOBar!");
+	/// Element at index is [Some] if the mapping of the BAR at the same index is successful, else [None] (if bar non-existent or IOSpace)
+	pub fn memory_map_bars(&self, no_cache: bool) -> [Option<(VirtAddr, usize)>; MAX_BARS] {
+		let mut should_skip = false;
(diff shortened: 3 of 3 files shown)


stlankes:
@cagatay-y @mkroening @jounathaen Sorry, this PR was my test if an LLM is able to support reviews. I am not sure, if this helps...

src/drivers/virtio/transport/pci.rs
```diff
@@ -695,9 +701,8 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
 			#[cfg(target_arch = "x86_64")]
 			PciCapability::MsiX(mut msix_capability) => {
 				msix_capability.set_enabled(true, device.access());
-				let (base_addr, _) = device
-					.memory_map_bar(msix_capability.table_bar(), true)
-					.unwrap();
+				let (base_addr, _) =
+					bar_mappings[usize::from(msix_capability.table_bar())].unwrap();
```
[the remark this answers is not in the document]
cagatay-y: If `memory_map_bar` returns `None`, either the BAR is in IO space, which is forbidden for MSI-X capabilities, or the BAR index is invalid, which would indicate an error with the device, or somewhere along the way we made a mistake that we cannot really recover from, so I think it is reasonable to panic but I added a commit to `expect` instead of `unwrap` to clarify the reasons.

src/drivers/virtio/transport/pci.rs
```diff
@@ -638,7 +644,7 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
 					continue;
 				}
 				let slot = cap.bar;
-				let Some((addr, size)) = device.memory_map_bar(slot, true) else {
+				let Some((addr, size)) = *bar_mappings[usize::from(slot)] else {
```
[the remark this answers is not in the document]
cagatay-y: The `LazyCell` also does the same call.

src/drivers/virtio/transport/pci.rs
```diff
@@ -625,6 +627,10 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
 	#[cfg(target_arch = "x86_64")]
 	let mut msix_table = None;
 
+	let bar_mappings = array::from_fn::<_, 6, _>(|slot| {
+		LazyCell::new(move || device.memory_map_bar(u8::try_from(slot).unwrap(), true))
```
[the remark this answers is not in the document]
cagatay-y: `device` is a reference that is provided to the function, so if I am not mistaken its lifetime is the whole function, and `bar_mappings` does not escape the function. I also suspect that the compiler would enforce that lifetime but I'm not sure.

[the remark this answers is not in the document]
cagatay-y: > `slot` is derived from `cap.bar` (a u8)

`slot` is derived from the size of the array, which is 6. It wouldn't panic because `memory_map_bar` would simply return `None`, but the code would be incorrect already by the size 7.

cagatay-y:
> Sorry, this PR was my test if an LLM is able to support reviews. I am not sure, if this helps...

I replied to the comments in case it is useful for the evaluation of the test :)

src/drivers/virtio/transport/pci.rs
```diff
@@ -625,6 +627,10 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
...
+	});
+
```
mkroening: Instead of having lazy options, why not eagerly map the BARs to have them available at the beginning? This would seem more straightforward to me.

Are you expecting a performance impact due to evaluating BARs that are not used afterwards?

mkroening (changes requested):
Thanks! :)

src/drivers/virtio/transport/pci.rs
cagatay-y: Not really. The lazy cell just seemed to better match what the function intends to do, but constructing the array eagerly also makes sense, considering the fact that the lazy cell consumes more memory (though probably also insignificant performance-wise). The related functions should be able to handle non-existent BAR values.

mkroening: If performance is not a concern here, I think eagerly evaluating is easier to understand, then. That way, we can avoid having to keep in mind when exactly this is evaluated later.

Or are we expecting that most BARs will end up unused?

cagatay-y: I didn't test them all but if I recall correctly the ones I tested used one for all the virtio capabilities and one for the MSI-X table, if it exists.

src/drivers/virtio/transport/pci.rs
```diff
@@ -695,9 +699,9 @@ pub(crate) fn map_caps(device: &PciDevice<PciConfigRegion>) -> Result<UniCapsCol
 			#[cfg(target_arch = "x86_64")]
 			PciCapability::MsiX(mut msix_capability) => {
 				msix_capability.set_enabled(true, device.access());
-				let (base_addr, _) = device
-					.memory_map_bar(msix_capability.table_bar(), true)
-					.unwrap();
+				let (base_addr, _) = bar_mappings[usize::from(msix_capability.table_bar())].expect(
+					"the capability should provide a valid BAR ID and \"[t]he BAR [...] must map Memory Space\" (PCIe spec. 6.0 sec. 7.7.2) for the MSI-X capability",
```
mkroening: Could you move this into a comment above the statement? :)

mkroening: Done :)

mkroening (approved):

mkroening (changes requested):
The CI failure does not seem to be spurious. I am wondering what goes wrong exactly. Maybe we need to go back to lazy cells after all. What kind of issue does this PR resolve? Is there a bug, or is this just to save resources?

By the way: where does the number six come from?

cagatay-y:
> The CI failure does not seem to be spurious. I am wondering what goes wrong exactly.

The issue is that if a base address is 64 bits, it occupies two slots, so the slot after it is invalid as a BAR by itself. When we blindly try to map the slot that is occupied by the second part of the previous BAR, it returns values that cannot be interpreted as a BAR. There does not seem to be a way to tell that it is not a BAR by itself without reading the previous one and skipping the current one accordingly.

> Maybe we need to go back to lazy cells after all. 

We can either switch back to lazy cells or solve the issue by iterating over the slot numbers and skipping one if the previous slot was a 64-bit one. That would require changing the current `memory_map_bar` implementation, though, as it does not provide the BAR type information to the caller. Which approach would you prefer?

> What kind of issue does this PR resolve? Is there a bug, or is this just to save resources?

I did not encounter an issue caused by the current behavior but mapping the same address twice seemed like an accident to me.

> By the way: where does the number six come from?

Per PCIe spec. sec. 7.5.1.2.1 "A Type 0 Configuration Space Header has six DWORD locations allocated for Base Address registers starting at offset 10h in Configuration Space." I should introduce a constant with a descriptive name for that.

By the way, do you think it would make sense to have the virtio-console test later amongst the virtio tests? For the common virtio issues it could be nice to have the tests where the console output is not affected earlier, so that the CI output is more helpful.

src/drivers/net/rtl8139.rs
```diff
@@ -746,19 +746,15 @@ pub(crate) fn init_device(
...
-		regs = Some(unsafe { VolatileRef::new(NonNull::new(addr.as_mut_ptr()).unwrap()) });
-	}
-
-	let mut regs = regs.ok_or(DriverError::InitRTL8139DevFail(RTL8139Error::Unknown))?;
+	let mut regs = device
+		.memory_map_bars(true)
+		.into_iter()
+		.find_map(|bar| {
+			bar.map(|(addr, _size)| unsafe {
+				VolatileRef::new(NonNull::new(addr.as_mut_ptr()).unwrap())
```
cagatay-y, line 754: The behaviour is slightly different:
- We now accept 64-bit BARs.
  32-bit BAR choice was made in 7246792, maybe @Gelbpunkt can chime in on if 64-bit BARs are also okay.
- We use the first BAR that we can map.
  The driver originally used the first IO BAR it could find and was changed to use the last one with https://github.com/hermit-os/kernel/commit/75013ce501de5e298c0a59b6e10c93324536873c#diff-81e90c26a83946cda84acea9ea30c793496f20651faf921f080cba1347c3c2e6R438-R444. There is not information in the commit message to indicate if that was intentional. We can reverse the iterator to achieve the same effect, if necessary.

mkroening:
> We can either switch back to lazy cells or solve the issue by iterating over the slot numbers and skipping one if the previous slot was a 64-bit one. That would require changing the current `memory_map_bar` implementation, though, as it does not provide the BAR type information to the caller. Which approach would you prefer?

Let's go lazy, then. ^^

> By the way, do you think it would make sense to have the virtio-console test later amongst the virtio tests? For the common virtio issues it could be nice to have the tests where the console output is not affected earlier, so that the CI output is more helpful.

I did originally put them to the non-virtio `stdin` test, but your reasoning makes sense. Feel free to submit a PR. 👍

cagatay-y:
> Let's go lazy, then. ^^

I've realized that rather than returning the BAR type information to the caller, we can move the array construction into the mapping function. The latest commit does that. Should I revert to lazy cells or is the current version good?

I've also left a review comment about a change in the RTL8139 driver's behavior if we end up going with the eager approach.

src/drivers/net/rtl8139.rs
Gelbpunkt: From what I can see, QEMU's RTL8139 only implements 32-bit memory BARs:

```
[    0.015647][0][INFO  rtl8139   ] found a bar Some(Io { port: 1056899072 })
[    0.016225][0][INFO  rtl8139   ] found a bar Some(Memory32 { address: 268435456, size: 256, prefetchable: false })
[    0.017355][0][INFO  rtl8139   ] found a bar None
```

so in practice this is probably fine (as you can see from the passing tests, we do test RTL8139 on AArch64). From what I can tell, Linux doesn't care whether they are 32-bit or 64-bit either.

mkroening (approved):
Thanks! :)

PR #2573, opened 2026-07-22
ci: add inline LLM review workflow for pull requests

stlankes:
Add a GitHub Actions workflow that runs on non-draft pull requests and posts inline review comments generated by an OpenAI-compatible LLM endpoint. The endpoint, model and API key are provided via repository secrets (LLM_API_BASE, LLM_MODEL, LLM_API_KEY).

PR #2574, opened 2026-07-23
ci: install PR-Agent

mkroening:
This PR replaces the custom python script for PR reviews from https://github.com/hermit-os/kernel/pull/2573 with [PR-Agent](https://docs.pr-agent.ai/).

I have configured it to do nothing automatically. Instead, it has to be used manually, which allows us to be mindful of the token use as well as avoid spamming.

For example,
- [`/describe`](https://docs.pr-agent.ai/tools/describe/) changes the description of the PR.
- [`/review`](https://docs.pr-agent.ai/tools/review/) creates a review comment.
- [`/improve`](https://docs.pr-agent.ai/tools/improve/) creates a suggestion comment.

For a demonstration, see https://github.com/mkroening/libhermit-rs/pull/52. I don't think the model is strong enough to be helpful currently, but it is nice to have for people who prefer it.


stlankes (approved):
A clearly better solution!

PR #2580, opened 2026-07-27
ci: add Pull Request Labeler

mkroening:
This PR adds a workflow to run [actions/labeler](https://github.com/actions/labeler) to automatically label PRs based on touched file paths.

This action is already used by Uhyve: [uhyve/.github/labeler.yml](https://github.com/hermit-os/uhyve/blob/09e301297e1a0bc9c083ead81ed017d11932f299/.github/labeler.yml).

The current labels in this PR are just a gut feeling. It would be great to have your feedback on them and on whether we should do this at all.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,155 @@
+# Meta
+
+cargo:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/Cargo.toml"
+          - "**/Cargo.lock"
+
+ci:
+  - changed-files:
+      - any-glob-to-any-file:
+          - ".github/**"
+
+tests:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "tests/**"
+
+# Crates
+
+hermit-builtins:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "hermit-builtins/**"
+
+hermit-macro:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "hermit-macro/**"
+
+xtask:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "xtask/**"
+
+# Architectures
+
+aarch64:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/arch/aarch64/**"
+
+riscv64:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/arch/riscv64/**"
+
+x86_64:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/arch/x86_64/**"
+
+# Platforms
+
+uhyve:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/uhyve/**"
+          - "**/uhyve.rs"
+
+# Transports
+
+mmio:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/mmio/**"
+          - "**/mmio.rs*"
+
+pci:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/pci/**"
+          - "**/pci.rs"
+
+# Drivers
+
+gem:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/drivers/net/gem/**"
+
+loopback:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/drivers/net/loopback.rs"
+
+rtl8139:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/drivers/net/rtl8139.rs"
+
+virtio-console:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/drivers/console/**"
+
+virtio-fs:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+  - changed-files:
+      - any-glob-to-any-file: 'tests/**'
+
+xtask:
+  - changed-files:
+      - any-glob-to-any-file: 'xtask/**'
+
+xtask:
+  - changed-files:
+      - any-glob-to-any-file: 'xtask/**'
```
Gelbpunkt: This is a duplicate

```suggestion
xtask:
  - changed-files:
      - any-glob-to-any-file: 'xtask/**'
```

.github/labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+  - changed-files:
+      - any-glob-to-any-file: 'src/arch/riscv64/**'
+
+arch-x86_64:
+  - changed-files:
+      - any-glob-to-any-file: 'src/arch/x86_64/**'
+
+driver:
+  - changed-files:
+      - any-glob-to-any-file: 'src/drivers/**'
```
Gelbpunkt: I would like the driver labeling to be a bit more fine-grained, e.g. having `driver-console`, `driver-fs`, `driver-net`, `driver-virtio`, `driver-vsock`, and maybe even the individual drivers inside the directories (like `gem`, `loopback`, `rtl8139`, `virtio-net`, ...)

.github/labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+  - changed-files:
+      - any-glob-to-any-file: 'src/fd/**'
+
+fs:
+  - changed-files:
+      - any-glob-to-any-file: 'src/fs/**'
+
+mm:
+  - changed-files:
+      - any-glob-to-any-file: 'src/mm/**'
```
Gelbpunkt: Shouldn't this also include `src/arch/**/mm/**`?

.github/labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+  - changed-files:
+      - any-glob-to-any-file: 'src/executor/**'
+
+fd:
+  - changed-files:
+      - any-glob-to-any-file: 'src/fd/**'
+
+fs:
+  - changed-files:
+      - any-glob-to-any-file: 'src/fs/**'
```
Gelbpunkt: Maybe also `src/drivers/fs/**`?

.github/labeler.yml
```diff
@@ -0,0 +1,71 @@
...
+      - any-glob-to-any-file: 'src/arch/aarch64/**'
+
+arch-riscv64:
+  - changed-files:
+      - any-glob-to-any-file: 'src/arch/riscv64/**'
+
+arch-x86_64:
+  - changed-files:
+      - any-glob-to-any-file: 'src/arch/x86_64/**'
+
```
sarahspberrypi, line 42: I think, a `src/console/**` label is missing here? This file includes all subdirectories, just not this one.

.github/labeler.yml
sarahspberrypi: I second this.

sarahspberrypi (commented):
Overall, I like the idea, thanks! :)

.github/labeler.yml
mkroening: I went by modules, not directories. All the single-file modules don't have a label. I believe the console module is too small to warrant a label, even though it has two files.

If you disagree, I can also add a label, though.

.github/labeler.yml
mkroening: I think I'd rather keep that separate as `driver-virtio-fs`. 🤔

.github/labeler.yml
mkroening: I had included them in my initial draft but thought they were too numerous, but I can't think of any concrete downside of having many tags. I have added them back. 👍

stlankes (approved):
I like the idea, too.

.github/labeler.yml
```diff
@@ -0,0 +1,136 @@
...
+  - changed-files:
+      - any-glob-to-any-file:
+          - "src/fs/**"
+
+mm:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "**/mm/**"
+
+plat-uhyve:
```
jounathaen: Nitpick: I think either "uhyve" or "platform-uhyve" would be better. "plat" is for my ears a bit ambiguous.

jounathaen (approved):

.github/labeler.yml
sarahspberrypi: Ah fair, then I misunderstood! That makes perfect sense and I agree with you.

sarahspberrypi (approved):

n0toose (approved):

Gelbpunkt (approved):

Gelbpunkt (approved):

PR #2583, opened 2026-07-28
test: prepare vsock

mkroening:
This is in preparation of https://github.com/hermit-os/kernel/pull/2434.


mkroening:
This was merged in https://github.com/hermit-os/kernel/pull/2337.

PR #2584, opened 2026-07-29
[codex] Add SBI console input on RISC-V

### LINKED ISSUE #1838: SBI-based debug console is not read on RISC-V
We should use `sbi_rt::console_read` and `sbi_rt::console_write` instead of `sbi_rt::console_write_byte` for our default RISC-V console.

Jonesxq:
## Summary

- replace per-byte RISC-V SBI console output with bulk `console_write`
- add default RISC-V console input through `console_read`, including readiness caching and async stdin polling
- enable the default RISC-V stdin CI test

## Why

The default RISC-V console only supported output through `console_write_byte` and never read input, so stdin could not work. The SBI debug console extension provides bulk read and write operations that address both limitations.

The SBI API accepts physical addresses, so the console uses a page-aligned bounce buffer embedded in the serial device. Kernel data is identity-mapped on RISC-V; keeping the buffer local to the device also avoids page-table translation during early logging or while the page-table lock is held.

## Impact

Default-console RISC-V guests can now receive stdin, and console output is sent in chunks instead of one SBI call per byte.

## Validation

- built the RISC-V kernel and `hermit-builtins`
- ran repository-wide `cargo xtask clippy` across architectures and feature combinations
- `cargo fmt --all -- --check`
- `git diff --check`

Local QEMU stdin execution was not available because `qemu-system-riscv64` is not installed. The RISC-V stdin CI job is enabled by this change.

Closes #1838

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,6 +258,9 @@ jobs:
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} ${{ matrix.rs_flags }} --package hello_world --no-default-features qemu ${{ matrix.qemu_flags }} --microvm
         if: matrix.arch == 'x86_64'
       - run: cargo xtask ci rs --arch ${{ matrix.arch }} --profile ${{ matrix.profile }} ${{ matrix.rs_flags }} --package stdin qemu ${{ matrix.qemu_flags }}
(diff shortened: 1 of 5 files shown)


mkroening:
Is this ready for review? If yes, please un-draft the PR.

Please attribute your AI assistants in the commit messages: [Attribution](https://docs.kernel.org/process/coding-assistants.html#attribution)

PR #2585, opened 2026-07-30
Fix aarch64 timer interrupt configuration

Gelbpunkt:
We currently misconfigure the timer interrupt on aarch64, which results in it constantly firing if we have a `BOOT_COUNTER` that is not zero. We were also losing precision in some hypothetical edge cases. After fixing both, I see around 7 Gbit/s of throughput in the networking benchmarks rather than 5 Mbit/s and no more constant timer interrupts firing.


mkroening (approved):
Thanks! :)

PR #2606, opened 2026-08-04
ci: enable virtualization HWE stack

mkroening:
Ubuntu has announced to provide more recent QEMU versions to the LTS versions of Ubuntu:
- [Ubuntu’s virtualization hardware enablement (HWE) stack: a new model for confidential computing enablement | Ubuntu](https://ubuntu.com/blog/ubuntu-virtualization-hwe-stack-confidential-computing)
- [Virtualization Hardware Enablement Stack - Ubuntu Server documentation](https://ubuntu.com/server/docs/how-to/virtualisation/virt-hwe/)

While [qemu-hwe](https://packages.ubuntu.com/source/resolute/qemu-hwe) is still on the same version as [qemu](https://packages.ubuntu.com/source/resolute/qemu) at the moment, upgrading to QEMU 11 would allow us to enable semihosting on `aarch64_be`.

Depends on:
- https://github.com/hermit-os/kernel/pull/2493

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,7 +141,7 @@ jobs:
       - name: Install QEMU
         run: |
           sudo apt-get update
-          sudo apt-get install --no-install-recommends qemu-system-x86
+          sudo apt-get install --no-install-recommends qemu-system-x86-hwe
       - uses: mkroening/rust-toolchain-toml@main
       - name: Unit tests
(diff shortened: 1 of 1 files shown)


Gelbpunkt (approved):
If we see any issues with the newer QEMU releases we will hopefully be able to fix them quickly and I believe those are going to occur very rarely if ever

PR #2607, opened 2026-08-05
refactor(virtio-net): use the checksum functions provided by smoltcp

cagatay-y:
This will need to wait until a smoltcp version with the newly exposed functions is released.


mkroening (approved):
Thanks! :)

PR #2609, opened 2026-08-06
fix(pci): disable decoding while sizing BARs

stlankes:
Determining the size of a BAR writes all-ones into it and restores the previous value afterwards. While decoding is enabled, the device decodes those intermediate values as real addresses and relocates itself. For a passed-through device the host follows that relocation and tries to map the BAR outside of the guest, which aborts the VM:

    kvm_set_user_memory_region: KVM_SET_USER_MEMORY_REGION failed,
    slot=5, start=0xfffffffffe000000, size=0x400000: Invalid argument

Disable decoding around the probe, as the PCI specification requires, and route the informational bus dump through get_bar so that it is covered as well.

Depends on https://github.com/hermit-os/kernel/pull/2611.

### DIFF
--- a/src/drivers/pci.rs
+++ b/src/drivers/pci.rs
@@ -76,9 +76,23 @@ impl<T: ConfigRegionAccess> PciDevice<T> {
 
 	/// Returns the bar at bar-register `slot`.
 	pub fn get_bar(&self, slot: u8) -> Option<Bar> {
-		let header = self.header();
-		let endpoint = EndpointHeader::from_header(header, &self.access)?;
-		endpoint.bar(slot, &self.access)
+		let endpoint = EndpointHeader::from_header(self.header(), &self.access)?;
+		let mut header = self.header();
+
+		// Determining the size of a bar writes all-ones into it and restores the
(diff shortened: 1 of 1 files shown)


src/drivers/pci.rs
```diff
@@ -181,6 +195,15 @@ impl<T: ConfigRegionAccess> PciDevice<T> {
 					return None;
 				}
 
+				// A legacy PCI interrupt is signalled level triggered and active
```
stlankes:
```suggestion
				// A legacy PCI interrupt is signaled level triggered and active
```

mkroening:
I rebased on main, since this included https://github.com/hermit-os/kernel/pull/2611, which is now merged.

mkroening (approved):
Great!

Do you think we should also upstream this to the `pci_types` crate?

stlankes:
Yes, I think it would be a good idea

mkroening:
I have opened a tracking issue so we remember this :)

- https://github.com/hermit-os/kernel/issues/2637

PR #2610, opened 2026-08-06
feat(xtask): Search in system paths for OpenSBI binary

Gelbpunkt:
This aligns the locations where xtask searches for the OpenSBI binary with the loader and uses the system-wide paths rather than a local directory.

See:
- https://github.com/hermit-os/loader/pull/629


mkroening (approved):
Thanks! :)

mkroening:
I added an option for local files so we macOS users can still run RISC-V VMs. :D

mkroening (approved):

PR #2611, opened 2026-08-06
fix(pci): configure legacy interrupts level triggered and active low

stlankes:
A legacy PCI interrupt is signalled level triggered and active low, while the IOAPIC defaults to the edge triggered, active high behavior of the ISA interrupts. A passed-through device therefore stops delivering interrupts after its first one, because the host only unmasks it once the guest has completed a level triggered interrupt.

Correct the redirection entry where a legacy interrupt line is handed out.


mkroening (approved):
Looks good to me! :)

PR #2613, opened 2026-08-08
feat(riscv64) Support multiple mmio devices

sbutz:
While there remove unsafe code to a minimum.

Will conflict with https://github.com/hermit-os/kernel/pull/2547.


mkroening (approved):
Thanks! :)

PR #2615, opened 2026-08-09
fix(x86_64): be UEFI agnostic

### LINKED ISSUE #2402: SMP boot code mapping got broken on Hermit

mkroening:
In preparation of https://github.com/hermit-os/kernel/pull/2546, this removes all occurrences of `env::is_uefi()` and handles every occurrence agnostically.

The first commit fixes https://github.com/hermit-os/kernel/issues/2402.

Depends on:
- https://github.com/hermit-os/kernel/pull/2619

### DIFF
--- a/src/arch/x86_64/kernel/acpi.rs
+++ b/src/arch/x86_64/kernel/acpi.rs
@@ -1,18 +1,14 @@
 use core::{ptr, slice, str};
 
 use align_address::Align;
-use free_list::{PageLayout, PageRange};
 use hermit_sync::OnceCell;
 use memory_addresses::{PhysAddr, VirtAddr};
 use x86_64::instructions::port::Port;
-use x86_64::structures::paging::PhysFrame;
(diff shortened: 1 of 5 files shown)


mkroening:
CC: @sarahspberrypi

PR #2616, opened 2026-08-09
feat: add the icmp-echo-reply feature

### LINKED ISSUE #2608: Missing support for ICMP pings
- [auto-icmp-echo-reply](https://github.com/smoltcp-rs/smoltcp/blob/v0.13.1/Cargo.toml#L101-L102)
- [smoltcp/examples/ping.rs](https://github.com/smoltcp-rs/smoltcp/blob/v0.13.1/examples/ping.rs)

stlankes:
smoltcp answers ICMP echo requests by itself once auto-icmp-echo-reply is set, for IPv4 as well as IPv6, so making a Hermit guest respond to ping needs nothing beyond the feature flag.

I tested the configuration on macOS as follows:

```sh
$ cargo build --release -Zbuild-std=std,panic_abort --target=aarch64-unknown-hermit -p httpd --features hermit/echo-reply,hermit/pci,hermit/virtio-net,hermit/dhcpv4
$ sudo qemu-system-aarch64 -display none -serial stdio \
  -machine virt,gic-version=3 \
  -cpu max,lpa2=off  \
  -smp 1 -m 1024M \
  -global virtio-mmio.force-legacy=off \
  -kernel hermit-loader-aarch64 \
  -device guest-loader,addr=0x48000000,initrd=target/aarch64-unknown-hermit/release/httpd \
  -netdev vmnet-shared,id=net0,start-address=192.168.68.2,end-address=192.168.68.254,subnet-mask=255.255.255.0 \
  -device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:56,disable-legacy=on,packed=on,mq=on
$ ping -c 5 192.168.68.3
PING 192.168.68.3 (192.168.68.3): 56 data bytes
64 bytes from 192.168.68.3: icmp_seq=0 ttl=64 time=2.790 ms
64 bytes from 192.168.68.3: icmp_seq=1 ttl=64 time=0.485 ms
64 bytes from 192.168.68.3: icmp_seq=2 ttl=64 time=0.724 ms
64 bytes from 192.168.68.3: icmp_seq=3 ttl=64 time=0.621 ms
64 bytes from 192.168.68.3: icmp_seq=4 ttl=64 time=0.855 ms
```

close #2608

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,8 +48,8 @@ jobs:
         run: |
           cargo clippy --target x86_64-unknown-none --features tcp,virtio-console,virtio-fs,virtio-net,virtio-vsock --all-targets
           cargo clippy --target x86_64-unknown-none --no-default-features --features tcp,virtio-console,virtio-fs,virtio-net,virtio-vsock
-          cargo hack clippy --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net
-          cargo hack clippy --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net --features pci
+          cargo hack clippy --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net,icmp-echo-reply
+          cargo hack clippy --each-feature --no-dev-deps --target x86_64-unknown-none --exclude-features dhcpv4,dns,gem-net,net,rtl8139,virtio-net,icmp-echo-reply --features pci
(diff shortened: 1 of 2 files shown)


mkroening (approved):
Great! :)

mkroening (changes requested):
Actually, since this requires no kernel changes, it might make sense to only add this in the wrapper crate, then.

Users that directly build the kernel can already enable `smoltcp/auto-icmp-echo-reply`. https://github.com/hermit-os/hermit-rs/pull/1060 adds a forward as `smoltcp-auto-icmp-echo-reply` similar to what we do for compile-time log levels.

stlankes:
> Actually, since this requires no kernel changes, it might make sense to only add this in the wrapper crate, then.
> 
> Users that directly build the kernel can already enable `smoltcp/auto-icmp-echo-reply`. https://github.com/hermit-os/hermit-rs/pull/1060 adds a forward as `smoltcp-auto-icmp-echo-reply` similar to what we do for compile-time log levels.

It is for me not clear, why this is the better way. But I am fine with it.

PR #2624, opened 2026-08-11
feat(start_info): add `StartInfo::modules`

mkroening:
This adds support for a start info impl to communicate physically loaded modules such as the initramfs.

Depends on https://github.com/hermit-os/kernel/pull/2625.

### DIFF
--- a/src/env/start_info/hermit_entry.rs
+++ b/src/env/start_info/hermit_entry.rs
@@ -4,7 +4,7 @@ use core::num::NonZero;
 use hermit_entry::boot_info::{BootInfo, RawBootInfo};
 use hermit_sync::OnceCell;
 
-use super::{FdtStartInfo, StartInfo};
+use super::{FdtStartInfo, Module, StartInfo};
 
 static START_INFO: OnceCell<BootInfo> = OnceCell::new();
 
@@ -47,6 +47,26 @@ unsafe impl StartInfo for BootInfo {
 			.starting_address
 			.addr();
 		NonZero::new(rsdp)
+	}
+
+	fn modules(&self) -> impl Iterator<Item = Module> {
+		fn initrd(fdt: fdt::Fdt<'_>) -> Option<Module> {
+			let chosen = fdt.find_node("/chosen")?;
+
+			let start = chosen.property("linux,initrd-start")?;
+			let end = chosen.property("linux,initrd-end")?;
+
+			let start = start.as_usize()?;
+			let end = end.as_usize()?;
+			let len = end.checked_sub(start)?;
+
+			// SAFETY: The bootloader guarantees the addresses to be valid.
+			let module = unsafe { Module::new(start, len) };
+
+			Some(module)
+		}
+
+		self.fdt().and_then(initrd).into_iter()
 	}
 }
 
--- a/src/env/start_info/mod.rs
+++ b/src/env/start_info/mod.rs
@@ -1,17 +1,16 @@
-#[cfg(feature = "hermit-entry")]
-mod hermit_entry;
-
-#[cfg(feature = "hermit-entry")]
-pub use hermit_entry::*;
-
-#[cfg(not(feature = "hermit-entry"))]
-mod unsupported;
+cfg_select! {
+	feature = "hermit-entry" => {
+		mod hermit_entry;
+		pub use self::hermit_entry::*;
+	}
+	_ => {
+		mod unsupported;
+		pub use self::unsupported::*;
+	}
+}
 
 use core::fmt;
 use core::num::NonZero;
-
-#[cfg(not(feature = "hermit-entry"))]
-pub use unsupported::*;
 
 pub unsafe trait StartInfo {
 	fn display(&self) -> impl fmt::Display {
--- a/src/env/start_info/mod.rs
+++ b/src/env/start_info/mod.rs
@@ -9,12 +9,22 @@ cfg_select! {
 	}
 }
 
-use core::fmt;
+mod module;
+
 use core::num::NonZero;
+use core::{fmt, iter};
+
+pub use self::module::Module;
 
 pub unsafe trait StartInfo {
 	fn display(&self) -> impl fmt::Display {
 		fmt::from_fn(|f| f.write_str("StartInfo::display not implemented"))
+	}
+
+	/// Returns the modules passed to the kernel at start.
+	#[cfg_attr(not(feature = "hermit-entry"), expect(dead_code))]
+	fn modules(&self) -> impl Iterator<Item = Module> {
+		iter::empty()
 	}
 
 	fn bootargs(&self) -> Option<&str> {
--- a/src/env/start_info/module.rs
+++ b/src/env/start_info/module.rs
@@ -1,0 +1,62 @@
+use core::{ptr, slice};
+
+use free_list::PageRange;
+
+use crate::page_range_ext::PageRangeExt;
+
+/// A module that is passed to the kernel at start.
+///
+/// This blob is loaded into physical memory by the bootloader or VMM and is described via the start info.
+///
+/// Examples of modules:
+/// - an initramfs
+/// - another kernel image
+///
+/// This type can be thought of as a physical-memory version of `&'static [u8]`.
+#[derive(Clone, Copy, Debug)]
+pub struct Module {
+	phys_addr: usize,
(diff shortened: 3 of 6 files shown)


src/env/start_info/mod.rs
```diff
@@ -61,3 +75,10 @@ pub trait UhyveStartInfo: FdtStartInfo {
 	#[cfg_attr(not(all(target_arch = "x86_64", feature = "smp")), expect(unused))]
 	fn uhyve_num_cpus(&self) -> Option<NonZero<usize>>;
 }
+
+#[expect(dead_code)]
+#[derive(Clone, Copy, Debug)]
+pub struct ModuleAddr {
```
fogti: Why is this `pub` but the functions above aren't?

mkroening: They are, since they are methods on a `pub trait`.

mkroening:
@fogti, I have added an implementation based on your code in https://github.com/hermit-os/kernel/pull/2390.

Does this look good to you? The only change should be that I used `NodeProperty::as_usize()` instead of doing the big-endian dance myself. I am not sure if this works, though. I don't know how to start Uhyve with an initramfs payload. Please tell me how to test this or test it yourself to make sure the parsing works. :)

src/env/start_info/hermit_entry.rs
```diff
@@ -49,6 +50,25 @@ unsafe impl StartInfo for BootInfo {
...
+
+			let start = chosen.property("linux,initrd-start")?;
+			let end = chosen.property("linux,initrd-end")?;
+
+			let start = start.as_usize()?;
+			let end = end.as_usize()?;
+
+			Some(ModuleAddr {
+				phys_addr: start,
+				len: end - start,
```
fogti: If we're already very careful above, we might as well use `end.checked_sub(start)?` here.

fogti:
`.as_usize()` apparently interprets the number as big-endian, which is what my code also expects, so this should be fine.

I'll try to rebase my branch on top of this later, and check if it still works.

src/env/start_info/hermit_entry.rs
mkroening: Good point. Done. 👍

src/env/start_info/hermit_entry.rs
```diff
@@ -49,6 +50,26 @@ unsafe impl StartInfo for BootInfo {
...
 	}
+
+	fn module_addrs(&self) -> impl Iterator<Item = ModuleAddr> {
+		fn module(fdt: Fdt<'_>) -> Option<ModuleAddr> {
+			let chosen = fdt.find_node("/chosen")?;
+
+			let start = chosen.property("linux,initrd-start")?;
+			let end = chosen.property("linux,initrd-end")?;
+
+			let start = start.as_usize()?;
```
fogti: Regarding style, it might be a good idea to put the `.as_usize()?` into the same line as `.property(_)?`, instead of interleaving them like this.

mkroening: I don't have a strong opinion, but I did also consider this and would like to keep it as is.

That way, the whole node property is available for debugging, and we don't have too much going on in one line (two `?`s for example).

jounathaen:
Can you explain to me your concept of a "Module"? I don't quite get it yet.

mkroening:
> Can you explain to me your concept of a "Module"? I don't quite get it yet.

I have added some docs. Essentially, it is some data that is written to memory by the entity that starts the kernel. For example, Uhyve is putting an initramfs into memory.

src/env/start_info/mod.rs
fogti: Another thing: I think `RawModuleSlice` would be a more appropriate name for this, given that it's not just an address.

src/env/start_info/mod.rs
```diff
@@ -65,3 +77,18 @@ pub trait UhyveStartInfo: FdtStartInfo {
...
+/// A module that is passed to the kernel.
+///
+/// The entity that starts the kernel may provide data to the kernel by writing it to memory.
+/// This type contains the corresponding physical address and length of that module.
+/// A common type of module is an initramfs.
+#[expect(dead_code)]
+#[derive(Clone, Copy, Debug)]
+pub struct ModuleAddr {
+	/// The physical address of the module.
+	pub phys_addr: usize,
```
fogti: Given that this is only used in very few places, wouldn't it make more sense to call this `physical_address` and get rid of the comment?

mkroening: `phys_addr` and `PhysAddr` should be very clear; these names are used everywhere and I consider them the canonical form. I just added a comment for completeness sake.

src/env/start_info/mod.rs
mkroening: It's also not a real slice, though, since slices correspond to pointers and virtual addresses, and this corresponds to a "physical slice". I don't want the name to imply easy sliceability, though, just as a place for start infos to put the physical addresses of their modules.

How about `PhysModule`, then?

fogti:
@jounathaen brought up that the "modules" name might be a bit confusing in regards to usual bootloader modules (e.g. in GRUB, `.mod`) or kernel modules (Linux' `.ko`).

mkroening:
> @jounathaen brought up that the "modules" name might be a bit confusing in regards to usual bootloader modules (e.g. in GRUB, `.mod`) or kernel modules (Linux' `.ko`).

When talking about starting a kernel, these things are usually called modules, and I would like to stick with the convention here. For example, Multiboot, PVH, and GRUB call them that.

This is, of course, different from a kernel module (`.ko`), but that should be clear from the context and the doc strings. I'll improve the docs a bit more; thanks for mentioning this. :)

src/env/start_info/hermit_entry.rs
```diff
@@ -48,6 +49,26 @@ unsafe impl StartInfo for BootInfo {
...
+			let end = end.as_usize()?;
+			let len = end.checked_sub(start)?;
+
+			Some(PhysModule {
+				phys_addr: start,
+				len,
+			})
+		}
+
+		self.fdt().and_then(|fdt| initrd(fdt)).into_iter()
```
fogti:
```suggestion
		self.fdt().and_then(initrd).into_iter()
```

fogti:
could you put the functionalities:
* to turn a `PhysModule` into a `PageRange`, and
* to turn a `PhysModule` into a `&[u8]`

either into methods on `PhysModule`, or just `impl From<PhysModule> for _`?

src/env/start_info/hermit_entry.rs
mkroening: I wonder why there is no Clippy lint for this. 🤔

mkroening:
> could you put the functionalities:
> 
>     * to turn a `PhysModule` into a `PageRange`, and
> 
>     * to turn a `PhysModule` into a `&[u8]`
> 
> 
> either into methods on `PhysModule`, or just `impl From<PhysModule> for _`?

This turned out to be more involved than the previous design. I chose the previous design because it is simple for start infos to implement. The new design should not be much worse, but it is a bit more cumbersome to make it sound.

src/env/start_info/module.rs
```diff
@@ -0,0 +1,57 @@
+#![expect(dead_code)]
```
fogti: Shouldn't this also be conditional on `feature = "hermit-entry"`?

mkroening: No, since there are accessors here that are not used even with `hermit-entry`. Perhaps I should move the attribute to the specific methods. 🤔 

What do you think?

mkroening: Done.

fogti:
Something still looks amiss:
```
error: associated items `new` and `phys_frame_range` are never used
  --> src/env/start_info/module.rs:28:16
   |
22 | impl Module {
   | ----------- associated items in this implementation
...
28 |     pub unsafe fn new(phys_addr: usize, len: usize) -> Self {
   |                   ^^^
...
55 |     pub fn phys_frame_range(&self) -> PageRange {
   |            ^^^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
   = help: to override `-D warnings` add `#[expect(dead_code)]` or `#[allow(dead_code)]`

error: could not compile `hermit-kernel` (lib) due to 1 previous error
```

PR #2632, opened 2026-08-12
fix(fs/mem): Detect when mount point is already a virtual directory

fogti:
This was noticed during initramfs testing.


mkroening (approved):
Thanks! :)

PR #2634, opened 2026-08-12
ci: improve clippy job

mkroening:
This should catch https://github.com/hermit-os/kernel/pull/2633 in the future.


fogti:
Why exactly do we have `cargo xtask clippy` if we don't use it in CI?

mkroening:
> Why exactly do we have `cargo xtask clippy` if we don't use it in CI?

We did use it once but found it not to be sufficient. Hence the migration to cargo-hack. I am planning to make this nicer in the future, though.

PR #2638, opened 2026-08-13
feat(network): Use ip= parameter for configuring networking

Gelbpunkt:
This is a breaking change and replaces `HERMIT_IP`, `HERMIT_MASK`, `HERMIT_GATEWAY`, `HERMIT_DNS1` and `HERMIT_DNS2` with a single command-line parameter, `ip=`, which is heavily inspired by [Linux](https://docs.kernel.org/admin-guide/nfs/nfsroot.html#kernel-command-line).

The kernel will fall back to DHCP if it is enabled, otherwise it will now also no longer fall back to "magic" IPs, instead it will leave the interface unconfigured. Instead, only xtask contains the default static IP configuration that used to reside there *and* in the kernel.

This is in preparation for adding IPv6 support, which we decided against introducing new variables for. The same mechanism introduced here can be used to support IPv6 with minimal changes.


mkroening:
@jounathaen and @sarahspberrypi are okay with the concept. Is this also ok for you, @stlankes?

src/env/mod.rs
```diff
@@ -61,24 +69,34 @@ impl Default for Cli {
...
-					env_vars.insert(String::from("HERMIT_IP"), ip);
-				}
-				"-mask" => {
-					let mask = expect_arg(words.next(), word.as_str());
-					env_vars.insert(String::from("HERMIT_MASK"), mask);
-				}
-				"-gateway" => {
-					let gateway = expect_arg(words.next(), word.as_str());
-					env_vars.insert(String::from("HERMIT_GATEWAY"), gateway);
-				}
```
mkroening, line 81: Could you add a warning that these are no longer supported and `ip=` should be used instead? Same thing for the corresponding `hermit_var!(...)`s, which are commonly set at compile time at the moment.

src/env/ip_configuration.rs
```diff
@@ -0,0 +1,179 @@
+use core::net::Ipv4Addr;
+use core::{fmt, str};
+
+use smoltcp::wire::{IpCidr, Ipv4Cidr};
+
+#[derive(Clone, Copy, Debug, Default)]
+pub enum IpAddressConfiguration {
```
mkroening: Nit: rename to `IpAddrConfig`. Similar for the other structs and the module.

src/env/ip_configuration.rs
```diff
@@ -0,0 +1,179 @@
...
+///
+/// Fields are specified separated by colons (:), for example:
+/// - ip=none or ip=off to skip configuring the default interface
+/// - ip=dhcp to use DHCP for configuring the default interface
+/// - ip=10.0.5.3/24:10.0.5.1 to configure a static IP and gateway on the default interface
+///
+/// This is heavily inspired by the Linux kernel's parameter of the same name:
+/// <https://docs.kernel.org/admin-guide/nfs/nfsroot.html#kernel-command-line>
+#[derive(Clone, Copy, Debug, Default)]
+pub struct IpConfiguration {
```
mkroening: Nit: Move the most important type to the top of the file and the helpers down to help readers with discoverability.

mkroening (changes requested):
Thanks for taking care of this! :)

Does the format also work well with IPv6 addresses? I assume we would do something like `ip=[2001:db8::8a2e:370:7334]:[2001:db8::8a2e:370:7334]`, right?

src/env/mod.rs
Gelbpunkt: Done, wasn't so sure about the warning messages but I hope this is fine

src/env/ip_configuration.rs
Gelbpunkt: I think I ordered them in a sensible order now, please take a look :)

Gelbpunkt:
> Does the format also work well with IPv6 addresses? I assume we would do something like `ip=[2001:db8::8a2e:370:7334]:[2001:db8::8a2e:370:7334]`, right?

Yes, this is roughly what I had in mind. It will work just fine with IPv6 addresses once we do a tiny bit of modifications (since this currently hardcodes IPv4 types, but that's a trivial change later).

I believe `dhcp` should just imply both `dhcpv4` and `dhcpv6` eventually if both are available and if we want to use both IPv6 and IPv4 networking on the same interface, we can pass `ip=` simply twice, once for configuring IPv4 and once for IPv6. Alternatively we could rename `dhcp` to `dhcpv4` and make it explicit.

mkroening (approved):
Looks good to me. 👍 

Let's wait for @stlankes's opinion in a week. :)

stlankes (changes requested):
I have a few questions...

If I the invalid  ip `ip=10.0.5.4/64`, I got following panic:

```
[0][PANIC] panicked at /Users/stefan/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/smoltcp-0.14.0/src/wire/ipv4.rs:101:9:
assertion failed: prefix_len <= 32
```

Also an empty gateway panics. For instance, `ip=10.0.5.3/24::` creates following panic.

```
[0][PANIC] panicked at src/env/mod.rs:86:31:
Could not parse configuration for default interface: invalid gateway IP address
```

Should we panic on invalid input, or just log a warning?

But I like the PR!

stlankes:
@Gelbpunkt I hope that I fix merge conflicts correctly

stlankes (approved):

stlankes:
I talked to @mkroening and now I am finde with the PR

PR #2642, opened 2026-08-15
refactor: implement `is_uhyve()` via FDT

mkroening:


fogti (approved):

src/env/start_info/hermit_entry.rs
```diff
@@ -33,9 +33,14 @@ unsafe impl FdtStartInfo for BootInfo {
...
-		use hermit_entry::boot_info::PlatformInfo;
+		let Some(fdt) = self.fdt() else {
+			return false;
+		};
 
-		matches!(self.platform_info, PlatformInfo::Uhyve { .. })
+		fdt.root()
+			.compatible()
+			.all()
+			.any(|compatible| compatible.contains("uhyve"))
```
fogti: testing for `",uhyve"` might be more precise. Or just for equality with `"hermit,uhyve"` (tho that might be too restrictive in regards to forward-compatibilty) (`.compatible().all()` splits on null-bytes)

mkroening: What is the benefit of ensuring we only match Uhyve models and not Uhyve manufacturers?

fogti: my main concern would be that we match any string containing `uhyve`, and not just those in which `uhyve` is a separate word (optimally, I'd prefer splitting (via `.flat_map(_)`) also on commas and maybe white-space, then match exactly `uhyve` on any of those components).

mkroening: What would be an example of a string containing `uhyve`, that we don't want to match?

fogti: idk, but given the usual VMM naming scheme of `${something}hyve`, smth. like `auhyve` wouldn't be that unexpected.

mkroening: Okay, then let's go with `hermit,uhyve` for now.

fogti (approved):

PR #2645, opened 2026-08-16
feat: make `loader` and `uhyve` non-default features

mkroening:
This PR prepares for:
- https://github.com/hermit-os/kernel/pull/2366

Currently, PVH is mutually exclusive with `loader` and `uhyve` and I don't think it is worth the effort to support multiple distinct start info structs at the same time. This PR makes `loader` and `uhyve` non-default features, so any start method needs to be opted into.


stlankes (approved):

mkroening:
We need to discuss this first.

PR #2650, opened 2026-08-16
refactor(acpi): migrate to arch-independent `acpi` crate

mkroening:
This is useful for:
- https://github.com/hermit-os/kernel/issues/2649

This closes:
- https://github.com/hermit-os/kernel/issues/391

This depends on:
- https://github.com/hermit-os/kernel/pull/2667

PR #2656, opened 2026-08-17
fix(scheduler): avoid boot loop when aborting before core scheduler is initialized

mkroening:
Found while working on:
- https://github.com/hermit-os/kernel/pull/2650


mkroening:
Hmm, I cannot reproduce this anymore.

PR #2662, opened 2026-08-18
make `read_entropy` forward-secure

joboet:
This makes the RNG used for `read_entropy` forwards-secure by employing J.D. Bernstein's [fast-key-erasure construction](https://blog.cr.yp.to/20170723-random.html) while keeping ChaCha20 as the primitive. The blog post contains a more detailed description, but simply put the problem with the current RNG is that it is possible to reconstruct past outputs (and not just future ones) by looking at the current RNG state. The fast-key-erasure RNG prevents this while preserving the performance of the underlying cryptographic primitive.

The other change introduced here is that `read_entropy` will not ever fail once the initial seed is available. The only consequence of this is that the RNG will not be able to recover from state compromise if the hardware or hypervisor source becomes inaccessible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -885,6 +885,7 @@ dependencies = [
  "virtio-spec",
  "volatile 0.6.1",
  "x86_64",
+ "zeroize",
 ]
 
 [[package]]
@@ -1492,7 +1493,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3e6af7f3e25ded52c41df4e0b1af2d047e45896c2f3281792ed68a1c243daedb"
 dependencies = [
  "ppv-lite86",
- "rand_core 0.10.0",
+ "rand_core 0.10.1",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


mkroening (approved):
Thanks, looks good to me! Could you take care of Clippy? :)

joboet:
Sure, done! I had to switch back to `rand_chacha` since `chacha20` doesn't compile with SSE2 disabled (LLVM returns "Do not know how to split the result of this operator!", see https://github.com/rust-lang/rust/issues/158532) – unfortunately that means the RNG has to be zeroised manually. Also, `rand_chacha` doesn't have `set_block_pos`, only `set_word_pos`, so I've emulated that.

mkroening:
> Sure, done! I had to switch back to `rand_chacha` since `chacha20` doesn't compile with SSE2 disabled (LLVM returns "Do not know how to split the result of this operator!", see [rust-lang/rust#158532](https://github.com/rust-lang/rust/issues/158532)) – unfortunately that means the RNG has to be zeroised manually. Also, `rand_chacha` doesn't have `set_block_pos`, only `set_word_pos`, so I've emulated that.

Thanks! :)

PR #2663, opened 2026-08-18
build(deps): Upgrade to arm-gic 0.9

Gelbpunkt:
Since the arm-gic crate can now figure out whether we have a GICv3 or GICv4 on its own, we don't need to inspect the compatible of the interrupt controller node anymore.

While I'm at it, I decided to replace the hardcoded lookup path for the GIC with a lookup by compatible to ensure that we retain the previous behavior of panic on unsupport GIC version.


mkroening (approved):
Thanks! :)

PR #2666, opened 2026-08-19
tests: update default loader path

mkroening:
Depends on https://github.com/hermit-os/kernel/pull/2660.


mkroening:
Seems too invasive. I'll remove this soon anyway, hopefully.

PR #2668, opened 2026-08-19
feat(acpi): support shutdown of HW-reduced ACPI platforms

### LINKED ISSUE #2649: Modern ACPI shutdown is missing
We currently implement ACPI shutdowns via PM1, which is legacy and not supported by all platforms (e.g., Cloud Hypervisor). We should use FADT to shut down instead.

mkroening:
This makes our ACPI shutdown procedure more conformant and adds support for shutting down HW-reduced ACPI platforms such as Cloud Hypervisor.

Closes https://github.com/hermit-os/kernel/issues/2649.

Depends on:
- https://github.com/hermit-os/kernel/pull/2650


mkroening:
I have rebased on the AML interpreter drop, and I dropped d5dea6634b0c376bdb6d9ecb0d0f868f8aed0d7e and 82e872ebf94ae46370511969b83d7fe764b182e3.

PR #2669, opened 2026-08-19
fix(virtqueue): order descriptor publication against the notification check

stlankes:
Both rings decide whether to notify the device by reading state the device writes, right after publishing a descriptor, with nothing ordering the two. The preceding barrier orders the descriptor contents against the field that publishes them, not that field against this later load, so a weakly ordered machine can read the flag before our publication is visible.

This causes deadlocks in polling mode, which are prevented by adding memory barriers.


mkroening (approved):
Looks good to me! 👍

PR #2677, opened 2026-08-24
Implement posix_getdents

### LINKED ISSUE #2034: Implement `posix_getdents`
We currently implement the following non-standardized Linux interface for getting directory entries, [getdents(2)](https://www.man7.org/linux/man-pages/man2/getdents64.2.html):

```c
ssize_t getdents64(int fd, void dirp[.count], size_t count);

struct linux_dirent64 {
    ino64_t        d_ino;    /* 64-bit inode number */
    off64_t        d_off;    /* Not an offset; see getdents() */
    unsigned short d_reclen; /* Size of this dirent */
    unsigned char  d_type;   /* File type */
    char           d_name[]; /* Filename (null-terminated) */
};
```

> <ins>d_off</ins> is a filesystem-specific value with no specific meaning to user space, though on older filesystems it used to be the distance from the start of the directory to the start of the next <ins>linux_dirent</ins>; see readdir(3).

We could gradually adopt the following standardized interface, [posix_getdents](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_getdents.html), [<dirent.h>](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/dirent.h.html):

```c
ssize_t posix_getdents(int fildes, void *buf, size_t nbyte, int flags);

struct posix_dent {
    ino_t          d_ino      File serial number.
    reclen_t       d_reclen   Length of this entry, including trailing
                              padding if necessary. See posix_getdents().
    unsigned char  d_type     File type or unknown-file-type indication.
    char           d_name[]   Filename string of this entry.
};
```

This would mainly benefit from being standardized and something we can commit to. Additionally, it is nice to get rid of the very unusual `d_off`.

RajeshRk18:
Closes #2034.

Adds `sys_posix_getdents` as specified by [POSIX.1-2024 `posix_getdents`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_getdents.html), alongside the existing `sys_getdents64`.

```rust
pub unsafe extern "C" fn sys_posix_getdents(fd: RawFd, buf: *mut PosixDent, nbyte: usize, flags: i32) -> isize
```

`PosixDent` matches `struct posix_dent { d_ino, d_reclen, d_type, d_name[] }` . `flags` must be `0` for now because POSIX defines no mandatory flags and `DT_FORCE_TYPE` is optional). Anything else returns `EINVAL`.

## Implementation

- The entry-serialization logic that was duplicated across `mem.rs` and `virtio_fs.rs` is now a single `DirentFormat::write_entry`, parameterized over the two entry formats. `ObjectInterface::getdents` takes the requested `DirentFormat`.
- The default `ObjectInterface::getdents` now returns `ENOTDIR` instead of `EINVAL`, matching both POSIX and Linux `getdents64(2)` behavior for non-directory descriptors.

## fixes surfaced by the new tests

- `DirectoryReader::getdents` was `unimplemented!()`, so `sys_opendir` + `sys_getdents64` panicked the kernel. It is now implemented.
- `MemDirectory::traverse_readdir` never resolved the final path component.`readdir("/tmp/gd")` listed the contents of `/tmp` instead of `/tmp/gd`. Both went unnoticed because nothing exercised the `sys_opendir` path before.

## Tests

New `tests/getdents.rs` integration test covers both syscalls through both the `sys_opendir` and `O_DIRECTORY` paths: entry layout/reclen for both formats, end-of-directory, partial reads with a one-entry buffer, rejected flags, `EINVAL` for a too-small buffer, and `ENOTDIR` for a file descriptor.

### DIFF
--- a/tests/getdents.rs
+++ b/tests/getdents.rs
@@ -8,6 +8,8 @@ extern crate alloc;
 
 mod common;
 
+use alloc::ffi::CString;
+use alloc::format;
 use alloc::string::String;
 use alloc::vec::Vec;
 use core::ffi::CStr;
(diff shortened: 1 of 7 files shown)


tests/getdents.rs
```diff
@@ -0,0 +1,150 @@
...
+	check_dir_fd(fd);
+
+	// A non-directory file descriptor
+	let mut buf = Buf([0; 1024]);
+	let fd = unsafe { sys_open(c"/tmp/gd/a".as_ptr(), O_WRONLY, 0) };
+	assert!(fd >= 0);
+	let len = unsafe { sys_posix_getdents(fd, buf.0.as_mut_ptr().cast(), buf.0.len(), 0) };
+	assert_eq!(len, -isize::try_from(ENOTDIR).unwrap());
+	assert_eq!(sys_close(fd), 0);
+}
```
jounathaen, line 146: Could you add a test case with a long directory and long filenames so that more than one call to `sys_posix_getdents` has to be performed? We had some bugs in the code before, exactly because we (read as "I") didn't test this.

jounathaen:
Overall, looks pretty good to me!

jounathaen (approved):
Very nice 👍. (Just the CI doc test needs a little fix)

RajeshRk18:
> Very nice 👍. (Just the CI doc test needs a little fix)

thanks! fixing it

jounathaen (approved):

jounathaen:
The merge commit is quite ugly. @RajeshRk18 do you want to do a rebase? Else, I can also do it in the following days.

RajeshRk18:
> The merge commit is quite ugly. @RajeshRk18 do you want to do a rebase? Else, I can also do it in the following days.

No worries, i will do it:)

jounathaen:
Thank you!

PR #2682, opened 2026-08-26
feat(virtio-fs): add support of fsync

stlankes:
Files send FUSE_FSYNC with the handle they already hold. Directory handles keep no handle, so fsync opens and releases one around FUSE_FSYNCDIR, the way getdents opens one for reading.

A host that does not implement the operation answers ENOSYS, which by FUSE convention means there was nothing to do; other errors now map to the reported errno instead of a blanket EIO.


mkroening (approved):
Looks okay to me. 👍

PR #2685, opened 2026-08-27
fix(env,logging): fix logging initialization

fogti:


fogti:
For good measure because we run into an allocation failure, this also tries to get rid of as many allocations as possible.

mkroening (commented):
Why not just increase the size of the initial heap?

fogti:
Because there is no upper limit to the amount of environment variables that can be passed?

Linux allows [32 * `PAGE_SIZE`](https://unix.stackexchange.com/questions/336934/raise-128kib-limit-on-environment-variables-in-linux) as the maximum length of the content of an environment variable, and a realistic limit for the entire env var block in Linux is `ARG_MAX`, which e.g. for my system is 2 MiB.

An unconditional ≈2 MiB initial heap just to be on the safe side sounds a bit … excessive

mkroening:
My point is that this solution is a bit involved and makes the argument parsing less local. So we trade clean code for unbounded env vars at the time of logging init. I am fine with that trade-off; I just want to be conscious about it. There are reasonable upper limits to what we can expect from environment variables, after all, so increasing the size would not be unreasonable.

fogti:
Note that I can also limit how "decentral" this makes the argument parsing by putting the logic for scanning for `HERMIT_LOG_LEVEL_FILTER` into `env/mod.rs`, and invoking it from `logging.rs`.

src/env/mod.rs
```diff
@@ -123,10 +121,10 @@ pub fn freq() -> Option<u16> {
 
 #[allow(dead_code)]
 pub fn var(key: &str) -> Option<&String> {
```
fogti, line 123: TODO: `early_var`

mkroening (approved):
Looks good to me, thanks! :)

mkroening:
CI is unhappy, though. ^^

fogti:
Now CI looks happier.

PR #2688, opened 2026-08-30
implement MappedPageBox

RajeshRk18:
Closes #2338

Added MappedPageBox and used it in two call sites.

Also removed `#[expect(dead_code)]` in src/arch/aarch64/mm/paging.rs because new public API makes it externally reachable.

### DIFF
--- a/src/mm/mapped_page_box.rs
+++ b/src/mm/mapped_page_box.rs
@@ -1,0 +1,64 @@
+use core::alloc::AllocError;
+
+use free_list::PageLayout;
+use memory_addresses::{PhysAddr, VirtAddr};
+
+use crate::arch::mm::paging::{self, BasePageSize, PageSize, PageTableEntryFlags};
+use crate::mm::{FrameBox, PageBox};
+
+/// A range of pages that is mapped for as long as this box is alive.
+pub struct MappedPageBox {
+	_frames: Option<FrameBox>,
+	pages: PageBox,
+}
+
+impl MappedPageBox {
+	/// Allocates the pages and frames described by `layout` and maps them.
+	pub fn new(layout: PageLayout, flags: PageTableEntryFlags) -> Result<Self, AllocError> {
+		let frames = FrameBox::new(layout)?;
+		let pages = PageBox::new(layout)?;
+		let page_count = pages.len().get() / BasePageSize::SIZE as usize;
+		paging::map::<BasePageSize>(
+			VirtAddr::from(pages.start()),
+			PhysAddr::from(frames.start()),
+			page_count,
+			flags,
+		);
+
+		Ok(Self {
+			_frames: Some(frames),
+			pages,
+		})
+	}
+
+	/// Allocates the pages described by `layout` and maps them to `phys_addr`.
+	///
+	/// # Safety
+	///
+	/// - The frames at `phys_addr` must not be deallocated while the returned box is alive.
+	pub unsafe fn map_phys(
+		phys_addr: PhysAddr,
+		layout: PageLayout,
+		flags: PageTableEntryFlags,
+	) -> Result<Self, AllocError> {
+		let pages = PageBox::new(layout)?;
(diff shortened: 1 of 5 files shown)


src/mm/mapped_page_box.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl Drop for MappedPageBox {
+	fn drop(&mut self) {
+		paging::unmap::<BasePageSize>(VirtAddr::from(self.pages.start()), page_count(&self.pages));
+
+		// `paging::unmap` only flushes the TLB of this CPU. Request the flush on
+		// all other CPUs before the frames can be reused.
+		#[cfg(all(target_arch = "x86_64", feature = "smp"))]
+		crate::arch::kernel::apic::ipi_tlb_flush();
```
mkroening: This seems costly. Is this necessary? We do an IPI TLB flush on mapping after all.

src/mm/mapped_page_box.rs
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+impl Deref for MappedPageBox {
+	type Target = PageRange;
+
+	fn deref(&self) -> &Self::Target {
+		&self.pages
+	}
+}
```
mkroening: This seems non-obvious to me, since there are two types of ranges inside. If we need something like this, I'd prefer accessor methods.

src/mm/mapped_page_box.rs
```diff
@@ -0,0 +1,79 @@
...
+
+fn page_count(pages: &PageBox) -> usize {
+	pages.len().get() / BasePageSize::SIZE as usize
+}
```
mkroening: Please remove this helper function.

src/mm/mapped_page_box.rs
```diff
@@ -0,0 +1,79 @@
...
+
+fn map(pages: &PageBox, phys_addr: PhysAddr, flags: PageTableEntryFlags) {
+	paging::map::<BasePageSize>(
+		VirtAddr::from(pages.start()),
+		phys_addr,
+		page_count(pages),
+		flags,
+	);
+}
```
mkroening: Please remove this helper function.

mkroening (changes requested):
Thanks! Did you use any AI assistance? :)

RajeshRk18:
> Thanks! Did you use any AI assistance? :)

yeah i am clearing unwanted stuff it added here!

src/mm/mapped_page_box.rs
RajeshRk18: will see how i can come with a better drop

RajeshRk18: @mkroening I removed the call now. But the ipi sent on mapping flushes TLB on other cpus only if the entry is present in the page table. since we delete the entry on unmap, stale TLB can still remains on other cpus.

src/mm/mapped_page_box.rs
```diff
@@ -0,0 +1,64 @@
...
+		let page_count = pages.len().get() / BasePageSize::SIZE as usize;
+		paging::map::<BasePageSize>(VirtAddr::from(pages.start()), phys_addr, page_count, flags);
+
+		Ok(Self {
+			_frames: None,
+			pages,
+		})
+	}
+
+	pub fn pages(&self) -> &PageBox {
```
mkroening, line 54:
```suggestion
	pub fn pages(&self) -> &PageRange {
```

mkroening (changes requested):
Thanks, this looks good to me. One more suggestion, though.

src/mm/mapped_page_box.rs
mkroening: @RajeshRk18, friendly ping. :)

mkroening:
Could you also add attribution to the AI coding assistants to your commits as described here? [AI Coding Assistants — The Linux Kernel documentation](https://docs.kernel.org/process/coding-assistants.html)

Issue #2614, opened 2026-08-09
riscv aia support


sbutz:
- [ ] Support APLIC in direct delivery mode (replacement for legacy PLIC)
- [ ] Support APLIC in msi delivery mode (includes IMSIC Support)
- [ ] Use MSIs to send IPI without trapping to m-mode
- [ ] Support MSI for PCI devices (PCI support is added with https://github.com/hermit-os/kernel/pull/2550)

sbutz:
@mkroening should an application binary support PLIC, APLIC (direct delivery mode) and APLIC+IMSIC (msi delivery mode) or should we add a feature flag(s) for aia support?

mkroening:
Good question! On x86-64, we went for a featureless introduction of MSI-X in https://github.com/hermit-os/kernel/pull/2101. That's because MSI-X is always faster and always preferred. Eventually, we would like to fade out non-MSI-X support on x86-64.

I am not too familiar with the situation on RISC-V. Is PLIC still the main way to do interrupts, or is APLIC or APLIC+IMSIC clearly the modern solution?

Are they compatible with each other? On x86-64 a given single system can support both legacy interrupts and MSI-X at the same time. If they are mutually exclusive, then features would make sense for sure.

Which feature names and documentation would you have in mind?

sbutz:
The spec allows to have systems which have an PLIC and an APLIC.
Each core can only receive interrupts from one of them.
So having PLIC and APLIC in one system implies heterogenous cores. So asymmetric multi processing.
IMHO it's unlikely we'll see such a configuration. Especially in virtualized environments.
Furthermore RustyHermit does not support AMP.
I'd save we can use feature flags instead of runtime support for both.

Before i propose a set of feature flags, i'd like to extend the scope of this question a bit.
RISC-V is modular architecture. The base ISA has a very limited set of registers and instructions.
Extensions add additional features. E.g. A (Atomic Instructions), F (Floating Point Operations), Zkr (Seed Register), ...
How should RustyHermit handle this.
Detecting extensions at runtime via devicetree is possible, but cumbersome and might impact performance.
Adding feature flags for every extension make configuration complicated.
The RISC-V community faced a similar problem. The modular architecture, lead to a lot of fragmentation.
The solution: Profiles. Profiles bundle a set of mandatory extension. The most prominent is RVA23.
Ubuntu 25.x and forward only run on RVA23 compliant hardware.
[...]

PR #2689, opened 2026-08-30
feat(riscv): Support advanced interrupt architecture

sbutz:
Implements #2614.

Feature flag is added in https://github.com/hermit-os/hermit-rs/pull/1077.

Follow-up tasks:

- [ ] For each configured msix vector. The external interrupt with the same number is enabled too. This might incur unnecessary cpu traps. Proposed solution: Implement msi vector allocator and store handlers per imsic.
- [ ] Bump riscv crate once new version available.
- [ ] Specify qemu cpu profile rva23s64 once qemu reaches version 10.1 on ci


src/arch/riscv64/kernel/interrupts/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+	trace!("Interrupt end");
+}
+
+/// Handles external interrupts
+fn external_handler() {
+	use crate::arch::kernel::core_local::core_scheduler;
+	use crate::scheduler::PerCoreSchedulerExt;
+
+	// Claim interrupt
+	let Some(irq) = EXTERNAL_INTERRUPT_CONTROLLER
```
sbutz: There's one APLIC or PLIC in the system. Therefore locking the global interrupt controller was reasonable.
However if the system fully supports AIA, there's an IMSIC per core and therefore no need to lock the platform level interrupt controller to claim an interrupt.

PR #2713, opened 2026-09-10
fix(physicalmem): Map physical memory after it has been reserved

mkroening:
Currently, we first make memory available, map that memory and then carve out some reservations that should not be used. That is wrong since we currently use the avialable memory for additional page tables during the mapping. To fix this, we map later, after carving out all reservations. This also reduces the total reported allocation size, making it more correct.


fogti:
Looks good at a glance, I'll have to test it, tho.

fogti (approved):

PR #2714, opened 2026-09-10
Small mm adjustments

zyuiop:
* Renames `mm::map` to `mm::device_map`, as it's only used for devices, and the current naming seems to imply a more general usecase
* Makes a couple offset computation functions `const` to hopefully replace them with their values at compile time

### DIFF
--- a/src/mm/device_alloc.rs
+++ b/src/mm/device_alloc.rs
@@ -43,8 +43,8 @@ unsafe impl Allocator for DeviceAlloc {
 impl DeviceAlloc {
 	/// Returns a pointer corresponding to `phys_addr`.
 	#[inline]
-	pub fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
-		let addr = phys_addr.as_usize() + self.phys_offset().as_usize();
+	pub const fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
+		let addr = phys_addr.as_usize() + const { Self.phys_offset().as_usize() };
 		ptr::with_exposed_provenance_mut(addr)
 	}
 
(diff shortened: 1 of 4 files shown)


src/mm/device_alloc.rs
```diff
@@ -43,8 +43,8 @@ unsafe impl Allocator for DeviceAlloc {
 impl DeviceAlloc {
 	/// Returns a pointer corresponding to `phys_addr`.
 	#[inline]
-	pub fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
-		let addr = phys_addr.as_usize() + self.phys_offset().as_usize();
+	pub const fn ptr_from<T>(&self, phys_addr: PhysAddr) -> *mut T {
+		let addr = phys_addr.as_usize() + const { Self.phys_offset().as_usize() };
```
mkroening, line 47: Is this explicit const context necessary?

```suggestion
		let addr = phys_addr.as_usize() + Self.phys_offset().as_usize();
```

mkroening (changes requested):
I believe we should not just rename `map` to `device_map`. Even though it is currently only used for devices, the function is clearly meant to be more general than that, due to the arguments.

It would be preferable to just remove the function and call the paging functions directly at the single call site.

Regarding constification. I think it is fine if you want this, though we lose one assertion. I don't believe there should be any performance benefits here, as inlining and constant folding should not be influenced by constness. Or did you find these to not optimize well enough? Do you need this to use these in future const-contexts? That would be the only reason to need constness as far as I am aware.

zyuiop:
I agree, we can remove the `map` function entirely. The goal with the renaming was indeed to prevent any additional usage, since this function seems entirely redundant.

Re. the constant changes, I was looking at assembly on Godbolt, but forgot to enable optimizations. With opt-level 2, it seems that this does not make a difference at all, and that `phys_offset` could indeed remain non-const

PR #2717, opened 2026-09-14
Skip the CI if no code files were touched

jounathaen:
Naive approach to reduce CI run amount a little.

Would (hopefully) skip the CI runs on PRs like https://github.com/hermit-os/kernel/pull/2716


jounathaen:
(Irony: This PR changes `benchmark.yml` and `ci.yml`, so it doesn't apply to itself :wink: )

PR #2721, opened 2026-09-15
ci: split clippy into three jobs

jounathaen:
With https://github.com/hermit-os/hermit-rs/pull/1088, the clippy job is now the longest.
I guess it makes sense to parallelize it per arch.


jounathaen:
So now we are at ~11 minutes for each clippy job.

We also have to change the "Required" settings to the individual clippy jobs.

mkroening (approved):
Looks okay to me.

PR #2724, opened 2026-09-18
feat(network): Add IPv6 support

Gelbpunkt:
This adds support for specifying IPv6 addresses in the ip= parameter and wires it up to the interface configuration. With these changes, we can reach a Hermit VM over IPv6. However, since our test setup is currently based on the assumption that many of the tests only perform e.g. a single HTTP request, test coverage is quite lackluster. The loopback example can trivially be adjusted to test both IPv4 and IPv6.

Loopback example change: https://github.com/hermit-os/hermit-rs/pull/1089


src/syscalls/socket/addrinfo.rs
```diff
@@ -429,14 +431,23 @@ fn getaddrinfo_node(
...
 	if ai_flags.contains(Ai::ADDRCONFIG) {
-		if want_ipv4 {
-			// Currently, Hermit always has an IPv4 address
-			want_ipv4 = true;
+		#[cfg(feature = "net")]
+		{
+			let Some(mut guard) = NIC.try_lock() else {
+				// FIXME: As per the man page, getaddrinfo can return EAGAIN, but might be inappropriate here
+				return Err(Eai::Again);
+			};
```
mkroening: I believe we should rather lock here. I doubt that applications handle this kind of failure, since this is not blocking on I/O but only blocking on other cores holding the lock for a short while.

xtask/hermit-ifup
```diff
@@ -4,4 +4,5 @@
 # Usage: .. -netdev tap,script=hermit-ifup,..
 
 ip address add dev "$1" local 10.0.5.1/24 broadcast 10.0.5.255
+ip address add dev "$1" local fdfb:5494:2d56::1/48
```
mkroening, line 7: Where does this address come from? Is this derived from the default MAC address? We should specify in a comment how to change this if a different MAC address is set via the QEMU cmdline.

mkroening (changes requested):
Thanks, this is great! :)

src/syscalls/socket/addrinfo.rs
Gelbpunkt: Sounds reasonable, changed accordingly

xtask/hermit-ifup
Gelbpunkt: This is just the first subnet that https://www.unique-local-ipv6.com/ output for me :D

mkroening: I see. So this is the same address that is also hardcoded in xtask. How does this address relate to the Hermit-generated link-local address? :)

Gelbpunkt: This is a ULA ([unique local address](https://en.wikipedia.org/wiki/Unique_local_address)) in the `fc00::/7` subnet which is destined for addresses only reachable in private networks. It's kind of the IPv6 equivalent to the IPv4 private networks like `10.0.0.0/8`, which was being used for IPv4.

The link-local addresses are in the subnet of `fe80::/10` and only reachable on a local link (and also ignored by routers).

mkroening (approved):

mkroening:
Thanks you so much! :)

mkroening:
I merged https://github.com/hermit-os/hermit-rs/pull/1089 first, to test it here.

